PR #131, opened 2026-06-20
feat(admin-client): OAuth bearer auth to admin (dual-auth fallback)

kavin-114:
## What & why

Bench-side counterpart to admin PR **Aerele-RnD/jarvis-admin#139** (bench→admin OAuth). The customer bench now authenticates to the admin control plane with **short-lived OAuth bearer tokens** instead of the static `api_key:api_secret`, shrinking the blast radius of a leaked bench credential.

## How it works

`admin_client._post` now:
1. **Prefers OAuth.** Exchanges the customer's password for an access token at admin's native `…/oauth2.get_token` (**password grant**, public client `jarvis-bench`, **no `client_secret`**), caches it in Redis (`frappe.cache()`, site-scoped), and sends `Authorization: Bearer <token>`.
2. **Reuses the cached token** until ~1 min before expiry; on a miss, **refreshes** (best-effort) then **re-mints** via the password grant. A bearer **401 re-mints once**, then falls back.
3. **Falls back to legacy** `token <api_key>:<api_secret>` when no password is stored — existing customers keep working (admin accepts both during migration).

### Credential storage
New `Jarvis Settings` fields (read-only, populated at onboarding):
- `jarvis_admin_customer_email` (Data) — the OAuth grant username.
- `jarvis_admin_customer_password` (Password) — exchanged for bearer tokens.

`onboarding.py` persists these via `write_connection` on the signup response, and captures the password admin delivers on the **verified signup poll** (`check_signup_payment_state`) — matching the admin-side change that defers the password until the email is verified.

## Testing

- `test_admin_client` (43, incl. 4 new): bearer preferred + omits `client_secret`, token caching across calls, legacy fallback without password, bearer-401 → re-mint → legacy.
- `test_onboarding_sync` (22, incl. extended): email/password persisted on verify-on + flag-off signup shapes, and password captured on the verified poll.
- All HTTP mocked, following the existing `_fake_post` pattern. Ran on `site.jarvis`.

## Deploy notes

- `bench migrate` adds the two `Jarvis Settings` fields.
- Depends on admin PR #139 (the `jarvis-bench` OAuth client + verified-poll password delivery) being deployed for the OAuth path to activate; until then, benches use the legacy fallback unchanged.

### DIFF
--- a/jarvis/admin_client.py
+++ b/jarvis/admin_client.py
@@ -1,14 +1,20 @@
 """HTTPS client for the Jarvis admin (jarvis_admin) app.
 
-Authenticated calls use Frappe's native api_key:api_secret. The customer's
-bench reads both from Jarvis Settings (set at signup) and sends them as
-`Authorization: token <api_key>:<api_secret>`.
+Authenticated calls prefer a short-lived OAuth bearer token: the bench
+exchanges the customer's password (Jarvis Settings.jarvis_admin_customer_password,
+username = jarvis_admin_customer_email) for an access token at admin's native
+Frappe OAuth endpoint using the shared public client id `jarvis-bench`, caches
+it in Redis, and sends `Authorization: Bearer <access_token>`. Customers
+onboarded before OAuth (no password stored) fall back to the legacy native
+api_key:api_secret (`Authorization: token <api_key>:<api_secret>`) - admin
+accepts both during the migration window.
 
 Guest calls (signup, get_plans) skip the header entirely; their admin
 endpoints are @frappe.whitelist(allow_guest=True).
 """
 
 import re
+import time
 
 import frappe
 import requests
@@ -24,6 +30,25 @@ from jarvis.exceptions import (
 # Admin's provision_healthz_timeout_s defaults to 60s for restart operations;
 # 90s leaves 30s buffer for network round-trip + handler overhead.
 DEFAULT_TIMEOUT_S = 90
+
+# OAuth password-grant config. The bench exchanges the customer's password
+# (Jarvis Settings.jarvis_admin_customer_password) for short-lived bearer
+# tokens against admin's native Frappe OAuth token endpoint, using the shared
+# public client id. Authenticated calls prefer the bearer; calls fall back to
+# the legacy api_key:api_secret when no password is stored (pre-OAuth
+# customers, dual-auth migration window).
+_OAUTH_CLIENT_ID = "jarvis-bench"
+_OAUTH_TOKEN_PATH = "/api/method/frappe.integrations.oauth2.get_token"
+_OAUTH_SCOPE = "all openid"
+# Site-scoped Redis cache for the access/refresh tokens (frappe.cache() is
+# per-site). Admin credentials are a per-site singleton, so one key suffices.
+_OAUTH_CACHE_KEY = "jarvis:admin_oauth_token"
+# Re-mint this many seconds before the cached access token's stated expiry so
+# a request can't race past the boundary.
+_OAUTH_EXPIRY_SKEW_S = 60
+# Upper bound the cache entry lives, so the refresh token outlives the
+# (~15min) access token; on entry expiry we re-mint with the password grant.
+_OAUTH_CACHE_TTL_S = 24 * 60 * 60
 
 # Cap on the cross-boundary message length. Long messages (e.g. a Frappe
 # 500 with a 10KB traceback that happens to embed a token mid-frame) get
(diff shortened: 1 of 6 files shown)


venkat102:
## Verdict

**Approve.** Strong pairing with admin PR Aerele-RnD/jarvis-admin#139 — dual-auth migration is honored, caching design is sound with explicit invariants, the hardening pass (commit `38e891f`) caught real edge cases before merge (cache eviction on persistent 401, zero-expiry storm prevention, broad `RequestException` catch). One concrete follow-up worth tracking; everything else is observation.

## Strengths

1. **Token caching design is careful.** Skew (60s) catches the boundary, refresh-token-first then password-grant fallback (never re-uses the password unnecessarily), 24h cache TTL upper-bounds Redis residency, zero-`expires_in` token explicitly NOT cached to prevent endpoint storm.

2. **The "harden per code review" commit shows real review hygiene.** Each item in `38e891f` is a real issue:
   - Evict cached token after both bearer attempts fail → prevents a poisoned token from replaying for the full 15min TTL after a revoke
   - Broad `requests.RequestException` catch in `_oauth_token_request` (SSL errors, redirect loops) → honors the "never raises" contract
   - Zero-expiry token guard → defends the token endpoint from being hit on every call
   - `strip()` symmetry on `api_secret` → trailing-newline silent break is a real bug class
   - `start_signup` persists email even when api_key/api_secret absent → forward-compatible with future admin response shapes

3. **Secret scrubbing at the boundary is comprehensive.** `_scrub_secrets` covers `api_key=`/`Bearer`/etc. patterns, `sk-` prefixes for OpenAI/Anthropic shapes, RFC 7519 JWTs, plus a 500-char message cap so a 10KB Frappe traceback can't blow up `last_sync_status` (a Data field). Six tests pin each pattern.

4. **The regression test `test_api_key_decrypted_via_get_password_not_attribute` (lines 142-170) is gold.** Encodes the "attribute access returns `*****` masked placeholder; only `get_password()` decrypts" rule as a test instead of just a code comment. Will catch a future refactor that reintroduces the bug.

5. **Migration honors dual-auth correctly.** `_post` prefers bearer when password is stored; falls back to api_key:api_secret when not. Re-mint on 401, then drop cache, then legacy. Customers onboarded before OAuth land cleanly on legacy without operator action.

6. **Test split is sane.** Bench mocks all HTTP (fast, deterministic, exercises the cache/header/error-routing logic); admin Aerele-RnD/jarvis-admin#139 has the actual integration test (`test_oauth_token.py::TestOAuthTokenIssuance`) that hits the real Frappe oauthlib endpoints. Bench tests the wire/caching machinery; admin tests the handshake. Right split.

## One concrete follow-up worth tracking

**No test for the refresh-token-grant path.** The code at `_admin_access_token:471-475`:

```python
refresh = cached.get("refresh_token")
if refresh:
    token = _oauth_token_request(admin_url, {
        "grant_type": "refresh_token", "refresh_token": refresh,
    })
if not token:
    token = _oauth_token_request(admin_url, {
        "grant_type": "password", "username": username, "password": password,
    })
```

The tests cover: bearer reuse (`test_caches_access_token_across_calls`), legacy fallback (`test_falls_back_to_legacy_without_password`), 401 re-mint then legacy (`test_bearer_401_remints_then_falls_back_to_legacy`), zero-expiry not cached (`test_zero_expiry_token_is_not_cached`). None of them exercise the "cached access_expires_at is past skew, refresh_token is set, expect grant_type=refresh_token" path.

Add:

```python
def test_uses_refresh_token_when_access_expired_but_refresh_present(self):
    # Cache an expired access token with a refresh token
    frappe.cache().set_value(admin_client._OAUTH_CACHE_KEY, {
        "access_token": "STALE", "refresh_token": "RT-good",
        "access_expires_at": time.time() - 60,
    })
    cap = {}
    token_resp = _mock_response(200, json_body={
        "access_token": "ACCESS-2", "refresh_token": "RT-good-2",
        "expires_in": 900,
    })
    with patch("requests.post", side_effect=self._route(
            token_response=token_resp, api_capture=cap)):
        admin_client.get_connection()
    # Mint via refresh_token, NOT password (which would expose the cred).
    grants = cap["token_calls"]
    self.assertEqual(grants[0]["grant_type"], "refresh_token")
    self.assertEqual(grants[0]["refresh_token"], "RT-good")
    self.assertNotIn("password", grants[0])
```

Without that test, a future refactor could silently break refresh-first (e.g., reordering to always-password) and we'd only see it as "we re-use the password every 15min instead of every 24h" — a real privacy regression that wouldn't show as a test fail.

## Smaller items (track but don't block)

### Architecture / brittle assumptions

- **`_OAUTH_CLIENT_ID = "jarvis-bench"` and `_OAUTH_SCOPE = "all openid"` are hardcoded on both sides.** Admin pins them via `BENCH_OAUTH_CLIENT_ID` in `oauth_setup.py`; bench pins them here. A typo on either side means bench falls back to legacy silently (if legacy present) or 401-loops (if not). A bench-side smoke test that does one real `_admin_access_token()` against the configured admin at install/upgrade time would catch the drift. Out of scope for this PR.

- **`_OAUTH_CACHE_TTL_S = 24h` upper-bounds Redis residency.** When the 24h cache TTL expires while a refresh token might still be valid much longer (Frappe's `validate_refresh_token` checks only existence + Active), the next call re-mints with the password grant — re-using the customer's password approximately every 24h. This is intentional belt-and-suspenders against Redis bloat from long-lived cache rows, but it's worth a code comment naming the tradeoff. Bench operators paranoid about password-on-the-wire frequency could lower this.

- **`Jarvis Settings.jarvis_admin_customer_password` is `read_only: 1`.** Correct posture (operator can't accidentally clear the persisted creds via UI), but means manual rotation requires `bench shell` + `db_set`. Worth a future "force re-mint" button for the operator-paranoid rotate path.

### Concurrency / load

- **No de-dup on concurrent mints near expiry.** Under high load with multiple RQ workers at exactly the expiry boundary (skew miss + 60-sec window), all lanes will independently re-mint and storm the token endpoint. Admin handles 4-5 parallel mints fine, but a `frappe.cache().lock("jarvis:admin_oauth_token:lock", ...)` around `_admin_access_token` would be defensive. Not blocking — single-process per site is the dominant case. Track as a P2 follow-up.

### Test scaffolding

- **Cross-reference to integration test.** `test_admin_client.py`'s docstring says "All HTTP is mocked." Worth adding "Real OAuth handshake against Frappe's oauthlib is covered by `jarvis_admin/tests/api/test_oauth_token.py`" so a future reader knows where to look.

- **No negative test for `_oauth_token_request` exhausting all paths.** If both refresh and password grant fail (admin down, customer password rotated and not yet propagated, etc.), `_admin_access_token` returns `None` and `_post` falls through to legacy. That's tested via `test_bearer_401_remints_then_falls_back_to_legacy` indirectly, but a direct test of "refresh fails AND password grant fails AND legacy present → legacy succeeds" would lock the contract.

### Doc / observability

- The `Jarvis Settings` doctype description for `jarvis_admin_customer_password` is good (mentions client_id `jarvis-bench`, mentions pre-OAuth fallback). Worth one line about "set by onboarding's verified poll on the verify-on path." Minor doc polish.

## What's not in this PR (correct scoping)

- Admin-side OAuth client registration + signup-time password issuance (admin PR Aerele-RnD/jarvis-admin#139, paired).
- Backfill of passwords for existing customers (separate patch on admin).
- Deprecation/removal of `api_key:api_secret` (deferred to after all benches migrated).
- The customer-facing "force re-mint" UI / rotation flow.

## Suggested next actions

1. **Add the refresh-token-grant test.** ~10 lines. Locks the "prefer refresh, fall back to password" contract.
2. **(Optional) Comment the 24h cache TTL rationale.** One line in `admin_client.py` near the constant.
3. **Confirm admin PR Aerele-RnD/jarvis-admin#139 lands first.** Bench's bearer path needs admin's OAuth client registered, password set on the customer's User. Until admin lands and runs the v1_9 patch, bench transparently uses legacy fallback — so no broken sequence.

Approving with these as track-able follow-ups.

PR #140, opened 2026-06-22
feat(tools): query v0.2 - DISTINCT, OFFSET, COUNT(DISTINCT), EXISTS/NOT EXISTS

venkat102:
## Summary

v0.2 of the qb-based `query` tool, closing the four gaps confirmed worth fixing proactively:

- **DISTINCT** (top-level) - \`{distinct: true}\` emits SELECT DISTINCT
- **OFFSET** (top-level) - \`{offset: N}\` with 0..100_000 ceiling
- **COUNT(DISTINCT field)** - new \`{distinct: true}\` modifier on the aggregate spec; applies uniformly to all five aggregates
- **EXISTS / NOT EXISTS** - new predicate operators with stripped sub-spec (from/alias/joins/where only) and \`{"\$field": "alias.col"}\` correlated-reference marker

Sub-spec recursion is capped at 2 levels (\`_MAX_SUBSPEC_DEPTH\`). Sub-specs that include select/group_by/having/order_by/limit/offset/distinct are rejected at validation time - subqueries don't need them.

Permission enforcement (via \`Engine.get_permission_conditions\`) and the row guard apply to outer queries as before. Sub-queries reference doctypes that have already been DocType-level perm-checked at the outer level.

## Test plan

- [x] \`bench run-tests --module jarvis.tests.test_query\` - all 49 tests pass (29 v0.1 + 20 new)
- [ ] Live smoke once plugin v0.0.9 + persona v0.38.3 land:
  - DISTINCT: \`query({from: "Customer", select: ["territory"], distinct: true})\`
  - OFFSET: \`query({from: "Sales Invoice", select: ["name"], limit: 5, offset: 5})\`
  - COUNT(DISTINCT): \`query({from: "Sales Invoice", alias: "si", select: [{agg: "count", field: "si.customer", distinct: true, as: "n"}]})\`
  - NOT EXISTS: replay timesheet "who hasn't filed" prompt, expect NOT EXISTS form

Plugin and persona PRs follow.

### DIFF
--- a/jarvis/tests/test_query.py
+++ b/jarvis/tests/test_query.py
@@ -640,3 +640,172 @@ class TestQueryV02Additions(FrappeTestCase):
 						"value": {"alias": "df"},  # missing 'from'
 					}],
 				})
+
+
+class TestQueryV03ExistsSubquerySecurityHardening(FrappeTestCase):
+	"""Verify EXISTS / NOT EXISTS sub-spec doctypes are subjected to
+	the SAME three permission gates as the outer query:
+
+	1. ``has_permission(dt, ptype="read")`` role gate
+	2. Per-site DocType allowlist gate
+	3. ``Engine.get_permission_conditions`` woven into the sub-query
+	   WHERE (User Permissions / DocShare / if_owner / hooks)
+
+	Pre-fix, ``_collect_doctypes`` walked only outer FROM + joins and
+	``_build_exists_criterion`` weaved no permission conditions on the
+	sub-query, opening a side-channel: caller could probe existence of
+	rows the operator was not allowed to read."""
+
+	def test_subspec_doctype_subjected_to_has_permission(self):
+		"""Role gate denies sub-spec DocType → PermissionDeniedError."""
+		# has_permission returns True for the outer doctype, False for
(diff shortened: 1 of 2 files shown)


venkat102:
## Security follow-up (commit 8ba3c46)

Pre-fix the EXISTS / NOT EXISTS sub-spec bypassed all three permission gates on the sub-spec's DocType (role gate, allowlist gate, and the \`Engine.get_permission_conditions\` weave). That opened an existence side-channel: a caller with role-read but a User Permission restricting which records they could see could probe existence over the full table.

Both layers fixed:

- \`_collect_doctypes\` now recurses into sub-specs (nested EXISTS included), so the caller's existing iteration covers role + allowlist gates for sub-spec doctypes too.
- \`_build_exists_criterion\` instantiates an Engine for the sub-query and ANDs \`get_permission_conditions\` into the sub-query WHERE for each sub-spec-local doctype.

5 new tests in \`TestQueryV03ExistsSubquerySecurityHardening\` cover both layers + the recursive doctype collection. 54/54 green.

This collapses the planned v0.2.1 into PR #140 since the leak is in unmerged code — no separate release tag needed.

PR #149, opened 2026-06-23
feat(chat): /jarvis SPA — render & produce every OpenClaw return type + settings console

l0gesh29:
## What this does

Replaces the old Desk chat with the **/jarvis frappe-ui SPA** and makes that chat **render and produce every OpenClaw return type**, plus a small OpenClaw-console-style settings surface.

The agent can now hand the customer documents, read what they attach, and draw charts — all inline in the conversation — and the operator gets a settings panel with dark mode and a usage view.

### Chat surface (SPA)
- Retires the Desk chat page and its dead code; ships `/jarvis` as a Vue 3 + Vite SPA (session **or** token auth, socketio realtime, `jarvis.chat.api.*`).

### Render every return type inline
- **html / svg** → sandboxed `<iframe srcdoc>`
- **pdf** → browser PDF viewer
- **image** (png/jpg/gif/webp) → ``
- **file** (xlsx/csv/…) → download card
- ` ```mermaid ` fences → lazy-loaded **SVG diagrams** (mermaid in its own chunk); other fences → `<pre>`
- Canvas detection/persistence generalized for all of the above **including artifacts in canvas subdirectories** (the old regex stopped at the first `/`).

### Generation + reading tools (Frappe-native)
- `export_excel` — real `.xlsx` from tabular data (`make_xlsx`) → private File → download card.
- `download_pdf` — fixed the v15 `print_format=` param; clean artifact title.
- `read_file` — reads an **attached** file gated by the caller's File permission: xlsx/csv → rows, pdf → text, text → decoded, image → metadata. Fixes the agent flailing on attachments (it used to try the browser tool on a private URL).
- Tool results carrying `{file_url, filename}` auto-attach to the in-flight assistant message's canvas and render inline.

### Settings console
- Gear → **Settings** with **Overview / Activity / Appearance** tabs.
- **Dark mode** (Light / Dark / System, persisted per device; palette is now a reactive CSS-var map).
- **Token usage** meter (clearly an *estimate* — OpenClaw's gateway doesn't emit real per-turn counts, so it's summed from message text against the monthly budget).
- Delete-this-chat; recent tool-run activity.
- Live tool activity now shows **one running tool at a time** + a compact "N tools done" tally instead of a row per call.
- Dropped the "ERPNext ·" prefix from the model pill.

## Verification
- Backend `read_file` verified for **xlsx (rows), pdf (text), image (metadata)**, and **end-to-end through the real chat** (agent called `jarvis__read_file` and returned the sheet rows).
[...]

PR #154, opened 2026-06-25
feat(chat): /jarvis UI overhaul — return types, artifact panel, action cards, confirm/auto-apply, sidebar + settings

l0gesh29:
## What

A broad overhaul of the `/jarvis` chat UI and the backend endpoints behind it. Builds on the canvas / return-type work in #149.

### Return types — render every type cleanly
- **Charts** — colorful, theme-aware mermaid (vibrant pie/flow palette that tracks light/dark), hover **download as PNG**, concurrency-safe render with retry, and a **shimmer skeleton** so raw markup never flashes.
- **Files** — produced via `export_excel` / `download_pdf` land as a **private File on the user's own site** and show as a download card; no container paths or raw URLs leak into prose.
- **Tables** — real pipe markdown.

### Artifact preview panel (ChatGPT / Claude style)
- A PDF/Excel card opens a **right side-panel**: PDF in a viewer (real file URL → renders inline in Chrome), images inline, **xlsx/csv rendered as a table** via the new `preview_file` endpoint.

### Action cards — the agent decides when
- A fenced ` ```jarvis-action ` JSON block renders as a **doc card** (Confirm & create / Edit / Discard) or an **email card** (Send / Copy / Regenerate); the block is stripped from the visible prose. Simple ` ```confirm ` bar as a fallback.
- **Confirm before changes** — a per-site `auto_apply_changes` setting (default off = confirm every create/update/submit/cancel/delete/amend). The worker folds it into the turn `[Context: …]`; on = auto mode, applies directly.

### Sidebar & settings
- Per-chat **⋯ menu** with inline **Rename** (`rename_conversation`) + **Delete**; the whole user card is clickable.
- Settings moved into the user card and reworked into a **centered modal** with a left section nav (General / Appearance / Activity).

### Header & load
- Quick **theme toggle** (sun/moon), an external-link **Desk** icon, and an **Auto** model option.
- **No flash on refresh** — theme painted in `<head>` before mount, app fade-in, and a `booting` flag so the welcome screen doesn't flash before the open chat loads.

## Commits (feature-wise)
1. per-site auto-apply (skip-confirmation) setting
2. inline file-preview endpoint for the artifact panel
3. rename conversations
4. /jarvis UI overhaul — charts, artifact panel, action cards, sidebar + settings
5. smoother page load — theme preload + fade-in

## Companion change
Agent behaviour (emit `jarvis-action` / mermaid, never paste paths, confirm CRUD) lives in **jarvis-persona** — sibling PR linked in the comments.

## Testing
Verified on the local stack: a single reply with table + colorful mermaid + Excel download card + PDF preview; the confirm card stops before `create_doc` and auto-mode runs it; `preview_file` returns sheet rows.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -10,6 +10,27 @@
 			href="https://fonts.googleapis.com/css2?family=Inter:wght@400;450;500;550;600;700&display=swap"
 			rel="stylesheet"
 		/>
+		<!-- Paint the saved theme BEFORE the app mounts so a refresh doesn't flash
+		     white (FOUC) on the way to dark, and fade the app in smoothly. -->
+		<style>
+			html, body { margin: 0; background: #ffffff; }
+			html.jv-preload-dark, html.jv-preload-dark body { background: #16161a; }
+			#app { animation: jv-appin 0.24s ease both; }
+			@keyframes jv-appin { from { opacity: 0; } to { opacity: 1; } }
+		</style>
+		<script>
+			(function () {
+				try {
+					var t = localStorage.getItem("jarvis-theme") || "system"
+					var dark =
+						t === "dark" ||
+						(t === "system" &&
+							window.matchMedia &&
+							window.matchMedia("(prefers-color-scheme: dark)").matches)
+					if (dark) document.documentElement.classList.add("jv-preload-dark")
+				} catch (e) {}
+			})()
+		</script>
 	</head>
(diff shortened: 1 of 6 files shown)


l0gesh29:
Companion persona change (agent emits the matching output formats): Aerele-RnD/jarvis-persona#84

PR #177, opened 2026-07-01
fix(settings): don't require llm_api_key on an unconfigured fresh site

kavin-114:
## Problem

On a freshly installed / reset `site.jarvis`, **any** save of `Jarvis Settings` throws:

> API-key auth mode requires llm_api_key

`Jarvis Settings.validate()` → `_validate_auth_mode_requirements()` reads `llm_auth_mode`, which **defaults to `api_key`**, and demands a key — even when no LLM has been configured yet and the save is unrelated to LLM config (e.g. toggling **Enable Sandbox Mode** during onboarding).

Net effect: a customer on a fresh site can't save `Jarvis Settings` at all, so onboarding can never enable sandbox mode and never starts. Under the unified LLM config (#155) the `models` table is the source of truth and starts empty, so the legacy `llm_auth_mode` mirror's `api_key` default becomes a hard blocker.

## Fix

`jarvis/jarvis/doctype/jarvis_settings/jarvis_settings.py`

- Add `_llm_is_configured()` — true when any of `models` / `preset` / `llm_model` / `llm_oauth_account_email` is set.
- `_validate_auth_mode_requirements()` now early-returns when the tenant is unconfigured (pre-onboarding). The api-key requirement is still enforced the moment an LLM is actually configured (a model / preset / oauth account is present), so no real misconfiguration slips through.

This treats "no LLM configured yet" as a valid pre-onboarding state, consistent with the unified-config design where onboarding is the canonical path that writes `models[0]`.

## Tests

- New regression `test_unconfigured_fresh_settings_does_not_require_api_key` (`jarvis/tests/test_settings_on_update.py`) — a bare `Jarvis Settings` validates without a key.
- Existing `test_api_key_mode_requires_api_key` still passes: a *configured* api-key model without a key still throws.

```
bench --site <site> run-tests --module jarvis.tests.test_settings_on_update   # 34 passed
bench --site <site> run-tests --module jarvis.tests.test_unified_llm_config    # 32 passed
```

## Context

Found during a from-scratch onboarding e2e on a reinstalled site, on top of the unified LLM config (#155). Companion fix for the same onboarding flow: #178 (OAuth-connect doctor ordering).

### DIFF
--- a/jarvis/jarvis/doctype/jarvis_settings/jarvis_settings.py
+++ b/jarvis/jarvis/doctype/jarvis_settings/jarvis_settings.py
@@ -70,12 +70,37 @@ class JarvisSettings(Document):
 
         self._validate_auth_mode_requirements()
 
+    def _llm_is_configured(self) -> bool:
+        """True once the tenant has actually chosen an LLM.
+
+        Configured via ANY of: the unified ``models`` table, a ``preset``, a
+        legacy direct ``llm_model``, or a connected OAuth account. A fresh
+        site with none of these is *unconfigured* - LLM gets set up later,
+        through onboarding - so credential validation must not fire yet.
+        """
+        return bool(
+            getattr(self, "models", None)
+            or getattr(self, "preset", None)
+            or (getattr(self, "llm_model", None) or "")
+            or (getattr(self, "llm_oauth_account_email", None) or "")
+        )
+
     def _validate_auth_mode_requirements(self):
         """Each auth mode requires its own credential field.
 
         REV-1: oauth/subscription mode has no bench-side credential
         requirement - openclaw owns the credential blob on the container.
+
+        Skipped entirely for an unconfigured/pre-onboarding Settings: on a
+        brand-new site ``llm_auth_mode`` DEFAULTS to ``api_key`` before any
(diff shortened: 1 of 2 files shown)


kavin-114:
Addressed the xhigh code-review findings in `4618d33`:

- **Keyless-direct bypass (CONFIRMED):** the gate now keys the legacy-direct "configured" check off `llm_model`/`llm_base_url`/`llm_oauth_connected_at`, so a keyless custom-endpoint config is still validated. Added a negative regression test.
- **Disabled-only table / bare preset / decrypt-error spurious throws (CONFIRMED):** the flat api_key check now **skips entirely** when the models table or a preset is present — `validate_models()` (in `on_update`) owns validation there, so it no longer races `before_validate`'s enabled-only mirror. Added a unified-path-skip test.
- **Dead oauth field (PLAUSIBLE):** switched to `llm_oauth_connected_at` (the signal `is_ready_for_chat` uses).
- **Test pollution (CONFIRMED):** clear `models`/`preset`/`oauth` in `setUp` and clean them in `tearDownClass`.

Suite: `test_settings_on_update` 36, `test_unified_llm_config` 32.

PR #205, opened 2026-07-04
AI write-safety: server-enforced confirmation gate for agent ERP writes (issue #186)

venkat102:
Closes the highest-severity finding in issue #186 (SEC-006/007/009): the agent's 'confirm before writing' rule lived only in persona Markdown, so a prompt injection in an ERP record/attachment/skill could drive an unconfirmed create/update/submit/cancel/delete/amend/run_method/send_email, and `set_auto_apply` was an ungated site-wide kill-switch. This makes confirmation a **bench-enforced invariant**.

## What it does
- **The gate** (`_run_tool` chokepoint): the 8 gated writes NEVER execute on the model path. Each parks a single-use token (atomic GETDEL, bound to conversation+owner+args-hash), the bench publishes an `action:pending` realtime event carrying the token to the UI (never to the model), and the write executes only via `confirm_tool(token)` on a human click. `dispatch_confirmed` is the sole confirmed executor and is not web-exposed. Gated tools also refuse model-initiated `preview=true` (closing an inline-side-effect bypass), and `run_method` parks with a described preview rather than a sandbox execution.
- **Auto-apply reshaped**: per-conversation, admin-gated to enable (API + a doctype-layer controller validate so a generic write path can't flip it), and it fast-paths ONLY reversible create/update - submit/cancel/delete/amend/run_method/send_email always confirm.
- **apply_action converged** to the human-edit path only (create/update, mandatory owner+conversation, audited); the confirm-as-proposed verbs go through the token gate.
- **Untrusted-data fence** on extracted-file text (breakout-hardened: trusted File-doc name, label sanitize, attribute/self-closing/homoglyph close-tag neutralization) - a probability-reducer layered under the gate.
- **Frontend**: an `action:pending` Confirm card wired to `confirm_tool`; retired the model-visible approval paths; per-conversation admin-gated auto-apply toggle.

## Review
Subagent-driven, TDD per task, a task review after each, and a whole-branch security review on the most capable model: **end-to-end invariant HOLDS across all traced seams** - no gated ERP write commits without a human confirm click (or the admin-enabled, per-conversation, non-destructive auto-apply opt-in). Every Critical/Important finding was fixed and re-verified, including a Critical fence-bypass via crafted file name and two run_method findings caught in the final review.

## Testing
New suites green on jarvis-test.localhost: test_pending_confirm, test_confirm_gate, test_action_pending, test_auto_apply, test_untrusted_fence; regressions green: test_actions_api, test_chat_worker, test_chat_api. Both dispatch flows (RQ + Path B) traced; per-user permission wall unchanged.

## Ships in lockstep with
jarvis-persona v0.40.0 (the model must call submit/cancel/delete/amend/email directly so the bench gates them). Merge the persona PR together with this one.

## Documented fast-follows (non-blocking, security holds)
- Confirm card is realtime-only: a page refresh mid-confirm loses it until token expiry (candidate: a pending-confirmations list endpoint + resurface on resync).
- Self-host: a confirmed write runs under the operator browser session, not the scoped tool user.
- Fence covers attachment text; vision-extracted image text is not yet fenced.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -80,9 +80,17 @@ export const getDoctypeFormMeta = (doctype) => call(AC + "get_doctype_form_meta"
 export const loadDocForEdit = (doctype, name) => call(AC + "load_doc", { doctype, name })
 export const applyAction = (action) => call(AC + "apply_action", { action: JSON.stringify(action) })
 // Write-safety gate (issue #186): confirm a parked ERP write by its one-time
-// token. Returns the tool result envelope {ok, ...} on success, or
-// {ok:False, error:{type:"InvalidConfirmation", ...}} if the token is gone.
-export const confirmTool = (token) => call(AC + "confirm_tool", { token })
+// token. Pass the conversation the click came from so the server enforces the
+// real conversation guard (#11). Returns the tool result envelope {ok, ...} on
+// success, or {ok:False, error:{type:"InvalidConfirmation", ...}} if the token
+// is gone.
+export const confirmTool = (token, conversation) =>
+	call(AC + "confirm_tool", { token, conversation: conversation || "" })
+// Resync (issue #186, R3 fix for #3): re-surface the caller's own currently
+// parked confirmation cards after a reload/reconnect. Returns
+// {ok, data:{pending:[{token, tool, preview, summary, conversation, run_id}]}}.
+export const listPendingConfirmations = (conversation) =>
+	call(AC + "list_pending_confirmations", { conversation: conversation || "" })
 
 export async function sendMessage(conversation, message, modelOverride, attachments, context) {
 	// Empty conversation is allowed: the backend creates (or focuses) an empty
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -26,10 +26,12 @@ export const setStar = (conversation, starred) =>
 	call("jarvis.chat.api.set_star", { conversation, starred: starred ? 1 : 0 })
 export const retryMessage = (message) => call("jarvis.chat.api.retry_message", { message })
 export const getChatUiSettings = () => call("jarvis.chat.api.get_chat_ui_settings")
-// Toggle "auto-apply changes" (skip the agent's confirmation step before
-// mutating ERP data). Off = confirm every change (default).
-export const setAutoApply = (value) =>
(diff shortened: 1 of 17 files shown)


venkat102:
## Max-effort review remediation (post-open)

Ran `/code-review max` across both PRs as one integrated change: 31 agents, 27 candidates verified, **15 findings**. The **core security invariant held** (no gated write commits without a human confirm click) - the findings were a reliability + self-host correctness cluster the earlier review layers missed. All 15 are now fixed in three reviewed rounds:

**R1 (`197a66a`) - gate identity + receipt + resync:** bind the confirm card to the conversation OWNER (not the tool user) so it reaches the right browser in self-host; execute the confirmed write under the scoped model identity (never exceeds the model path's permissions); persist a transcript receipt for confirmed writes; make the conversation guard a real check; `preview=true` on a gated write returns a legible error instead of silently parking; new owner-scoped `list_pending_confirmations` resync endpoint.

**R2 (`fadbf0d`) - robustness:** coerce non-string attachment `file_name` (was a live TypeError crashing any turn with a malformed attachment); wrap `consume`'s `getdel` so a transient Redis blip returns a graceful "expired" instead of a 500 (single-use atomicity preserved); drop a now-tautological auto-apply comparison.

**R3 (`6bef397`) - frontend reliability:** queue multiple confirm cards (a second parked write no longer overwrites the first); re-surface parked confirmations on reload/reconnect via the resync endpoint; honor Stop; attach a conversation-less card to the active conversation; render a graceful note for legacy write/email cards during the persona rollout window.

Each round was TDD'd and independently reviewed (identity model hand-traced, invariant re-confirmed HOLDS). Regression sweep green: test_pending_confirm, test_confirm_gate, test_action_pending, test_auto_apply, test_untrusted_fence, test_actions_api, test_chat_worker, test_chat_api.

Residual (documented, non-blocking): self-host confirmation is undeliverable only when a turn's conversation cannot be resolved at all (fails closed - no unconfirmed write); pre-existing em-dashes in the large ChatView.vue predate this work.

venkat102:
Addresses **Aerele-RnD/jarvis-admin#186** (SEC-006/007/009 - AI write-safety). Ships in lockstep with the persona PR Aerele-RnD/jarvis-persona#107.

PR #208, opened 2026-07-04
fix(chat): relay-path context overflow parks for recovery — completes #206 for the production driver

Vigneshsekar:
Completes #206. The user hit the raw overflow error AGAIN after #206 because managed turns are driven by **`chat.send` + `relay_turn_events`**, whose failures surface as **`relay:error` terminals** — a different branch from the lifecycle-error path #206 patched. Gateway log confirmed the same recoverable pattern (error → `auto-compaction succeeded; retrying prompt` → retried run completes in the session) while the chat showed the dead error.

`relay:error` carrying a context overflow now parks the turn (`_mark_recovering` + `run:recovering`, same as `relay:interrupted`); recovery finalizes the retried answer; a retry that also dies hits the recovery ceiling and errors honestly.

**Plan-agnostic by design:** customers on $20/$100/$200 plans get different context windows, but openclaw derives the window from the model catalog and auto-compacts within it — smaller windows simply compact sooner. With both surfacing paths parked, no plan ever sees the raw overflow error; the visible behavior is a longer "working" state, then the answer.

Worker module 29/29.

### DIFF
--- a/jarvis/chat/turn_handler.py
+++ b/jarvis/chat/turn_handler.py
@@ -720,7 +720,28 @@ def handle_chat_send(payload: dict) -> None:
 				return
 
 			if terminal["kind"] == "relay:error":
-				_publish_run_error(terminal.get("error") or "agent error")
+				err_text = terminal.get("error") or "agent error"
+				# Context overflow is NOT terminal on openclaw: it emits the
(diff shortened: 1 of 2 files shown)


Vigneshsekar:
Depends on #200

PR #209, opened 2026-07-04
feat: File Box + Approvals — directed inbound-document entry

Vigneshsekar:
Companion to jarvis-persona#110 (`ocr-data-entry` skill). Born from the "Book Draft Loreal Invoice" failure (one balancing line, 4+ interrogation rounds).

## What lands
- **`Jarvis Approval` doctype** — a decision the agent needs a human for. The agent creates it itself via the generic `jarvis__create_doc` (no new plugin tool), links it via the new `conv:` id in the per-turn context line, and ends the turn. **Deciding posts the decision back into the linked conversation** (normal `send_message`), so the agent resumes the same chat with full context — "if needed, taken to the same chat interface" by construction.
- **`approvals_api`** — `list_approvals` (conversation-owner or System-Manager scoped), `pending_count`, `decide`.
- **File Box** (`jarvis.chat.filebox`) — `drop_file`: creates a conversation named after the uploaded file and sends ONE directed prompt (process via the ocr-data-entry skill; decide by convention; queue approvals; ≤1 question) through the existing send_message+attachments machinery. `list_inbound` derives status (processing / needs approval / done / error) from live chat state — no bookkeeping doctype to drift out of sync.
- **SPA** — two new rail buttons beside Skills/Macros: **File Box** (drag-drop zone + inbound list with status chips; dropping opens the processing chat) and **Approvals** (pending queue: question, collapsible context, option buttons + free-text, Approve/Reject; deciding jumps into the resumed chat; pending-count badge on the rail, refreshed with the conversation list).

## Verified
- Approval roundtrip smoke-tested (create → list → decide → status/decision recorded → cleanup).
- Both endpoints HTTP-verified as the SPA calls them.
- Frontend builds clean.
- Migration creates the doctype (both sites migrated).
- End-to-end validation with the original Loreal PDF: see the follow-up comment on this PR after the live run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #210, opened 2026-07-04
perf: dedicated jarvis_chat queue - parallel File-Box turns, no long-queue starvation

Vigneshsekar:
## Why
One in-flight chat turn = one occupied long-queue RQ worker for the turn's whole wall-clock (the worker relays the openclaw event stream; tool callbacks ride gunicorn). Drop 10 documents into the File Box on a default bench and they process serially - and every other long job (emails, migrations, reports) waits behind minutes-long turns.

## What
- `_turn_queue()`: agent turns enqueue on a dedicated **`jarvis_chat`** queue when the bench provisions it, else `long` exactly as today.
  - Dual gate: queue declared in `common_site_config.workers` (also required by frappe's `validate_queue`) **and** a live RQ worker listening on it - a declared-but-dead queue can never blackhole turns. Probe cached 60s per site.
  - `jarvis_chat_queue` site_config value overrides verbatim (`"long"` = force off).
- `at_front` only on `long`: on the all-chat dedicated queue it would invert batch drops to LIFO; drop order is preserved.

## Enabling (per bench, optional)
```json
"workers": {"jarvis_chat": {"timeout": 800, "background_workers": 4}}
```
then `bench setup supervisor` + reload (Frappe Cloud: bench worker configuration). Chat workers are I/O-idle relays - cheap in CPU, ~100-150MB each.

## Compatibility
Verified against frappe `version-15` and v16: `get_workers(queue=None)`, `generate_qname`, custom `workers` timeout config, and `frappe.cache()` semantics are identical on both. No schema, no frontend, single funnel `_dispatch_turn` touched (send_message, retry and macros all inherit).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,45 @@ chat at **`/app/jarvis-chat`**. Full walkthrough: **getting-started** (see Docum
 **Local single-bench dev** (run openclaw yourself, no control plane) → the
 **local-dev** guide (see Documentation).
 
+## Optional: parallel chat turns (dedicated worker queue)
+
+Each in-flight chat turn occupies one background (RQ) worker for the turn's
+whole duration, and turns run on the shared `long` queue by default — so on a
+bench with one long worker, multiple conversations (e.g. a batch of File Box
+documents) process **one at a time**, and other long-queue jobs wait behind
+them.
+
+To process turns in parallel and isolate them from the rest of the bench,
+declare a dedicated `jarvis_chat` queue in `common_site_config.json`:
+
+```json
+"workers": {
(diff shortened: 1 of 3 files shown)


Vigneshsekar:
Follow-up commit d02dbfb closes the adversarial-review findings: orphan-turn sweep in stale_scan (deterministic RQ job ids, payload salvage, one self-heal then error+retry surface), probe TTL 60s->10s, interactive-vs-background at_front (File-Box drops FIFO via #209's `background=1`), and site_config override validation. #209's latest commit depends on this PR.

PR #211, opened 2026-07-04
security: harden canvas relay against skill exfil (traversal + size cap)

Vigneshsekar:
## Why
A customer tried to make the agent zip and hand over its (proprietary) skills. Auditing the artifact-delivery path found the canvas relay would fetch and save to the user's site any `canvas/<path>.<ext>` reference the agent writes, gated only by a filename **extension allowlist**.

## Two holes closed in `jarvis/chat/canvas.py`
1. **Path traversal** — the path regex allows `.` and `/`, so `canvas/../skills/erpnext-accounts/SKILL.md` matched and would be fetched from the gateway and saved as a user download, regardless of whether the gateway normalizes `..`. `detect_canvas_names` now drops any traversal path (empty / `.` / `..` / all-dots segment, leading slash, backslash, NUL); `fetch_canvas` re-checks as a direct-call backstop. Unit-tested.
2. **No size/content bound** — `fetch_canvas` trusted the extension only (no magic-byte, no size cap), so bulk bytes staged as `canvas/x.xlsx` would be delivered. Now streams with a **25MB** per-artifact cap.

## Scope
Part of a defense-in-depth set: persona "skills are internal" refusal rule (jarvis-persona main), fleet native-tool deny trim (jarvis-fleet-agent#110), and this relay hardening. Network egress control and an optional plugin guardrail remain product decisions.

### DIFF
--- a/jarvis/chat/canvas.py
+++ b/jarvis/chat/canvas.py
@@ -24,6 +24,10 @@ MSG = "Jarvis Chat Message"
 
 # Cap how many artifacts we'll fetch per turn (defensive).
 _MAX_CANVAS_PER_TURN = 8
+# Per-artifact byte cap. Legit charts/PDFs/xlsx are well under this; the cap
+# bounds how much can leave via a file staged in canvas/ and named with an
+# allowlisted extension (the extension check alone trusts the filename).
+_MAX_CANVAS_BYTES = 25 * 1024 * 1024
 
 # Supported artifact extension -> render type.
 #   html/svg -> sandboxed iframe srcdoc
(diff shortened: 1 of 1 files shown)


Vigneshsekar:
Closing: the canvas traversal check here lives in the **jarvis app**, which ships to the customer's Frappe Cloud bench — editable source, so not a trustworthy boundary. And it turns out to be **redundant**: I tested the container gateway directly (raw, un-normalised `--path-as-is` request for `/__openclaw__/canvas/../workspace/skills/…/SKILL.md`) and it returns the SPA fallback page, not the file — the gateway (Aerele-controlled, customer-inaccessible) already confines the canvas endpoint to its own directory, so skills are unreachable through it even with the gateway token. The real skill-exfil controls belong in the internal layers (fleet tool-deny, persona refusal rule, and a container-side guardrail + egress control tracked as an issue), not in the editable customer app. The 25MB fetch size cap was minor hygiene, not security; can re-add standalone later if wanted.

PR #223, opened 2026-07-06
feat(chat): auto-continue multi-step plans after Apply/Confirm

venkat102:
## Problem

Asking Jarvis to do a multi-write task ("create a timesheet" = create N Tasks, then a Timesheet against them, then submit) stalls after the first step. Root cause on the bench side: after a human clicks Apply (draft card) or Confirm (gated write), `apply_action` / `confirm_tool` only insert a passive receipt; no agent turn ever follows an approval, so the agent cannot continue its plan until the user types "continue".

## Change

- **`enqueue_continuation` (chat/api.py)**: after a human Apply/Confirm, dispatch a follow-up agent turn through the existing macro-engine path (`_enqueue_turn` -> `_dispatch_turn`), so **both dispatch flows (RQ and Path B)** are covered identically, including the NUMSUB fallback guard. The prompt is a hidden user message: `[System] Applied: <receipt>. Continue the remaining steps...` - it carries the new record's name, which the agent needs for dependent steps (Timesheet rows referencing just-created Tasks).
- **`apply_action`**: accepts `continue:1` on the card payload (the model marks non-final plan steps; the SPA forwards it). Without the flag, behavior is unchanged - instant receipt, no LLM cost, no latency regression for single creates.
- **`confirm_tool`**: always dispatches the continuation after a confirmed gated write - the model only ever saw "awaiting the user's confirmation", so this is also what finally shows it the real outcome (including a bounded error message on failure).
- **`Jarvis Chat Message.hidden`** (Check): internal rows feed the agent transcript but never render; `get_conversation` filters them, covering first load and resync reloads.
- Continuation turns skip the request-path session bootstrap (the worker creates a missing `session_key` on its pooled connection, turn_handler.py:640), so the human's click POST never pays or fails on a WS handshake.
- Safety: a continuation only ever fires from a human click inside the caller's own owner-checked conversation; it grants no write authority and the #186 gate semantics are unchanged - the human remains the rate limiter, no autonomous loop path.

## Tests

- `test_actions_api` grew a `TestContinuation` class: apply with flag -> receipt + hidden `[System] Applied:` user row + one dispatch; apply without flag -> receipt only, no dispatch (regression pin); confirm_tool -> continuation dispatched; hidden rows excluded from `get_conversation`.
- Green on jarvis-test.localhost: test_actions_api (19), test_confirm_gate (21), test_chat_api (5), test_auto_apply (20), test_action_pending (4). SPA `vite build` passes.

## Live smoke checklist (post-merge, needs the persona PR + container restart)

- Fresh chat: "Create a timesheet for last week: tasks A, B, C on project P" -> plan statement -> Task card -> Apply -> agent auto-continues through Tasks 2, 3 -> Timesheet card -> gated submit confirm -> final one-line ack. Zero typed "continue".
- Regression: "create a customer X" -> card -> Apply -> instant receipt, no extra LLM turn.
- Auto-apply ON conversation: the same ask completes in one turn.
- Run on BOTH dispatch flows (RQ and Path B).

Companion persona PR: Aerele-RnD/jarvis-persona (multi-step plan discipline + timesheet workflow).

### DIFF
--- a/frontend/src/views/ChatView.vue
+++ b/frontend/src/views/ChatView.vue
@@ -1964,6 +1964,10 @@ async function openDraftPanel(a) {
 	draftPanel.value = {
 		verb, doctype: a.doctype, docName: verb === "update" ? (a.name || "") : "",
 		title: a.title || "", submittable: !!meta.is_submittable,
+		// Multi-step plans: the card marks non-final steps "continue": 1; the
+		// bench then dispatches a follow-up agent turn after Apply so the agent
+		// stages the next step without the user typing "continue".
+		cont: a.continue ? 1 : 0,
 		fields, tables, applying: false, error: "", updatedToast: false,
 	}
 }
@@ -2090,6 +2094,7 @@ async function applyDraft(submitFlag) {
 		await api.applyAction({
 			verb: p.verb, doctype: p.doctype, name: p.docName || "",
 			values, submit: submitFlag ? 1 : 0, conversation: currentId.value || "",
+			continue: p.cont ? 1 : 0,
 		})
 		closeDraftPanel()
 		await loadConversation(currentId.value)
--- a/frontend/src/views/ChatView.vue
+++ b/frontend/src/views/ChatView.vue
@@ -342,7 +342,7 @@
 						</div>
 						<div v-if="pa.error" class="jv-draft-error" style="margin:0 14px 10px">{{ pa.error }}</div>
 						<div class="jv-action-foot">
-							<button class="jv-action-primary" :disabled="pa.busy" @click="confirmPending(pa)">✓ Confirm</button>
+							<button class="jv-action-primary" :disabled="pa.busy || convStreaming" :title="convStreaming ? 'Waiting for the current reply to finish' : ''" @click="confirmPending(pa)">✓ Confirm</button>
 							<button class="jv-action-discard" :disabled="pa.busy" @click="discardPending(pa)">Discard</button>
 						</div>
 					</div>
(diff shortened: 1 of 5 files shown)


venkat102:
## Self-review applied (commit pushed)

Ran a multi-lens adversarial review of this PR and verified the load-bearing findings against source. Two things fixed in the new commit; the rest are noted follow-ups.

### Fixed
1. **Injection framing (the one worth blocking on).** The continuation prompt interpolated the receipt inline next to a `[System] Applied:` marker the persona treats as an authoritative system line. The receipt carries attacker-influenceable text - a record `name` under field/prompt autoname, or a DocType error echoing a field value - and on the **update** path that name can belong to a record *another* user named, making it a cross-user channel to dress record data up as a system directive (the exact pattern #186's untrusted-data discipline exists to stop; the gate still blocks any unconfirmed write, but the plan could be steered). Fix: the receipt is now neutralized like the attachment seam's out-of-fence file-name label (`_safe_label_name`: one line, backticks disarmed) and quoted as inline-code **data**. A real `<untrusted-data>` fence can't be used here - the `content` field's HTML sanitizer strips the unknown tag and silently breaks the fence (verified). New tests pin the breakout neutralization on the apply and FAILED-confirm paths.
2. **Mid-turn confirm race.** The `action:pending` card is published while the parent turn is still streaming; Confirm was gated only on its own busy flag, so a click dispatched a continuation turn concurrent with the parent. Confirm is now also disabled while the conversation is streaming (`store.streamingConvId`), with a defense-in-depth early-return.

### Noted, not changed here (follow-ups)
- `confirm_tool` dispatches a continuation **unconditionally**, so every gated write now spends an LLM turn (correct for plans; a cost change for one-off confirms) - flagging as a deliberate design call.
- `_next_seq` is a lockless `MAX(seq)+1` with no unique index on `(conversation, seq)` - a **pre-existing** race this PR makes easier to reach, not introduced here.
- Continuations skip `validate_can_send` (subscription/credit gate).
- Telemetry: continuation omits `enqueued_at_ms` (`queue_wait_ms` logs -1).

### Verified clean
Owner-binding holds, no autonomous-loop path (continuation only fires from a human click in the caller's own conversation), hidden rows aren't leaked to the browser, and the `hidden` Check migrates as `NOT NULL DEFAULT 0` (no patch, no legacy NULL).

Tests: test_actions_api (21), test_confirm_gate (21), test_chat_api (5) green on jarvis-test.localhost; SPA build passes.

PR #237, opened 2026-07-07
Write-UX Phase 1: summary-first confirmation card

venkat102:
## Phase 1 of the write-interaction UX: summary-first confirmation

First of five phases from the design spec (`docs/superpowers/specs/2026-07-07-jarvis-write-interaction-ux-design.md`) and plan (`docs/superpowers/plans/2026-07-07-jarvis-write-interaction-ux.md`).

**The problem it solves:** a `jarvis-action` create/update used to auto-open a full editable form that mirrors the desk form, so users asked "why fill a form in Jarvis when I could just open the doctype." This makes a read-only **summary/diff card** the default surface instead.

### What changed

- **`buildDraftModel` extracted** from `openDraftPanel` (pure refactor) so the summary card and the editable panel share one construction path.
- **New pure helper** `frontend/src/lib/actionSummary.js` (+ `node:test`, 8/8) computes the create identifying-fields view, the update old -> new diff, and a line-item summary.
- **`.jv-summary` card** renders inline: create shows key fields + a line-item table with a total; update shows the field diff. **Confirm** applies through the existing single write path (`applyDraft`, parameterized) without opening the panel; **Edit** opens the existing editable panel unchanged.
- The editable panel **no longer auto-opens** on a fresh action (it still refreshes if already open mid-Edit).
- **Risk-tier cue:** the reversible summary card has a blue border vs the amber gated `.jv-pending` card.

### Design and theme fidelity

- Reuses the existing chat card primitives (`.jv-action`, `.jv-action-primary/-2nd`, `.jv-draft-error`) so it looks native.
- New CSS is **tokens-only** (`--surface`, `--text`, `--blue`, `--green`, `--border`, `--blue-bd`, ...) from `theme.js`, so it is **light/dark compatible** with no `.jv-dark` override needed.
- All model-proposed values render through escaped `{{ }}` interpolation (no `v-html`, no XSS surface).

### Verification done

- `node --test src/lib/actionSummary.test.js`: 8/8 passing, pristine.
- `@vue/compiler-sfc` parse + compileScript + compileTemplate on `ChatView.vue`: zero errors.
- Per-task spec+quality review on every task, plus a whole-branch final review: 0 Critical, 0 Important. One stale-summary WYSIWYG edge case found and fixed mid-review.

### Live smoke PENDING (please run before merge)

No headless browser test exists for the SFC, so the visual/behavior smoke is for you. In a **fresh chat on `jarvis-test.localhost`**, in **both light and dark**:

1. Create with line items ("create a sales order for Acme with 10 Widget A and 5 Widget B") -> summary card (not an auto-opened form) with an Items table + total; Confirm creates the record.
2. Update ("set that order's delivery date to next Friday") -> old -> new diff card; Confirm writes.
3. Edit -> the editable panel opens and Apply still works.
4. Repeat step 1 with `socketio_backend: python` (Path B) to confirm the card renders identically.

### Notes

- Fixed a plan-defect in the currency-total selection (the plan's verbatim code would have summed a unit `rate` column; corrected to prefer a named amount column).
- Deferred (pre-existing, inherited from the editable panel, not introduced here): a required-but-unproposed child table shows `count = 1` with a blank row.
- Phases 2-5 (read-only Preview, few-vs-many input, learning moment, persona) follow in their own PRs.

### DIFF
--- a/frontend/src/lib/actionSummary.js
+++ b/frontend/src/lib/actionSummary.js
@@ -21,6 +21,8 @@ export function changedFields(model) {
 }
 
 export function lineItemSummary(table) {
+  // Prefer a named amount column; else fall back to the first Currency column
+  // (which may be a unit rate - the total is display-only, not authoritative).
   const amountCol = table.columns.find((c) => AMOUNT_COLS.has(c.fieldname)) || table.columns.find((c) => c.fieldtype === "Currency")
   const total = amountCol
     ? table.rows.reduce((s, r) => s + (Number(r[amountCol.fieldname]) || 0), 0)
--- a/frontend/src/lib/actionSummary.js
+++ b/frontend/src/lib/actionSummary.js
@@ -1,0 +1,44 @@
+// frontend/src/lib/actionSummary.js
+// Pure display helpers for the summary-first confirmation card. No Vue, no DOM.
+
+const AMOUNT_COLS = new Set(["amount", "net_amount", "base_amount", "total"])
+
+export function identifyingFields(model, limit = 6) {
+  const nonEmpty = (f) => String(f.value ?? "").trim() !== ""
+  const seen = new Set()
+  const picked = []
+  const push = (f) => { if (f && nonEmpty(f) && !seen.has(f.fieldname)) { seen.add(f.fieldname); picked.push(f) } }
+  push(model.fields.find((f) => f.fieldname === model.titleField))
+  model.fields.filter((f) => f.reqd).forEach(push)
+  model.fields.forEach(push) // fill remaining non-empty up to limit
+  return picked.slice(0, limit).map((f) => ({ fieldname: f.fieldname, label: f.label, value: f.value }))
+}
+
+export function changedFields(model) {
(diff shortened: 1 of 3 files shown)


venkat102:
Updated per design discussion: summarization is now **model-driven** instead of a frontend heuristic.

- Dropped `identifyingFields` and the `AMOUNT_COLS` code-computed total (they only fit order/invoice-shaped doctypes, not Timesheets / Leave Applications / Journal Entries).
- The create card now renders **the fields the model proposed** (`action.fields`); child tables show a compact row summary with **no code total**.
- The `jarvis-action` block may carry an optional **`summary`** headline the model writes with doctype knowledge (totals go there). Proposed-field rendering is the always-on default, so the card still works if the model writes no headline.
- The update diff stays mechanical (old -> new). Headline renders as escaped interpolation (no XSS surface).

Persona rule to author the headline lands in Phase 5; the frontend degrades gracefully until then. Commits `ae60509` + `3c9c6d1`.

PR #252, opened 2026-07-09
fix(chat): revive the stale-pairing self-heal + adopt gateway-reissued device tokens

kavin-114:
## The regression

After the 2026-07-08 deployment, existing benches failed every chat connect with:

```
connect rejected: INVALID_REQUEST: unauthorized: device token mismatch (rotate/reissue device token)
```

and never recovered — a hard chat outage requiring manual `rotate_chat_device`.

## Root cause (verified by live reproduction)

openclaw sends connect **auth** failures with the generic `error.code = "INVALID_REQUEST"`; the machine-readable reason lives only in `error.details.authReason`. It also maps **every** explicit-deviceToken failure — device not paired (container replaced / state reset), token mismatch (gateway rotated the stored token), token revoked — to the single coarse reason `device_token_mismatch`.

The Sprint-3 classifier reads `error.code` and matches openclaw's *internal* hyphenated reasons (`device-not-paired`, `token-mismatch`, …) which never appear on the wire, so `_is_stale_pairing` never fired: **the designed two-shot self-heal (clear + re-pair) was dead code**, and the exact scenario it was built for (admin replaces the tenant container) became a permanent outage.

Verified against the openclaw image from 2026-06-16 **and** current `:latest` (bundle diff — behavior identical in both), plus a live reproduction against a running gateway; the wire frame in the test fixture is captured verbatim:

```json
{"code": "INVALID_REQUEST",
 "message": "unauthorized: device token mismatch (rotate/reissue device token)",
 "details": {"code": "AUTH_DEVICE_TOKEN_MISMATCH", "authReason": "device_token_mismatch",
             "canRetryWithDeviceToken": false, "recommendedNextStep": "update_auth_credentials"}}
```

## Fixes

1. **Classifier** — `OpenclawUnreachableError` now carries openclaw's `error.details`; `_is_stale_pairing` additionally classifies on `details.authReason == "device_token_mismatch"`. The internal-code set is kept as belt-and-braces; message-substring matching stays banned (pinned by a new negative test).
2. **Reissued-token adoption** — openclaw's hello-ok can carry a rotated `auth.deviceToken` (the gateway replaces the stored token at connect when the entry no longer lines up with the requested scopes/issuer; the rotation is already durable gateway-side). We previously discarded the hello-ok payload, so any server-side rotation bricked every following connect. `_handshake` now returns the reissued token and the client persists it (`device.update_device_token`, guarded on `device_id` so a concurrent re-pair is never clobbered) and adopts it in-memory. Persistence failure never fails the user's turn — the revived self-heal recovers the next connect.
3. **Hermetic self-heal tests** — they depended on the site having no `chat_device_id` (the repair convoy-guard reads real Settings) and failed on any dev site with a live pairing; `_persisted_device_id` is now pinned in the shared helper.

## Tests

- `test_chat_openclaw_client`: 31 passed (7 new: real wire-shape triggers repair; INVALID_REQUEST *without* auth details must NOT wipe creds; reissued-token persisted/adopted, no-op cases, persist-refusal and persist-failure resilience).
- `test_chat_device`: 13 passed (2 new for `update_device_token` including the concurrent-re-pair guard).

## Not in this PR

- Ops remediation for currently-affected sites: run `jarvis.chat.device.rotate_chat_device` (System Manager) — after this PR the self-heal makes that automatic.
- Pinning the openclaw image by digest (currently `ghcr.io/openclaw/openclaw:latest`, so deployments silently roll the gateway version) — admin/fleet-side change, recommended follow-up.

### DIFF
--- a/jarvis/chat/device.py
+++ b/jarvis/chat/device.py
@@ -237,6 +237,28 @@ def clear_credentials() -> None:
 	frappe.db.commit()
 
 
+def update_device_token(new_token: str, *, device_id: str) -> bool:
+	"""Persist a gateway-REISSUED device token for the current pairing.
+
+	openclaw's hello-ok can carry a rotated ``auth.deviceToken`` (the
+	gateway replaces the stored token whenever the existing entry no
+	longer lines up with the connect's scopes/issuer). The rotation is
+	already durable on the gateway side by the time the client sees
+	hello-ok, so a bench that keeps signing with the pair-time token
+	fails every FOLLOWING connect with "device token mismatch".
+
+	Guarded on ``device_id``: if another worker re-paired while this
+	connect was in flight, Jarvis Settings now holds a DIFFERENT
+	device's credentials, and the rotated token of the old device must
+	not overwrite them. Returns True when persisted."""
+	settings = frappe.get_single("Jarvis Settings")
+	if (settings.chat_device_id or "").strip() != device_id:
+		return False
+	settings.db_set("chat_device_token", new_token)
+	frappe.db.commit()
+	return True
+
+
 def _normalize_metadata(value: str) -> str:
 	"""Mirrors openclaw's normalizeDeviceMetadataForAuth: trim + ASCII lowercase
 	(only [A-Z] → [a-z]; Unicode left alone, matching the deterministic
--- a/jarvis/chat/device.py
+++ b/jarvis/chat/device.py
@@ -250,13 +250,26 @@ def update_device_token(new_token: str, *, device_id: str) -> bool:
 	Guarded on ``device_id``: if another worker re-paired while this
 	connect was in flight, Jarvis Settings now holds a DIFFERENT
 	device's credentials, and the rotated token of the old device must
-	not overwrite them. Returns True when persisted."""
-	settings = frappe.get_single("Jarvis Settings")
-	if (settings.chat_device_id or "").strip() != device_id:
-		return False
-	settings.db_set("chat_device_token", new_token)
-	frappe.db.commit()
-	return True
+	not overwrite them. The check-then-write runs under the same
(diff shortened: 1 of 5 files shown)


kavin-114:
**High-effort multi-agent review completed** (4 finders → 10 verifiers on the pooled candidates). Outcome:

| # | Finding | Action |
|---|---|---|
| 1 | `update_device_token` check-then-write not serialized vs a concurrent re-pair (could persist old device's token under new device's identity) | **Fixed** in f043152 — runs under the `chat_device_pair_repair` lock; lock-unavailable skips persist |
| 2 | Repair path could wipe a pairing a peer just healed (rotation keeps `device_id`, so the id-only guard passes) | **Fixed** — repair now also compares the persisted token vs the one the rejected connect presented; if it moved on, skip the wipe and retry |
| 3 | hello-ok parsing assumed `payload`/`auth` are dicts → a successful connect could crash on an unexpected shape | **Fixed** — shape-defensive; tested against list/scalar/non-string shapes |
| 4 | `error.details` threaded only at the connect raise site | **Fixed** — all three rejection raise sites now carry it |
| 5 | `chat_device_token` persisted via `db_set` → Password field stored plaintext in `tabSingles` | **Deferred, deliberate** — this is the existing pattern for ALL chat_device_* write sites (`_save_credentials`, `clear_credentials`); converting one site alone desyncs `__Auth` vs table-fallback reads and can serve stale credentials. Follow-up PR should move all write sites to encrypted storage together (+ a migration for existing rows). |
| 6 | `_STALE_PAIRING_CODES` layer is dead code against current openclaw | **Kept, deliberate** — belt-and-braces for a future openclaw that promotes internal reasons to wire codes; removal is the riskier edit for a regression-fix PR |

Tests after the fix wave: `test_chat_openclaw_client` 33 passed, `test_chat_device` 14 passed.

PR #276, opened 2026-07-10
Encrypt Settings Password-field writes, migrate plaintext/stale secrets, dedup pool syncs

kavin-114:
Closes the jarvis-app HIGH findings from the 2026-07-02 proxy security audit (Password fields written in plaintext via `db_set`) and adds the customer-plane half of the zero-interruption goal: change detection so pool-unrelated saves stop re-POSTing the full pool + secrets.

## What

- **`jarvis/_password_utils.py` (new):** `set_settings_password` — `set_encrypted_password` into `__Auth` + `db_set` of a fixed 10-char mask (fixed-length so the mask never leaks secret length; fires no hooks, preserving the deliberate no-`on_update`-retrigger property of the db_set call sites). `clear_settings_password` — blanks the column AND drops the `__Auth` row, so `get_password` can't keep serving a revoked secret.
- **Fixed every plaintext write site:** `onboarding.py::write_connection` (4 fields), `chat/device.py::_save_credentials` / `update_device_token` / `clear_credentials`, `api.py::rotate_agent_token`, and `selfhost.py::save_self_hosted` — a site the original audit missed (customer-supplied openclaw bearer written in the clear; blank token now explicitly clears, matching the old `db_set("")` semantics).
- **`patches/v1_11_encrypt_plaintext_settings_passwords`:** one-time backfill over all 7 Password fields on the `Jarvis Settings` Single, three-branch: plaintext → encrypt + mask; masked → untouched; **empty → purge any stale `__Auth` row** (pre-fix clears blanked the column without `remove_encrypted_password`, and `get_password` falls through to `__Auth` on a falsy column — resurrecting revoked device tokens). Idempotent; verified under a real `bench migrate` (Patch Log row present).
- **Pool-sync dedup (`jarvis_settings.py`):** `_pool_state_snapshot` (preset + routing_mode + every field `build_pool_payload` serializes per models[] row, secrets sha256-digested so `doc.flags` can never leak one) + `_pool_sync_is_redundant` gate. The enqueue is skipped only when a before-doc exists AND snapshots are equal AND `last_sync_status` starts with `"ok"` (failed syncs stay retryable by re-saving); `flags.force_admin_sync` bypasses the gate. The snapshot is captured in `validate()` — by `on_update` time `_save_passwords` has masked freshly-typed child secrets to `'*'*len(value)`, so a same-length key rotation would compare "unchanged" and be silently dropped; validate-time capture makes rotations reliably read as changes (pinned by test).

## Tests

All local suites green on a fresh DB: `test_password_write_encryption` 11/11 (incl. the stale-`__Auth` purge and mask-idempotency pins), `test_unified_llm_config` 83/83 (incl. 4 new gate tests + same-length-rotation), `test_settings_on_update` 37/37, `test_chat_device` 14/14, `test_onboarding_sync` 26/26, `test_selfhost` 40/40, `test_dev` 8/8, `test_api` 28/28. Local site lacks hrms after a reinstall — CI is the authority.

### DIFF
--- a/jarvis/_password_utils.py
+++ b/jarvis/_password_utils.py
@@ -1,0 +1,61 @@
+"""Shared helper for writing Jarvis Settings Password fields OUTSIDE the
+normal Document.save() pipeline.
+
+Frappe only encrypts a Password field into the __Auth table when a value
+passes through Document._save_passwords() - i.e. a real doc.save(). Writing
+via doc.db_set(...) or frappe.db.set_value(...) bypasses that entirely and
+stores the raw value in plaintext in the Single's row in tabSingles. Several
+call sites in this app use db_set deliberately (to avoid re-triggering
+on_update's admin-sync push during onboarding/rotation flows) but were
+writing secrets straight into that plaintext column.
+
+This mirrors the idiom already established in
+jarvis_settings.py::_on_update_unified_llm for mirroring models[0].api_key
+into the legacy llm_api_key field: encrypt the real value into __Auth via
+set_encrypted_password, then write only an all-asterisk mask to the doc's
+own field (db_set) so tabSingles never holds the secret. Every reader in
+this app already goes through get_password(), so this is a drop-in swap.
+"""
+
+from frappe.utils.password import set_encrypted_password
+
+SETTINGS = "Jarvis Settings"
+
+# Same masking shape jarvis_settings.py already uses for llm_api_key - a
+# fixed-length mask (not len(value)-length like Document._save_passwords)
+# so the mask itself never leaks the secret's length.
+_MASK = "*" * 10
+
+
+def set_settings_password(doc, fieldname: str, value: str, *, update_modified: bool = True) -> None:
+	"""Encrypt ``value`` into __Auth for Jarvis Settings.<fieldname>, then
+	write only the mask to ``doc`` via db_set (never the plaintext).
+
+	No-op when ``value`` is falsy - callers already guard on truthiness
+	before calling (matching the db_set(...) call sites this replaces, which
+	only wrote when the incoming value was present).
+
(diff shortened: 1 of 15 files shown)


kavin-114:
Code review (multi-agent, high effort) ran on this PR — 7 raw findings deduped to 6; resolution:

**Fixed in `4b2dc1a`:**
1. **CONFIRMED — `update_device_token` silently no-ops on a falsy token but returned `True`** (contract says "Returns True when persisted"; the helper's falsy guard made the write vanish). Now guards `if not new_token: return False` before the lock, with a regression test.
2. **CONFIRMED — `save_self_hosted`'s blank-token branch imported `clear_settings_password` locally**, bypassing the fully-mocked test class's seam (a future blank-token test would have performed a real `__Auth` delete). Import moved to module level; new seam test asserts the clear helper is called and no real DB is touched.
3. **CONFIRMED (cleanup) — the v1_11 patch duplicated the mask literal**; now imports `_MASK` from `_password_utils`.

**No change (rationale):**
4. **`rotate_agent_token` two-step failure window** — pre-existing failure mode, byte-identical consequences with the old single `db_set` under Frappe's request-transaction rollback (admin/container rotated, bench write rolled back). The PR neither widens nor narrows it materially; a resync path is a separate follow-up if wanted.
5. **Migration read-then-write race under concurrent traffic** — standard Frappe patch semantics; `bench migrate` runs in the maintenance window and this is a one-time backfill.
6. **Per-field `get_single_value` loop** — deliberate: the raw per-field read with `cache=False` is the documented way to see the stored Singles value; 7 SELECTs in a one-time patch.

Post-fix: `test_chat_device` 15/15, `test_selfhost` 41/41, `test_password_write_encryption` 11/11.

PR #295, opened 2026-07-14
fix(ai-models): five correctness bugs from the #294 review (two could destroy a customer's OAuth connection)

kavin-114:
Follow-up to **#294**. A high-effort review pass ran *after* it merged and found **5 CONFIRMED correctness bugs**. Two of them can **destroy or misattach a customer's OAuth credential**, so this is not cosmetic.

The two worst are **verified live** on a tenant with a real connected subscription.

---

## 1. 🔴 Re-picking the same provider WIPED a connected subscription

`JvCombo.choose()` emits `update:model-value` **unconditionally** — even when you re-select the option that is **already chosen**. `onUpstreamChange()` drops every connected account (an OAuth account is authorized against **one** provider).

So clicking **"OpenAI"** on a row already set to OpenAI **silently disconnected a working subscription.** No confirmation, no undo.

The onboarding combo has carried a same-value guard all along — *with a comment saying exactly this*:
> *"reselecting the CURRENT provider must be a no-op rather than wiping a finished OAuth connect"*

#294 added a **second** combo in Settings and **did not carry the guard over**. I copied the widget and not the warning.

**Verified live:** account count **1 → 1** after re-selecting the same provider. Before the fix: **1 → 0**.

## 2. 🔴 The config panel tracked its row by ARRAY INDEX while the list stayed mutable

`panel.index` is an array index, but **Reorder / Remove** on the always-visible failover list were never gated.

> Open **Edit** on row 2 → start OAuth → while the sign-in tab is open, reorder row 1 → `panel.index` now points at a **different row** → the pasted callback attaches the new account to the **wrong model** and **auto-saves** it.

Reorder + Remove are now **disabled while the panel is open**. Edit/Reconnect stay live — they *re-set* `panel.index`, so they cannot desync it.

**Verified live:** Remove flips **enabled → disabled** the moment the panel opens.

## 3. Switching a subscription row back to API key left a stale model id
`setCredType`'s reset was gated on `singleMode`, so it **never ran in Settings** — leaving e.g. `gpt-5.5` on an **Anthropic** api-key row. A provider/model mismatch that is **invisible** (the field is hidden) and only fails at the upstream.

## 4. A stored subscription row with an empty model was permanently unsavable
The model id is derived in `setCredType`/`onUpstreamChange` — but **neither fires for a row that merely LOADS**. Legacy data with an empty model rendered *"Model not set"* with **no field to type one into** (removed in #294), and every Save then failed `validatePool`'s *"Every model needs a model id"* — **with no way out**. `seedRows()` now backfills it on the same rule the editors use.

## 5. A pool-wide verdict was painted on every row's health dot
`sync.subscription_status` is **one** verdict for the pool. With two subscriptions, an `unverified` **flagged the healthy row too**, and a `verified` **vouched for a row that was never probed**. Now attributed only when there is exactly **one** subscription row; otherwise neutral — don't assert what we didn't measure.

*(Contract 1.11 added per-model verdicts. Key off `model_statuses` once it's plumbed admin → customer — currently the fleet emits it and nobody consumes it.)*

---

**Onboarding (`singleMode`) is untouched** — its guard was already there, and these fixes restore parity with it. That's the through-line: every one of these is a place where a settings-side clone of an onboarding control lost the invariant its original carried.

### DIFF
--- a/frontend/src/components/LlmPoolEditor.vue
+++ b/frontend/src/components/LlmPoolEditor.vue
@@ -44,7 +44,7 @@
 
       <div v-if="!rows.length && !showDirectRow" style="font-size:13px;color:var(--text-3);padding:8px 0;">No models yet. Add one below.</div>
 
-      <div v-for="(row, i) in rows" :key="i" class="jv-flist-row">
+      <div v-for="(row, i) in rows" :key="row._uid ?? i" class="jv-flist-row">
         <span class="jv-pool-badge">{{ i + 1 }}</span>
         <span class="jv-flist-chip">{{ sourceChip(row) }}</span>
         <span class="jv-flist-model" :class="{ 'jv-flist-model--unset': !row.model }">{{ rowModelLabel(row) }}</span>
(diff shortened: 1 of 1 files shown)


kavin-114:
## Follow-up: the index gating dead-ended a live tenant

Reported from a real tenant while testing this branch: **Save configuration did nothing.**

It was doing something — it was correctly refusing. But the way out was blocked:

1. The tenant's pool is empty (`models=[]`), so `load()` seeds a **blank row** to fill in.
2. The customer added a real model on top of it and hit Save.
3. `validatePool` rejects the blank row — *"Every model needs a provider and a model id."*
4. And the blank row **could not be removed**, because this branch's fix #2 disabled Remove while the config panel was open.

**Save blocked. Remove blocked.** That gating was mine, and it was the wrong shape of fix.

### Fixed at the root

The gating existed only because `panel` tracked its target row by **array index**, so mutating `rows` underneath a live panel repointed it at a neighbour. Rather than freeze the list, the panel now tracks its row by **identity**:

- every row carries a client-only `_uid`; `panel.uid` replaces `panel.index`
- `panelRow` resolves by `find(r => r._uid === panel.uid)`, and closes if that row is removed
- Reorder/Remove are **live again**

The original defect — reorder mid-OAuth repointing the panel so a pasted callback attaches the account to the **wrong model** and auto-saves — is now *structurally impossible* rather than merely unreachable. `buildSaveModels` maps explicit fields so `_uid` never reaches the payload, and `reorder()` splices the same object refs so it survives a move.

### Second fix: a blank row must not block Save

`save()` now prunes a blank **API-key** row when other rows are real. It is a placeholder, not a choice. Deliberately narrow:
- an **in-progress subscription** row is *not* pruned — it represents a connect the customer started, and should draw *"connect your account"* rather than silently vanish
- if pruning would empty the pool, every row is kept so validation still speaks up

### Verified live on the affected tenant

| | before | after |
|---|---|---|
| Remove, panel open | disabled | **enabled** |
| Remove the edited row's neighbour mid-edit | panel goes headless | **panel stays on its own row** |
| blank + real model → Save | ❌ `Every model needs a provider and a model id` | ✅ **saves; payload carries 1 model** |

No tenant data was written during the test (the save call was intercepted).

PR #325, opened 2026-07-15
Settings revamp: one canonical user DocType + dialog slim / consistency / a11y

kavin-114:
## Settings revamp: one canonical user DocType + dialog slim / consistency / a11y

Two parts. Design + plan live in `docs/superpowers/{specs,plans}/2026-07-15-settings-revamp*.md`.

### Part A — DocType consolidation (backend)
Makes `Jarvis User Settings` the single per-user store.

- **Merged `Jarvis User Preference` into `Jarvis User Settings`** — its 3 greeting-cadence fields (`business_greeting_state`, `business_greeting_chat_count`, `business_greeting_hidden_at_count`) now live under a new **Preferences** tab; the doc is reorganized into **Preferences** / **Usage & limits** tabs.
- **Data migration** `v1_14_merge_user_preference_into_settings` — copies every preference row onto the matching settings row; existence-guarded + idempotent.
- **`greeting.py` rewired** to the new table (behavior unchanged — only the backing table moved).
- **Old DocType dropped** via `v1_15_drop_user_preference_doctype` (two-step `delete_doc` + `DROP TABLE IF EXISTS`, existence-guarded). Patches ordered v1_14 → v1_15; both no-op on a fresh CI DB.

### Part B — Settings dialog (frontend + thin theme slice)
- **Theme now persists on Frappe's native `User.desk_theme`** (Light/Dark/Automatic ↔ light/dark/system) via `set_user_theme` + boot `jarvis_desk_theme`; `theme.js` sources from boot and retires the `jarvis-theme` localStorage write. Theme roams per-user across devices.
- **Removed the Appearance and Macro runs panes** (rail 11 → 9). Theme is switchable from the header/user-menu; macro-run history lives at `/macros/runs`.
- **Real dialog semantics (fixes audit ship-blocker P0-5)** — `role="dialog"`, `aria-modal`, focus trap + initial focus + focus-return, background `inert`, `prefers-reduced-motion`. The dialog is `Teleport`ed to `<body>` so `inert` on `#app` doesn't freeze it.
- **Disconnect → in-app `confirmDialog`** (frappe-ui) in both `DirectSubscriptionCard` and `LlmPoolEditor`, replacing raw `window.confirm()`.
- **New AI-models row no longer defaults to "Anthropic"** — seeds an empty provider ("Select a provider"); save-time `validatePool` already requires one.
[...]

PR #327, opened 2026-07-15
fix(ai-models): show per-model health for api-key failover rows

kavin-114:
## Problem

When a customer configures a **failover pool** with an api-key model, a broken member (bad key → 401, wrong/dead `base_url` → unreachable) was **invisible** in the AI-models UI: the row showed the same green **"key set"** as a healthy one, and the save reported **"Applied"**. They'd discover the fallback was dead only when the primary failed — the worst possible moment.

Root cause (traced on `site.jarvis`, 2026-07-15): the fleet computes a per-model verdict (`model_statuses`, contract 1.11) and returns it on the pool-apply response, but it was **dropped at the bench boundary** — `jarvis_settings` persisted only `subscription_status` + `warnings`, and `LlmPoolEditor.accountHealth` **hardcoded every api-key row green** (the acknowledged *"drop this guard once model_statuses is plumbed"* TODO).

## Fix — plumb `model_statuses` to per-row health

- **Persist** `last_model_statuses` alongside `last_sync_warnings` (`jarvis_settings.py`), with the same no-op / clear-on-failure reasoning. New hidden `Long Text` field on Jarvis Settings.
- **Return** it from `onboarding.get_llm_sync_status` (corrupt / non-list → `[]`, never 500s the poller).
- **Render** it: `LlmPoolEditor.accountHealth` now keys api-key rows off their own verdict via a new **pure** `apiKeyModelHealth(row, model_statuses)`:
  - `failed` → **warn** "Not working" + actionable title (check key / model id / base URL; skipped during failover)
  - `unchecked` → "Not verified yet"
  - `verified` / absent → quiet green (now *meaningful*; absent = pre-1.11 fleet or not-yet-probed)
  - Matches by model id, disambiguating the legal provider/model-id collision by provider (rows store the provider *label*, `model_statuses` the *id*, so the match normalizes via `providerId`).

No admin change: `jarvis_admin.fleet.llm_pool.update_llm_pool` returns the raw fleet result unmodified, so `model_statuses` already flows admin→customer.

## Pairs with fleet-agent PR #137

PR #137 makes an unreachable `base_url` return `failed` (instead of silent `unchecked`). Together: a dead fallback leg now goes **red on its own row** here, and also raises the pool-wide "N model(s) need attention" banner. Merge order doesn't matter (this degrades gracefully on a pre-#137 fleet — `unchecked` renders "Not verified yet").

## Tests

- `frontend/src/llm/pool.test.js` (5, `node --test`): the health mapping incl. the model-id collision and the pre-1.11 / no-verdict → green fallbacks. **30/30 pass.**
- `jarvis/tests/test_onboarding_sync.py` (3): the poller returns parsed `model_statuses` and degrades corrupt/empty/non-list to `[]`. (Runs on CI — the app is installed from the main checkout, not this worktree.)
- Frontend **vite build clean** (13.9s).

No migration patch: a new Single-doctype field is synced by `bench migrate`'s meta reload; values live in `tabSingles`.

### DIFF
--- a/frontend/src/components/LlmPoolEditor.vue
+++ b/frontend/src/components/LlmPoolEditor.vue
@@ -521,6 +521,7 @@ import * as api from "@/api"
 import {
   deriveMode, reorder, presetToModels, missingVendorKeys, validatePool,
   PROVIDER_LABELS, providerLabel, providerId, seedRowsFromConfig, defaultSubscriptionModel,
+  apiKeyModelHealth,
 } from "@/llm/pool"
 import { errMessage as _err } from "@/lib/errors"
 import { useConfirm } from "@/composables/useConfirm"
@@ -559,7 +560,7 @@ const selectedPreset = ref("")
 const keysByVendor = ref({})
 const err = ref("")
 const saving = ref(false)
-const sync = ref({ last_sync_status: "", pending: false, subscription_status: "", warnings: [] })
+const sync = ref({ last_sync_status: "", pending: false, subscription_status: "", warnings: [], model_statuses: [] })
 const savedSnapshot = ref("__init__")  // savable pool as of last load/save; drives the unsaved-changes notice
 let pollTimer = null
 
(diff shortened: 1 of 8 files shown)


kavin-114:
Addressed the high-effort review's CONFIRMED finding (single-match branch misattributing a verdict by model id alone) in 05b9c0c.

`modelVerdict` now keys on **(provider, model) together** in a single `find()` — a model id is not a key (validate() only forbids duplicate provider/model *pairs*), so a lone `model_statuses` entry for one provider can no longer attach its verdict to a different-provider row that happens to share the id. This also collapses the old length===1 / length>1 branching into one lookup.

New test: a single entry under a *different* provider stays quiet green instead of inheriting the other provider's `failed`. 31/31 pass, vite build clean.

kavin-114:
CI note: the initial run's sole failure was `TestOnUpdateAdminDispatch.test_admin_path_restart_returned_action_reflects_in_status` — the **pre-existing order-dependent flake** that class documents (a stale Jarvis Settings Single makes on_update misclassify a provider change and hit the unmocked admin path → "admin unreachable"). All four new `model_statuses` tests passed; **main's HEAD is red on the same flake** (tripped by an unrelated commit). This PR's new tests shift discovery order and trip it too.

Fixed in f7933ba by applying the **verified fix from in-flight #325** (`frappe.clear_document_cache` in `setUp`) so the test is order-independent and this PR is self-contained + green. It's a test-only change and reconciles trivially with #325 if that lands first.

kavin-114:
**CI red is a `main` regression, not this PR.** `main`'s current HEAD `bd0ce02` (#326, "reclaim idle sessions") fails `TestOnUpdateAdminDispatch.test_admin_path_restart_returned_action_reflects_in_status` on its **own** CI ([run 29441020792](https://github.com/Aerele-RnD/jarvis/actions/runs/29441020792)) — same test, same `test_settings.py:217` assertion (`'failed: admin unreachable' != 'ok (restart via admin)'`) — even though #326 already added `clear_document_cache` to that setUp. So the cache-clear alone doesn't fix it; the test still routes the provider-change save to the unmocked pool path (likely leftover `models` child-table rows from a prior test → `compute_proxy_active` true → `post_update_llm_pool` instead of the mocked `post_update_llm_creds`).

PR CI builds *this branch merged with `main`*, so #327 inherits that failure. **All of this PR's own tests pass** (the 4 new `model_statuses` backend tests ✔, frontend 31/31, vite build clean), and this PR's net diff doesn't touch `test_settings.py` (my earlier duplicate cache-fix was superseded by #326's on merge). This PR goes green as soon as `main` is fixed. Flagging so the `main` regression gets addressed separately.

PR #330, opened 2026-07-15
feat(personalise): mine chat transcripts into Personalise questions

Vigneshsekar:
## Summary

Closes the loop the team asked for — **chat → question → answer → note → wiki/skills** — by adding the one missing half: a daily sweep that mines recent chat transcripts into **Personalise questions**. The downstream (answer → `Jarvis Voice Note` → `voice_facts` sweep → wiki *context* facts / learned *rule* skills) already exists and is untouched; this adds a **fourth Personalise question source** alongside behavioural-learning patterns, org rules and reviewer follow-ups.

### Why this shape (and not "summarise chats straight into the wiki")
The transcript is a **signal for what to ask**, never the knowledge itself — the human's answer is the verified knowledge. So nothing unverified ever reaches the wiki, which fits Jarvis's trust model and is the differentiated version of raw-RAG chat ingestion. Behavioural learning watches what you *did* (transactions); this watches what you *said/asked* (chat) — complementary, not redundant.

### How it works
- New `jarvis/learning/chat_mining.py`, registered in `scheduler_events['daily']`, self-gating + never-raises (the `voice_facts.process_daily` template).
- Selects conversations with fresh user messages since a **durable Settings watermark** (first run looks back 7 days only — never an unbounded backfill), excluding File Box and macro/agent-run machine dialogues and Administrator/Guest owners.
- Rebuilds a clean user+assistant transcript (hidden/tool/errored/streaming rows and ` ```jarvis-ask ` fences stripped), **single-owner batches** (attribution can never cross users), ≤5 LLM calls/run.
- Routes mined facts through `lifecycle.upsert_candidate` under one manual `Jarvis Pattern Run` (detector `chat-context`) — the **same vehicle `voice_facts` uses** — so the question materializes via the existing hook and inherits the **per-user daily cap, `(source_pattern, user)` dedupe, Deleted-suppression** and the existing **"From your chat patterns"** origin.
- **Zero frontend changes** (`OriginBadge` already maps that origin), **zero new endpoints**, answer path untouched.

### Safety / cost
- Every transcript fenced as `<untrusted-data>`; mined statements + questions run through `scan_instruction_injection` (mined text re-enters prompts when answered).
- A failed batch **holds the watermark** (retried next run, no loss); re-mining is harmless (`pattern_key` is content-derived).
- Chat-mined patterns stamped `personalise_origin=1` so a reviewer promoting them org-wide is warned to scrub personal nuance.
- Gated by a `jarvis_chat_mining_disabled` site_config kill switch, a new `chat_question_mining_enabled` Check (NULL=ON, seeded by `roles.after_migrate`) and the personalise master toggle. Not self-host gated (all work is bench-side).

### New Jarvis Settings fields (Personalisation section)
`chat_question_mining_enabled` (Check, default 1) + read-only `chat_mining_watermark` / `chat_mining_last_run_at` / `chat_mining_last_run_status`.

## Pre-merge checklist
- [x] New/changed behaviour has tests — `test_chat_mining` (22): gates, minting, cap+dedupe+Deleted-suppression, exclusions, transcript hygiene, injection defence, failure/watermark, answer flow.
- [x] Adjacent suites green — `test_personalise_pipeline` (37), `test_voice_facts` (22), `test_personalise_doctypes` (44).
- [x] Verified **live end-to-end on dev** against a real OpenRouter call: two durable facts mined → two natural questions minted with the right origin/context.
- [x] `ruff format` + import check clean (tabs / double quotes / 110 cols).
- [ ] CI `tests` green (coverage floor 85%).

### DIFF
--- a/frontend/src/components/personalise/PersonalisationSettings.vue
+++ b/frontend/src/components/personalise/PersonalisationSettings.vue
@@ -195,6 +195,25 @@
 								<Switch v-model="settings.personalise_enabled" size="md" />
 							</div>
 
+							<div class="flex items-start justify-between gap-3 rounded-lg border p-4">
+								<div>
+									<div class="text-base font-medium text-ink-gray-9">
+										Learn from chats
+									</div>
+									<div class="mt-0.5 text-sm text-ink-gray-6">
+										Once a day, Jarvis reviews recent chats and drafts questions so people can
+										confirm what it should remember. Answers become wiki notes and skills.
+									</div>
+									<div
+										v-if="settings.chat_mining_last_run_status"
+										class="mt-1.5 text-xs text-ink-gray-5"
+									>
+										Last run{{ chatMiningLastRunAgo }}: {{ settings.chat_mining_last_run_status }}
+									</div>
+								</div>
+								<Switch v-model="settings.chat_question_mining_enabled" size="md" />
+							</div>
+
 							<FormControl
 								type="number"
 								label="Max learning / chat-pattern questions per person per day"
@@ -276,6 +295,7 @@ import {
 	getPersonalisationSettings,
 	setPersonalisationSettings,
 } from "@/api/personalise"
+import { timeAgo } from "@/utils/datetime"
 
 const props = defineProps({
 	open: { type: Boolean, default: false },
@@ -467,9 +487,16 @@ function confirmDeleteRule(rule) {
 const settings = reactive({
 	daily_question_cap: 5,
 	personalise_enabled: true,
+	chat_question_mining_enabled: true,
+	chat_mining_last_run_at: null,
+	chat_mining_last_run_status: "",
 })
 const settingsLoading = ref(false)
 const savingSettings = ref(false)
+
+const chatMiningLastRunAgo = computed(() =>
+	settings.chat_mining_last_run_at ? ` ${timeAgo(settings.chat_mining_last_run_at)}` : "",
+)
 
 async function loadSettings() {
 	settingsLoading.value = true
@@ -477,6 +504,9 @@ async function loadSettings() {
 		const s = await getPersonalisationSettings()
 		settings.daily_question_cap = intOr(s.daily_question_cap, 5)
 		settings.personalise_enabled = !!s.personalise_enabled
+		settings.chat_question_mining_enabled = !!s.chat_question_mining_enabled
+		settings.chat_mining_last_run_at = s.chat_mining_last_run_at || null
+		settings.chat_mining_last_run_status = s.chat_mining_last_run_status || ""
 	} catch (e) {
 		toast.error(errMsg(e))
 	} finally {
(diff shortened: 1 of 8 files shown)


Vigneshsekar:
## Post-development review (3× Sonnet: correctness, security, UX)

Ran an adversarial review pass with three independent agents. **Fixed in `6f6f86c`:**

**Correctness / security**
- **Owner-salted `pattern_key`** (`chat|{owner}|{stmt}`) — two users stating the same fact no longer collapse into one learned-pattern row (one user's question was being silently absorbed, and cross-run `_apply_evidence` could overwrite the other's provenance). *Flagged critical by 2 of 3 reviewers.* New test.
- **Pattern Run finalized in `try/finally`** + guarded per-candidate post-processing — a mid-loop crash can no longer leave a `Running` chat run that wedges the nightly behavioural engine (the orchestrator treats any `Running` row, regardless of `scan_mode`, as in-progress). New test.
- **Fenced the "Already asked" block** — it's prior LLM output from the same user's chat, now wrapped in `<untrusted-data>` like every transcript excerpt.
- **Bounded transcript fetch** (`seq desc` + `limit`) instead of loading a whole thread to slice its tail.
- **Tie-safe watermark** — a conversation tied on the exact boundary timestamp just past the page cap can no longer be stranded forever.
- **Window clamp** (`MAX_WINDOW_DAYS=30`) so a long stall can't keep widening the `GROUP BY` window.

**UX / operability**
- **Toggle surfaced in the Personalise settings dialog** ("Learn from chats" switch + "Last run … : status") — was desk-form-only, invisible to the Jarvis Admin persona. Wired through `get/set_personalisation_settings`.
- **Provenance trail** — chat-question context now reads *"Noticed in a chat conversation on `<date>`"* so a user can place the fact before confirming.
- **Prompt hardened** — excludes personal/HR/health/legal/interpersonal content; questions phrased to acknowledge the user already mentioned it.

Tests: `test_chat_mining` 25 · `test_personalise_pipeline` 37 · `test_personalise_api` 84 — all green. Owner-salt + settings API + provenance re-verified live on dev against a real model call.

### Consciously deferred — product decisions, not code bugs (flagged for a call)
1. **Per-user opt-out.** Consent today is a single org-wide toggle. A user can't exclude their own chats from mining. Reasonable to ship with the global toggle for v1, but someone should own this decision before wide rollout (voice notes are opt-in by construction; chat mining is not).
2. **Shared daily-cap fairness.** Chat-mined questions share the 5/day `LEARNING_ORIGINS` bucket with behavioural + voice questions; a busy chat day could crowd out higher-intent sources (the daily backstop scans newest-first). Candidate follow-up: a reserved sub-share, or oldest-first backstop.
3. **Throughput ~5 owners/run.** `MAX_BATCHES_PER_RUN=5` (≤5 LLM calls/day) means large orgs cycle over several days; the watermark keeps it fair (oldest-activity-first) and durable facts don't change fast, but worth revisiting if it feels slow at scale.
4. **Semantic near-duplicate dedup.** Exact-statement dedup is structural; near-rephrasings rely on the model honoring the (now fenced) "Already asked" list. A future embedding pass would harden this.

Happy to spin any of #1–#4 into follow-up issues.

Vigneshsekar:
✅ CI `tests` green (6m1s) — full app suite + 85% coverage floor pass. Ready for review/merge.

PR #347, opened 2026-07-17
feat: per-tenant customization discovery (describe_customizations tool + context clause)

Navin-S-R:
## Per-tenant customization discovery (Option B + D)

Customers run custom apps (custom DocTypes, workflows, reports, print formats) and customize core doctypes (Custom Fields, Property Setters). The agent could always *see* these through `get_schema`/`get_list`, but nothing told it to look — discovery was 100% pull-based against a catalog that only describes core apps. This ships the decided design: a **live, permission-fenced index tool** plus a **counts-only context clause**, with activation telemetry to decide whether a generated-skill escalation (Option A) is ever needed. **Zero skill pushes, zero container restarts** — the feature's core promise.

### Stage map (one commit each)

1. **`get_schema` custom flags + classification groundwork** — `site_profile/apps.py` is the single source of truth for app-level custom/core classification (`_KNOWN_APPS` + `Jarvis Settings.core_apps_override`, module→app mapping via Module Def). `get_schema` gains a doctype-level `custom` flag (custom=1 OR module of a non-core app; classification failure reads standard) and per-field `is_custom` under the absent-means-not-configured contract. Schema cache key now versioned via a single `_cache_key` helper (stale pre-deploy entries can never be served). Two new Settings fields in one schema change.
2. **`describe_customizations` tool** — three-layer pipeline: `collect.py` (permission-free, set-based metadata queries; two-union doctype discovery — `custom=1` UNION module→custom-app, since app-shipped doctypes have `custom=0`; `is_system_generated` fields excluded; scripts as counts only) → `fence.py` (per-user `has_permission(read)`, memoized, fail-closed per item; no precomputed cross-doctype totals, so post-fence counts hold by construction) → `render.py` (pure budgeted renderer, six-tier shed ladder, retrieval recipe always survives). Returns `{markdown}`. Read-only by absence from `_WRITE_TOOLS`. No caching in v1 (never cache fenced/rendered output site-wide).
3. **`[Context:]` clause** — cached counts-only line (`; site customizations: custom app(s) X - N custom doctypes, ... - call jarvis__describe_customizations before assuming standard fields`), ≤200 chars by a shrink ladder, app names folded before entering the trusted bracket. Inserted after the wiki clause, before the personal clause (ordering invariant preserved). Invalidated by the same five schema doc_events as the schema cache; deliberately counts only sources those events cover. Kill switch `enable_customizations_clause` (NULL=ON).
4. **Telemetry + analyzer** — `jarvis/telemetry.py` (never-raise JSON lines to `jarvis.tool_telemetry`): every `describe_customizations` call, custom-target `get_schema`/`query`/`get_list` calls, one line per turn with `touched_custom`. `site_profile/analyze.py` computes the **activation rate** (index consulted before the first custom-entity touch). Escalation gate documented: activation < ~70% after ~2 weeks live → revisit Option A as a slim generated router; not before that evidence.

### Test output

All green (92 tests):

```
test_site_profile_render        Ran 21 tests  OK   (pure - no frappe import)
test_describe_customizations    Ran 18 tests  OK
test_customizations_clause      Ran 10 tests  OK
test_get_schema                 Ran 25 tests  OK
test_get_schema_meta_cache      Ran  7 tests  OK
test_site_profile_apps          Ran  6 tests  OK
test_registry                   Ran  5 tests  OK
```

### E2E (bench-side, live jarvis.local)

Adding a Custom Field to Sales Invoice + a throwaway custom DocType: the doc_events invalidated the clause cache; the rebuilt clause reflected the new counts; the tool's index named both fixtures; `get_schema` showed `"is_custom": true` on the field and `custom: true` on the doctype; **`custom_skills_synced_at`/`sync_status` unchanged before/after (zero pushes)**. Fixtures removed after. Real-site scale check: the clause renders at 191 chars with 2 discovered apps / 104 doctypes / 33 customized core doctypes / 10 workflows.

The **in-chat** E2E (agent calls the tool unprompted in a conversation) is deferred until the branch is deployed into the local tenant container (plugin rebuild + persona pull + restart; container runtime was down during development). Companion PRs: jarvis-openclaw-plugin + jarvis-persona (same branch name).

### Review gates (both blocking, independent passes)

- **R1 (security & correctness): PASS, zero blocking findings.** Verified: fence coverage + fail-closed direction, counts-post-fence by construction, no `roles_for_doctype`/`get_all_perms` in new paths, metadata-only payloads (no Select options / descriptions / script bodies), parameterized set-based queries, no fenced-output caching, read-only registration, clause fold + cap, versioned schema cache at all key sites, telemetry never-raise + hashed users. One cosmetic docstring finding fixed in a follow-up commit.
- **R2 (architecture & regression): PASS, zero blocking findings.** Verified: invariants 1–7, none of the explicitly-forbidden artifacts present (no sync machinery, no push-path changes, no new patches — `patches.txt` untouched), statically confirmed nothing can trigger a push/restart, clause ordering + invalidation coverage, registry↔plugin mirror exact, persona family clean, test adequacy against the charter, regression sweep of `turn_handler`/`api`/`hooks`/Settings JSON.

### Observation for reviewers

On the live dev site the classifier reports `india_compliance` and `lending` as custom apps. `lending` is the design working (no persona family teaches it → the index covers it). `india_compliance` has a persona family, so its 24 doctypes become redundant in the index once the separate allowlist fix ships — if it should count as core, either add it to `_KNOWN_APPS` or set it per site via the new `core_apps_override` field. Left as charter-specified pending that call.

### DIFF
--- a/jarvis/api.py
+++ b/jarvis/api.py
@@ -1045,8 +1045,9 @@ def _run_tool(tool: str, raw_args: dict | str | None,
 		# returns it verbatim, so the card + expiry ride the event, the record, and
 		# every resync identically (never rebuilt -> cannot diverge). build_card
 		# returns None for uncovered shapes; the SPA falls back to the raw preview.
-		import time
-
+		# NOTE: time is imported at module level - a function-local import here
+		# would make `time` local to ALL of _run_tool and break the read path's
+		# perf_counter above it (UnboundLocalError; caught by CI 2026-07-17).
 		from jarvis.chat import confirm_card
 		if isinstance(preview, dict):
 			preview["card"] = confirm_card.build_card(tool, args, preview)
--- a/jarvis/api.py
+++ b/jarvis/api.py
@@ -1,9 +1,10 @@
 import json
+import time
 
 import frappe
 from frappe.utils import strip_html
 
-from jarvis import audit
+from jarvis import audit, telemetry
 from jarvis._http import validate_bearer as _validate_bearer  # noqa: F401 (kept for callers in mcp.py)
 from jarvis._plugin_auth import PluginAuthError, validate_plugin_request
 from jarvis._session import impersonate
@@ -1086,7 +1087,15 @@ def _run_tool(tool: str, raw_args: dict | str | None,
 			"status": "pending_confirmation", "preview": model_preview, "tool": tool,
 		}}
 
-	return _dispatch_and_wrap(tool, args, is_write)
+	# Customization-discovery telemetry (read path; audit covers writes):
+	# record_tool is a fast no-op for untracked tools and never raises.
+	t0 = time.perf_counter()
+	result = _dispatch_and_wrap(tool, args, is_write)
+	telemetry.record_tool(
+		tool=tool, args=args, conversation=conversation,
+		duration_ms=int((time.perf_counter() - t0) * 1000), result=result,
+	)
+	return result
 
 
 # _error lives in jarvis/_responses.py - single source of truth for the
--- a/jarvis/api.py
+++ b/jarvis/api.py
@@ -1045,9 +1045,8 @@ def _run_tool(tool: str, raw_args: dict | str | None,
 		# returns it verbatim, so the card + expiry ride the event, the record, and
 		# every resync identically (never rebuilt -> cannot diverge). build_card
 		# returns None for uncovered shapes; the SPA falls back to the raw preview.
-		# NOTE: time is imported at module level - a function-local import here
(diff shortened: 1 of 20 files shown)


Navin-S-R:
Resolved the flagged observation: india_compliance is persona-taught core (we ship the india-compliance skill family), so it now lives in `_KNOWN_APPS` and no longer appears in the index/clause. Apps without a skill family (e.g. lending) remain discoverable by design. Regression test added; affected suites green (7+18+10).

PR #348, opened 2026-07-17
fix(fleet-gating): resync installed_apps to admin after app installs

Navin-S-R:
## Resync installed_apps to admin after app installs/removals

`installed_apps` — the signal the fleet uses to gate persona skill families (`agents.defaults.skills`) and shortcut-tool denies (`tools.deny`) — reaches the control plane in exactly one place: the `post_update_llm_creds` payload, fired from an LLM-settings save. Installing an app never fired it. Consequence: a tenant that adds a core app post-onboarding (e.g. hrms) gets **no hrms-* skills and keeps the hrms shortcut tools denied** until some unrelated credentials save happens — possibly never. Removals leave dead families allowed. The gap is silent.

### Fix

Every app change ends with `bench migrate`, so a new `after_migrate` handler compares the live `frappe.get_installed_apps()` against a snapshot of the last-synced list (new read-only Settings field `installed_apps_synced`) and, on change, enqueues the **existing** creds-restart sync — same `job_id` (coalesces with a real restart), same redis lock, same convergence handling. The worker re-reads Settings and its payload carries the live list; admin persists it, re-renders `openclaw.json`, restarts the container.

Correctness details:
- The snapshot is stamped **only after `post_update_llm_creds` returns** (admin persists `installed_apps` desired-first) — a failed send leaves it stale so the next migrate retries. Stale snapshot = one redundant resync; a premature stamp would silence the gap forever.
- An **empty snapshot seeds the baseline without a sync** — deploying this to the fleet does NOT trigger a restart wave; onboarding already sent the then-current list. Pre-existing stale tenants converge on their next app change or creds save.
- Self-hosted and pre-onboarding (no `jarvis_admin_url`) benches skip. Never blocks a migrate (best-effort + `log_error`, matching the other `after_migrate` handlers).

### Tests

`jarvis.tests.test_installed_apps_sync` — 10/10 green: seed-without-restart on first sight, no-op on unchanged, enqueue with the exact job kwargs on change (and snapshot stays stale), self-host/unconfigured skips, garbage-snapshot reseed, never-blocks-migrate, snapshot roundtrip, and both `_sync_via_admin` call-site behaviors (stamps after a successful send; does NOT stamp on `AdminAuthError`). Existing `test_settings` suite unaffected (green).

Related: surfaced while building Aerele-RnD/jarvis#347 (customization discovery) — a newly installed *core* app is deliberately ignored by that feature's index (`_KNOWN_APPS`), which made this staleness the only channel for its skills, and silent.


Navin-S-R:
**Review update (blocking finding fixed).** A consolidated Opus review found the original resync routed pool tenants through the single-model creds restart, whose fleet render would knock `proxy_active` tenants off Bifrost pool routing. Fixed in 26331ef + 7552edf: the resync now branches on `proxy_active` — pool tenants ride the existing pool sync job, whose payload now carries the live `installed_apps` (pairs with Aerele-RnD/jarvis-admin#236, which persists it before the fleet forward). Version-skew safe both ways: the bench stamps its snapshot only when the admin echoes `installed_apps_persisted` (an old admin silently ignores the extra field — verified against frappe's `get_newargs` arg filtering), and an unreadable `proxy_active` defers instead of guessing a leg. Reverification verdict: PASS. 13/13 tests green. **Deploy order: admin PR first (or together).**

PR #351, opened 2026-07-17
fix: blank draft card — fail loudly, and stop hiding Frappe's defaults

venkat102:
## The bug, from a real transcript

A user asked: *"Create the below tasks and assign to me — test the confirmation ui / test admin v2 / fix issues in the list"*.

The agent called `get_creation_context("ToDo", …)` and then emitted:

```
{"kind":"doc","verb":"create","doctype":"ToDo","summary":"3 open ToDos assigned to Administrator","docs":[{…},{…},{…}]}
```

**The card proposed nothing** — the model's headline, zero field rows, and an enabled Confirm. (Click Edit or Preview and you get one empty Description box.)

`docs` is not a `jarvis-action` key. Nothing in either frontend reads it. `buildDraftModel` reads `a.fields` → `undefined` → every non-`reqd` field is skipped → one empty box. **Nothing failed** — no throw, no error envelope, and the agent never learns the card rendered blank.

## Three fixes

**1. Fail loudly** (`ChatView.vue`, `pwa/ActionCard.vue`) — a block carrying `docs`, or a fieldless create, now renders an error naming the problem instead of a blank form. Silent-blank is the worst outcome available: it reads as *"Jarvis didn't try"* rather than *"Jarvis emitted a shape I can't draw"*.

**2. Split `auto`** (`get_creation_context.py`) — `"auto": bool(df.fetch_from) or (df.default not in (None, ""))` conflated two different things:
- `fetch_from` → Frappe **computes** it. Skipping is safe.
- `default` → Frappe **guesses** it. That's a decision, made silently.

`ToDo.date` has `default: "Today"`, so it was marked `auto`, so the note said skip it. Those three ToDos are created **due today** and nobody sees it. `auto` is now `fetch_from` only; a defaulted field surfaces its `default` and the note tells the model to decide it.

**3. Persona** (separate repo, `jarvis-persona` — not in this PR) — states that one `jarvis-action` block is one record.

## Notes for review

- **The batching rule already existed.** `AGENTS.md:115` covers several-records ("ask ONCE… then `create_doc(docs=[...])` in batches of up to 20"). The model **violated** it rather than falling through a gap. What was genuinely missing is that the block shape is documented positively and never negatively, and nothing validates it.
- **The PWA guard deliberately differs from the desktop's.** The PWA neither renders nor applies `tables`, so the desktop's "tables-only create is legal" escape would let it render blank again.
- **Not guarded via `parseAction` returning null** — `STRIP_RES` strips the fence from the prose unconditionally, so a null parse would make the proposal *vanish entirely*, which is worse than the blank card.
- `openDraftPanel` gets a try/catch and the Edit button gets `:disabled` — without both, the new throw kills a live button.
- **Nothing in app code reads `auto` back** from `get_creation_context` (only the model, and the plugin's tool description), so fix 2's blast radius is behavioural, not structural.

## Testing

`test_get_creation_context` **24/24** (5 new). Mutation-tested 4/4: reverting the `auto` conflation, setting `default` unconditionally, never surfacing `default`, and dropping `remaining` from the note each turn the right test red.

Both SFCs compile; `npm run build` green; no `v-html`.

⚠️ **Browser smoke is PENDING** — all three desktop surfaces (summary card / Preview / Edit) and the PWA card need a human. Compilation proves less than it looks here: `@vue/compiler-dom` is silent on `v-for`+`v-if` misuse.

## Known open (recorded, not fixed here)

- The card doesn't render Frappe-applied defaults — fix 2 puts the default in the model's hands instead.
- Nothing tells the **agent** its block failed to render.
- `kind:"email"` has the same silent-blank family (a `messages:[…]` fusion).
- **The PWA drops child tables on apply** — pre-existing; `apply()` never reads `action.tables`, so a fields+tables draft applied from the phone writes the parent without its rows.

Spec: `docs/blank-draft-card-spec.md` (Fable-reviewed over two rounds: don't-ship with 4 blockers, then ship after 2 edits).

### DIFF
--- a/docs/superpowers/specs/2026-07-17-blank-draft-card.md
+++ b/docs/superpowers/specs/2026-07-17-blank-draft-card.md
@@ -1,0 +1,367 @@
+# The blank draft card: three fixes
+
+Status: **draft, awaiting Fable review** · Date: 2026-07-17
+Repos: `Aerele-RnD/jarvis` (app), `Aerele-RnD/jarvis-persona`
+
+## What happened (a real production transcript, 2026-07-17)
+
+User: *"Create the below tasks and assign to me — test the confirmation ui / test admin v2 / fix
+issues in the list"*
+
+The agent called `get_creation_context("ToDo", …)`, then emitted:
+
+```
+```jarvis-action
+{"kind":"doc","verb":"create","doctype":"ToDo","title":"Create 3 assigned tasks",
+ "summary":"3 open ToDos assigned to Administrator","docs":[{…},{…},{…}]}
+```
+```
+
+**The user got a card that proposed nothing** — the model-written headline, zero field rows, and an
+enabled Confirm. (Click Edit or Preview and you get a single empty Description box. See "What the
+user actually saw" below: the summary card and the Edit panel fail differently, and the smoke must
+check both.)
+
+## Root cause: `docs` is not a `jarvis-action` key
+
+The persona documents exactly one shape for a doc action (`AGENTS.md:122`):
+
+```json
+{"kind":"doc","verb":"create","doctype":"Sales Order","title":"…","summary":"…",
+ "fields":[{"label":"customer","value":"…"}],"tables":[{"fieldname":"items","rows":[…]}]}
+```
+
+`fields` + `tables`. **No `docs`.** Verified: **nothing** in `frontend/src` or `pwa/src` reads
+`docs` off an action block. (The two `args.docs` hits in `actionSummary.js:160/180` are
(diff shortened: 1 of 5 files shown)


venkat102:
Folded into #352 — same subsystem, one review is easier than two.

Both commits are on `feat/confirm-card-primitive` now (`d9bbc54` split `auto`, `832a348` fail loudly) along with the spec, and #352's body covers this half in full. Nothing is lost.

The persona sentence stays separate at Aerele-RnD/jarvis-persona#162 — different repo, and it needs a container roll to take effect.

PR #357, opened 2026-07-17
feat(ai-models): pre-save API-key Test button + real failure reasons on the health chip

kavin-114:
## Motivation

Today, in Settings -> AI models, adding an LLM API-key row has no way to test
the key before saving. The customer saves, and only afterwards sees a bare
"Not working" chip with no reason.

Real case that motivated this PR: a **valid** GLM (Z.ai) key against an
account on a "GLM Coding Plan" (not pay-as-you-go credits). The same key
works fine against a different base URL. z.ai's own API returns a precise
error for the wrong endpoint:

```json
{"error":{"code":"1113","message":"Insufficient balance or no resource package. Please recharge."}}
```

...but the UI only ever showed "Not working", so the customer had no way to
tell a bad key from a valid key hitting the wrong plan/endpoint. That's why
this PR always surfaces the provider's own message, never a bare boolean -
a separate PR (jarvis#355) handles offering the right GLM endpoint; this one
just makes sure the provider's real text reaches the user regardless of
which endpoint they end up on.

## What this adds

**(A) A pre-save "Test" action** on an API-key model row in the Edit panel
(`frontend/src/components/LlmPoolEditor.vue`), backed by a new endpoint
`jarvis.llm_key_probe.test_llm_api_key`:

- Synchronous, no persistence, live HTTP, structured
  `{ok, checks:[{check, ok, detail}], provider, local_endpoint, caveat}` -
  mirrors the shape and gating precedent of `jarvis/selfhost.py`'s
  `test_connection`/`validate_connection`.
- Sends a minimal 1-token completion using whatever
  provider/base_url/model/api_key is currently typed into the panel (nothing
  saved), against the provider's own wire protocol (OpenAI-compatible /
  Anthropic / Gemini), and surfaces the **provider's own error message** on
  failure - not a generic "failed". This is deliberate: a valid key on the
  wrong plan/endpoint and an actually-bad key both come back "failed", but
  only the provider's own text tells them apart.
- **Not** the mutating `PUT /v1/containers/{name}/llm-pool` apply - that
  rotates secrets, rewrites the tenant's `openclaw.json`, and can restart the
  container (10-30s chat outage). This probe never touches the fleet,
  admin_client, or any container, and never persists anything.
- **Caveat handled honestly**: the bench probes from the bench's own network,
  while live tenant chat runs from inside the tenant's bifrost container. For
  public providers (GLM/OpenAI/Groq/...) these agree. For `ollama`/`vllm`
  (local providers), the button still works but is flagged
  `local_endpoint: true`, the frontend shows a disclaimer before clicking,
  and a guard rejection gets a locality-aware message instead of implying a
  guarantee the bench can't make.

**Security (non-negotiable per the request):**
- `base_url` is customer-supplied, so this is an SSRF vector. Rather than
  writing a new guard, this **extends** the existing SSRF guard in
  `jarvis/chat/link_fetch.py` (already used for Personalise link fetches,
  flagged as an open risk in a prior security audit): a new
  `request_pinned()` generalizes the existing `_open_pinned` pinning seam to
  support POST + JSON body + arbitrary headers, reusing the same
  `_validate_url` scheme/host/DNS-rebinding checks and connection pinning -
  no separate/duplicate guard logic.
- The `api_key` is **never** echoed back: not in a check detail, not in a
  raised exception, not in a log. Provider error bodies are scrubbed for a
  literal key match and capped in length before they ever reach the
  response.

**(B) The post-save health chip now surfaces the real reason.** Fleet PR
`fleet-agent#141` (merged, CI-green) adds an optional `detail` string to each
`model_statuses` entry carrying the provider's real error text, and bumps
`contract_version` to 1.12 - so the data this consumes now exists.
`frontend/src/llm/pool.js`'s `apiKeyModelHealth` renders it
(`"Not working: Insufficient balance..."`) in both the chip label (truncated)
and its tooltip, when present - and falls back to today's generic "Not
working" text, **unchanged**, when the field is absent, since the user's own
running fleet-agent predates contract 1.12.

## Verified

- Backend (`bench --site <site> run-tests --module ...`, via
  `PYTHONPATH=<worktree>`): `jarvis.tests.test_llm_key_probe` (30 tests, incl.
  the GLM insufficient-balance case end to end, api_key scrubbing, SSRF
  rejection both via a mocked guard and end-to-end through the *real* guard
  with only `socket.getaddrinfo` mocked, and the whitelisted-endpoint gate),
  `jarvis.tests.test_link_fetch` (34 tests, 5 new covering the extended
  `request_pinned`/`_open_pinned`), `jarvis.tests.test_llm_pool_endpoints`
  and `jarvis.tests.test_selfhost` (unaffected, still green).
- Frontend: `node --test src/llm/pool.test.js` (37 tests, 6 new covering the
  detail-present / detail-absent-fallback / non-string-detail /
  non-failed-status / truncation cases), and a full `vite build` (clean,
  confirms the Vue SFC compiles with every new template binding actually
  defined in the script).
- `ruff check` / `ruff format --diff` (pinned 0.14.10) clean on every touched
  Python file; no em/en dashes in any new user-facing string (grep-verified
  across every changed file).
- Did **not** call the mutating `/llm-pool` endpoint, touch live tenant data,
  or touch the main checkout / live-served site.

## Base branch note

Opened against `beta` by `gh`'s default and retargeted to `develop` per the
current merge convention - `main` no longer exists on this repo; `develop` is
2 commits ahead of `beta` (both unrelated PWA fixes), and this branch's base
commit is an ancestor of `develop`, so the retarget was clean with no rebase.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -174,6 +174,11 @@ export const saveLlmPool = (models, preset = null, routingMode = "failover") =>
 		preset: preset || "",
 		routing_mode: routingMode,
 	})
+// Pre-save "Test" probe for ONE api-key model row (never persists, never
+// touches the fleet/container - see jarvis/llm_key_probe.py). args:
+// {provider, model, api_key, base_url}. Returns
+// {ok, checks:[{check,ok,detail}], provider, local_endpoint, caveat}.
+export const testLlmApiKey = (args) => call("jarvis.llm_key_probe.test_llm_api_key", args)
 
 // --- Onboarding wizard (managed signup + self-hosted connect) ---
 // Arg names mirror the real backend signatures (jarvis/onboarding.py,
--- a/frontend/src/components/LlmPoolEditor.vue
+++ b/frontend/src/components/LlmPoolEditor.vue
@@ -872,18 +872,28 @@ function sourceChip(row) {
 // per-row, since only one panel is ever open at a time - testResult is cleared
 // whenever the panel's row identity changes OR its own provider/model/apiKey/
 // baseUrl fields are edited (a stale green check must not survive an edit).
-const panel = ref({ open: false, mode: "add", uid: null, source: "subscription", addBackups: true, testing: false, testResult: null })
+// testGen additionally guards testApiKeyRow's in-flight request against a stale
+// response landing after an edit (see that function's doc comment).
+const panel = ref({ open: false, mode: "add", uid: null, source: "subscription", addBackups: true, testing: false, testResult: null, testGen: 0 })
 const panelRow = computed(() => rows.value.find((r) => r._uid === panel.value.uid) || null)
 // A removed panel row leaves panelRow null; close rather than render a headless panel.
 watch(panelRow, (r) => { if (panel.value.open && !r) panel.value = closedPanel() })
 // Invalidate a stale Test verdict the instant any field it depends on changes -
 // otherwise editing the key after a failed test would leave the old red result
-// on screen, implying it still applies to what's now typed in.
+// on screen, implying it still applies to what's now typed in. Array-of-getters
+// form (not a joined string key: two different field combos could join to the
+// same string, e.g. provider="a",model="b c" vs provider="a b",model="c") -
+// same idiom AgentsList.vue already uses for a multi-source watch. Bumping
+// testGen here (not just nulling testResult) also ABANDONS an in-flight Test:
+// its response, when it lands, will see the generation mismatch and skip
+// writing testResult/testing (testApiKeyRow's `stale()` guard) - so `testing`
+// must be reset to false HERE too, or the button would stay stuck on
+// "Testing…" until that now-irrelevant response arrives (if it ever does).
 watch(
-  () => panelRow.value && [panelRow.value.provider, panelRow.value.model, panelRow.value.apiKey, panelRow.value.baseUrl].join(" "),
-  () => { panel.value.testResult = null },
+  [() => panelRow.value?.provider, () => panelRow.value?.model, () => panelRow.value?.apiKey, () => panelRow.value?.baseUrl],
+  () => { panel.value.testResult = null; panel.value.testGen++; panel.value.testing = false },
 )
-function closedPanel() { return { open: false, mode: "add", uid: null, source: "subscription", addBackups: true, testing: false, testResult: null } }
+function closedPanel() { return { open: false, mode: "add", uid: null, source: "subscription", addBackups: true, testing: false, testResult: null, testGen: 0 } }
 function isRowEmpty(r) {
   if (!r) return true
   if (r.credentialType === "subscription") return !((r.accounts || []).length)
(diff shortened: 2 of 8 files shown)


kavin-114:
## Code review (4-angle Sonnet review: line-by-line, removed-behavior, cross-file trace, cleanup)

Ran a manual equivalent of this repo's `/code-review` workflow at high effort (the `Workflow` tool itself wasn't available in this session, so I dispatched the same angle structure as parallel Sonnet subagents and verified/applied the findings myself). Fixed in `5f037e8`:

**Real bugs found and fixed:**
1. **Corrupted `watch()`** - a literal NUL byte had landed inside a `.join(" ")` call (tool artifact from an earlier edit), which broke standard `grep`/`ripgrep` on the whole file (silently treated as binary) without breaking the app. Fixed by switching to Vue's native array-of-sources `watch([...], cb)` form, which also matches an existing idiom in `AgentsList.vue` and drops the join (and its theoretical string-collision) entirely.
2. **False-negative "Enter a base URL before testing"** on a legitimately-saved API-key row for a standard provider (OpenAI/Anthropic/...) - `validatePool` never requires a `base_url` outside `openai_compat`/`vllm`/`zai`, so an existing saved row can have a blank one; Test now falls back to the provider's known default endpoint, same as picking the provider fresh already does.
3. **Stale-response race** - `testApiKeyRow` read/wrote `panel.value` across an awaited network call with no identity guard, so a slow response for one row could land after the customer switched to a different row's panel and overwrite the wrong one. Added an object-identity + generation-counter guard (same class of protection the existing OAuth paste-back flow in this file already has via a nonce check), including correctly resetting `testing` to `false` when an edit abandons an in-flight request.
4. Added a "choose a provider" pre-check (only model/key were gated before).

**Hardening:**
- `probe_api_key`'s "NEVER raises" contract could be broken by `_extract_provider_message` on a sufficiently hostile response body from the customer-supplied `base_url` (only `ValueError`/`TypeError` were caught). Wrapped so any exception degrades to a generic message.
- `_open_pinned`'s new `extra_headers` param could theoretically let a caller override the `Host:` header. Confirmed **not** a live SSRF bypass (the actual TCP/TLS destination is pinned independently via `host=ip`/`assert_hostname`, both derived from the validated hostname, not the headers dict), but reordered so `Host`/`User-Agent` are always computed last regardless.

**Checked and confirmed clean:** SSRF-guard ordering in every code path, `_open_pinned`'s backward compatibility with its one pre-existing caller, the `pool.js` `modelVerdict` → `modelVerdictEntry` refactor's behavior preservation, gating parity with `save_llm_pool`, every template↔script binding, the whitelisted-method string, key-scrubbing order, and the desk onboarding bundle (confirmed to be a live re-export, not a stale prebuilt bundle, and to never call the changed function).

Re-verified after the fixes: `jarvis.tests.test_llm_key_probe` (31, +1 new), `jarvis.tests.test_link_fetch` (34, unaffected), `vite build` clean, `node --test src/llm/pool.test.js` (37, unaffected).

PR #360, opened 2026-07-17
ci: guard that beta only accepts PRs from develop

venkat102:
Part of locking down `develop` and `beta`. This is the piece that has to be code — the rest is repository rulesets (config, see below).

## Why a workflow

Rulesets have **no native "restrict the PR source branch" rule**. So "beta only accepts PRs from develop" has to be a **required status check**: this job fails any PR into `beta` whose head is not `develop`.

## Require this on the BETA ruleset ONLY

It is filtered to `pull_request: branches: [beta]`, so on a PR into `develop` it **never runs** — and a required check that never reports leaves a PR **stuck pending forever**.

That trap is live in this repo already: `ci-base-image.yml` is paths-filtered to the Dockerfile, so **`build` does not run on most PRs** (#352 reported only `tests`; #359 reported both only because it touched that workflow). **Requiring `build` would block every normal PR.**

## The ruleset config this pairs with

`jarvis` is public, so rulesets are available on the free plan. The other three repos are private and GitHub answers: *"Upgrade to GitHub Pro or make this repository public to enable this feature."* — so they stay convention-only for now.

**Ruleset `protect-develop`** → target `refs/heads/develop`, enforcement active:
- Restrict deletions
- Block force pushes (non-fast-forward)
- Require a pull request before merging
- Require status checks: **`tests`** — and only `tests`

**Ruleset `protect-beta`** → target `refs/heads/beta`, enforcement active:
- Restrict deletions
- Block force pushes
- Require a pull request before merging
- Require status checks: **`tests`** + **`source-branch`** (this workflow)

Bypass list empty, so nobody pushes directly — including admins.

### DIFF
--- a/.github/workflows/beta-source-guard.yml
+++ b/.github/workflows/beta-source-guard.yml
@@ -1,0 +1,39 @@
+name: Beta source guard
+
+# beta accepts promotions from develop and nothing else. There is no native
+# ruleset rule for "restrict the source branch of a PR", so this job exists to be
+# a REQUIRED STATUS CHECK on beta: the ruleset requires it, and it fails any PR
+# whose head is not develop.
+#
+# Only ever require this on the BETA ruleset. It is filtered to
+# `pull_request: branches: [beta]`, so on a PR into develop it never runs - and a
+# required check that never reports leaves a PR stuck pending forever. (Same trap
+# as requiring `build` from ci-base-image.yml, which is paths-filtered and only
+# runs when the Dockerfile changes.)
+
+on:
+  pull_request:
+    branches: [beta]
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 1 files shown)


venkat102:
Scrapped — not pursuing branch protection for now.

The workflow itself was only ever needed as a required status check for the beta ruleset (rulesets have no native "restrict the PR source branch" rule). Without the rulesets it enforces nothing, so it would just be an extra job on every PR into beta.

Worth keeping in the thread if this comes back:
- `jarvis` is public so rulesets are free; the other three repos are private and GitHub answers "Upgrade to GitHub Pro or make this repository public to enable this feature".
- Require `tests` ONLY, never `build` — ci-base-image.yml is paths-filtered to the Dockerfile, so requiring it would leave every normal PR stuck pending on a check that never runs.

PR #397, opened 2026-07-22
Add brand asset pack with reproducible generator

kavin-114:
## Why

The Jarvis mark existed only as an inline SVG path inside `JarvisMark.vue`. There was no file to hand to docs, slides, marketing, or a designer.

The five PWA manifest icons had the same problem from the other direction: made by hand, no source file, and they have since **drifted from the current brand colours**.

## What

`brand/generate.py` defines the glyph geometry and the two brand colours once, then emits four variants as SVG masters plus PNGs at 16 through 1024:

| Variant | Use |
|---|---|
| `jarvis-mark-tile` | White spark on the gradient square, the app-icon look |
| `jarvis-mark` | Gradient spark, transparent background, for docs and slides |
| `jarvis-mark-mono-black` / `-white` | One colour, for print and dark or busy backgrounds |

No new dependencies. Pillow already ships with Frappe, and the glyph is a straight-line polygon, so no SVG rasteriser is involved. Output is byte-stable, which makes `--check` meaningful.

```bash
python brand/generate.py            # write the pack
python brand/generate.py --check    # verify the committed pack is current
```

## Drift control

Adding a second copy of the glyph creates a real risk against the component that documents itself as the single source of truth. `jarvis/tests/test_brand_assets.py` closes it rather than ignoring it. The test fails if the path in `JarvisMark.vue`, the hex values in `main.css`, and the constants in `generate.py` ever disagree, or if the committed pack goes stale.

It was verified by mutation, not just by passing. Changing `--brand-1` alone, changing the glyph path alone, and corrupting one committed PNG each fail with the intended message, and the suite returns green once reverted.

This deliberately does **not** modify `JarvisMark.vue`. Wiring a build-time import into a shipping component would put runtime risk into an asset-export change.

## Manifest icons, reported not changed

`brand/manifest_icons.py` renders the live PWA icons from the same definition, but only reports by default. All five currently differ from the current brand gradient:

```
committed icon-192   top-left #8575f4   bottom-right #705dea
current definition   top-left #7383fd   bottom-right #8664f8
```

Regenerating them changes the tab and home-screen icon for every customer, so `--write` stays opt-in and is out of scope here. Worth its own decision.

## Out of scope

- No wordmark. Only the spark exists in code, and inventing brand typography without a designer is not this task.
- No changes to `JarvisMark.vue` or the live manifest PNGs.

## Verification

- `ruff check` and `ruff format --check` clean
- `python brand/generate.py --check` reports all 32 assets current
- 4/4 tests pass, and fail correctly under all three mutations
- `git status` confirms `frontend/`, `jarvis/public/`, and `pwa/` untouched

Note: the design doc could not be committed, since `docs/superpowers/` is gitignored in this repo.

https://claude.ai/code/session_01JePsDHsnqevk5xwMTS14kF


kavin-114:
## Review response

Automated high-effort review pass: 7 findings. Five fixed in 3e91405, two rejected with evidence.

### Fixed

**Byte-exact PNG comparison vs unpinned Pillow.** The real one. Pillow is not version-pinned, and its encoder can emit different compressed bytes for an identical image across releases, so a routine Pillow bump would have turned the drift test red with no actual brand drift. `stale_assets()` now compares decoded pixels for PNGs and raw bytes for everything else.

Verified in both directions:

| Case | Bytes | Pixels | Result |
|---|---|---|---|
| Re-encode at `compress_level=1` | 5111B to 11541B | identical | accepted, correct |
| Change one pixel | differ | differ | caught, correct |

**Staleness logic duplicated three times.** That same helper now backs `generate.py --check`, `manifest_icons.py`, and the test, so they cannot disagree about what "current" means.

**Module shadowing.** `manifest_icons.py` imported `generate` by bare name after a `sys.path` insert, which another module of that name could hijack inside a larger bench process. Now binds by explicit path under a unique name, matching how the test already did it. Verified with a decoy `generate.py` on `PYTHONPATH`.

**`report()` always returned 0**, silently contradicting `generate.py`'s `--check` contract. Added `--check` which exits 1 on drift. A bare run still exits 0 deliberately: the live icons are known to have drifted, so a default red exit would block everyone.

**`variant()` if/elif duplicated `VARIANTS`.** Replaced with a `builders()` dict that is now the only place variant names live. `VARIANTS` is gone.

### Rejected

**"icon-192/512 wrongly rendered with rounded corners, corrupting the favicon."** They are declared `purpose="any"` in `pwa/vite.config.js:51-72`, and the live committed `icon-192.png` already has a transparent corner `(0,0,0,0)`. `bleed=False` reproduces the shipped convention rather than corrupting it. Whether "any" icons *should* be rounded is a pre-existing design question, not something this PR changes, and `--write` is opt-in and was never run.

**"`build_assets()` in the test is a CI timing risk."** Measured: 0.82s for `build_assets()`, ~1.1s for the whole module, against a ~210s sharded suite. That is 0.5%.

### Post-fix verification

- `ruff check` and `ruff format --check` clean
- 32 assets regenerate, `--check` green
- 4/4 tests pass; colour drift, glyph drift, and pixel drift each still fail with the intended message
- `manifest_icons.py` exits 0 bare, 1 under `--check`
- `frontend/`, `jarvis/public/`, `pwa/` confirmed untouched

PR #405, opened 2026-07-23
refactor(settings): migrate the settings dialog to frappe-ui

kavin-114:
Migrates the settings dialog shell and all eight panes off the hand-rolled `jv-*` system onto frappe-ui components and semantic Tailwind tokens, then verifies each tab against the live bench.

The settings dialog was the last major surface in the customer SPA still running its own overlay, focus trap, button classes, switch and inline SVG icon set. The app shell around chat and the whole pages surface (Agents, Skills, Macros, Wiki) are already frappe-ui, so opening settings was a visible break in the app.

Net **-221 lines** across 14 files.

## What changed

**Three shared components**, so panes stop cloning the same markup:

- `KvRow` collapses four near-identical read-only rows (`jv-set-row`, `jv-mon-kv`, `jv-acct-kv`, `jv-dsub-kv`) into one (anti-pattern 19).
- `ToggleRow` replaces the hand-rolled `.jv-switch` with frappe-ui `Switch` (anti-pattern 9).
- `SettingsPane` owns the single pane header, the scroll region and the one error surface (anti-patterns 7 and 16).

**Dialog shell** drops the hand-rolled overlay, focus trap, focus-return bookkeeping and Escape listener for frappe-ui `Dialog`, which provides all four. Rail icons become `FeatherIcon` instead of hand-pasted SVG path data (anti-pattern 13), and the three rail blocks collapse into one `NAV` array. `SettingsDialog.vue` goes from 455 lines to 152.

**Eight panes** migrated: General, Usage, Activity, Shortcuts, Connection, Plan and billing, Billing and metering, User usage. Charts in Billing and metering are untouched, including the `useJarvisTheme` import that feeds them their colours.

**`AppShell` takes over the `assets/settings.css` import.** ChatView and OnboardingView render `.jv-btn` / `.jv-iconbtn` from that stylesheet and were only getting them because `SettingsDialog` happened to be mounted alongside. The dependency is now explicit.

## Two errors in design.md, routed around

Flagging these because following the doc literally produces broken output:

1. The legacy porting table maps the primary CTA to `bg-surface-gray-9`. That token **does not exist** on the pinned frappe-ui 0.1.278, where the surface gray scale stops at 7, and every other passage in the same document uses `bg-surface-gray-7`. Used `-7`.
2. Sections 2.6 and 3.4 give contradictory input focus recipes (`focus-visible:ring-2` vs `focus:shadow-sm`). Neither is hand-written here; `FormControl` renders its own focus state.

## Stacking, verified rather than assumed

frappe-ui `Dialog` portals to `<body>` with no z-index, while `ConfirmDialog` relies on `z-index: 200` inside `#app` to open above settings. Verified in the browser that `#app` creates no stacking context (every trigger at its default), so a single `.dialog-overlay { z-index: 60 }` rule in `main.css` clears Sidebar's `z-10`/`z-20` layers while staying below the confirm. Escape-to-close confirmed working after removing the hand-rolled handler.

## Backend verification

All eight panes exercised against the live bench on `site.jarvis`. Seven of eight endpoints return 200 with the expected payload shape:

| Endpoint | Result |
| --- | --- |
| `chat.api.get_usage` | 200 |
| `chat.user_settings_api.get_my_settings` | 200 |
| `account.get_llm_connection_status` | 200 |
| `onboarding.get_llm_config` | 200 |
| `account.get_llm_usage` | 200 |
| `onboarding.get_llm_sync_status` | 200 |
| `chat.user_settings_api.admin_list_user_usage` | 200 |
| `account.get_account` | 417, local env, see below |

Write paths verified to the database, not just the UI:

- Set a monthly limit through the migrated `FormControl` and confirmed `monthly_token_limit: 250000` landed in `Jarvis User Settings`. This was the specific regression risk in swapping a native number input for `FormControl`, since `v-model.number` had to keep yielding an integer rather than a string. It does.
- Flipped "Show tool activity" and confirmed `activity_detail` went 1 to 0 server-side.
- Both values restored afterwards, so the bench is back where it started.

`account.get_account` returns 417 with "admin authentication failed". Traced to `onboarding.py:50`, and it is a **local environment problem, not a code defect**: `site.jarvis` was never onboarded against the admin v2 control plane rebuilt on 2026-07-20, so `Jarvis Settings` holds placeholder `test...` credentials and admin v2 has no tenant row for this bench. The admin plane is reachable and correctly rejects the bogus token. The pane's error path renders correctly (header, single Retry button, message in the error slot), but its happy path is unverified locally and needs a properly onboarded site.

## Defects found and fixed

- **The notify toggle looked broken.** The store gates `notifyEnabled` on `Notification.permission` as well as the stored preference, so the switch renders off while the server row says `notify_enabled: 1`, with nothing on screen explaining the gap. Now carries a hint for both the blocked and unsupported cases.
- **A dead "Delete all" button.** `clearAllHistory` is registered by ChatView at runtime and is genuinely absent on non-chat routes, where the button sat disabled with no explanation. Now says why.
- **Shortcuts tip broke across two lines.** An arrow icon inside an inline `<kbd>` stranded the chip, since `<kbd>` is inline and the icon's svg does not sit on the text baseline.
- **Activity had no empty state**, just a blank area. It is empty most of the time.
- Copy fixes: em dashes removed from user-facing strings, sentence case throughout, no glyph characters in labels.

## Deliberately not done

- `AiModelsPane` and the 3,959-line `LlmPoolEditor` it wraps are excluded. That editor drives live tenant LLM pool provisioning and a regression there is far more costly than a cosmetic inconsistency, so it gets its own PR. Until then the AI models pane visibly differs from its siblings.
- `assets/settings.css` is **not** pruned. 85 of its 112 classes are provably dead, but an automated prune corrupted the file (a comment between rules swallowed the next selector), and chat still renders `.jv-btn` / `.jv-iconbtn` / the `jv-fade` transition classes from it, with no test covering chat styling. Deferred to a follow-up rather than risking a mangled global stylesheet.

## Testing

- `npm run build` green.
- All eight panes walked in the browser in both light and dark.
- Local `bench run-tests --app jarvis` was run, but `site.jarvis` is role polluted from a prior migrate, so **CI is the authority** for anything permission shaped.

---
https://claude.ai/code/session_01WLXuBvzxvSt2WqHQRXdRBz


kavin-114:
## Code review outcome

A high-effort review pass ran against this branch. It found **9 confirmed defects**. Four were real regressions and are fixed in `4b98f01`; five are cleanup items I consciously left, listed below so a reviewer knows they were seen rather than missed.

### Fixed (all four were regressions this PR introduced)

Each was confirmed in the browser before and after the fix, not just reasoned about.

1. **No close button.** Overriding `Dialog`'s `#body` slot replaces the default content, which is where frappe-ui renders `DialogClose`. The dialog shipped with no X at all, leaving Escape and backdrop-click as the only exits. Measured: `closeButtonCount: 0`.
2. **Broken accessible name.** The same override drops `DialogTitle`, so reka's generated `aria-labelledby` pointed at a nonexistent element and screen readers announced an unlabeled dialog. Measured: `aria-labelledby="reka-dialog-title-v-31"`, `labelledByResolves: false`. Now resolves to "Settings".
3. **AI models lost its heading.** That pane is deferred to #406, but it depended on the shared `jv-settings-head` this PR removed, so it rendered with no title or description. Measured: `headings: []`. It now supplies its own header; `LlmPoolEditor` underneath is still untouched.
4. **Unmigrated pane could clip silently.** The pane wrapper was a plain block, so `.jv-settings-body`'s `flex:1; overflow-y:auto` never became a scroll region. Once the pool editor grew past the dialog height its content would have been cut off with no scrollbar. Wrapper is now `flex-col`.

### Accepted, not fixed

These are all real. They are markup duplication rather than defects, and fixing them now would mean another full verification cycle on a PR that is already large.

- Stat-card markup repeated 8x in `UsagePane` and 6x in `BillingMeteringPane`. A `StatCard` component would be consistent with the `KvRow` / `ToggleRow` extractions this PR makes, and is the fairest criticism of the set.
- `<kbd>` chip markup repeated 13x in `ShortcutsPane`.
- `grid-template-columns` duplicated between the header row and data rows in `UsageAdminPane`, where an edit to one and not the other would misalign the table.
- `statusTheme` implemented separately in `ConnectionPane` and `GeneralPane`.

Happy to fold any of these in if a reviewer would rather they land here than in #406.

### Not covered by the review

`get_account` returns 417 on the local bench. That is #408 (local env, never onboarded against admin v2), not a code defect, so the Plan and billing happy path is unverified. Its error path is verified.

kavin-114:
## Local CI: passing

Hosted Actions could not run (org billing exhausted, jobs are refused before any code executes), so these gates were run on a developer machine instead.

| Gate | Result | Detail |
|---|---|---|
| `tests` | **SKIP** | bench resolves jarvis to /Users/kavin/frappe/v16/bench-16/apps/jarvis, NOT the directory under review, so a run here would test the wrong code. Check the branch out in the primary checkout to get a real result. |

Branch `feat/settings-frappe-ui` at `13fd41d`, run 2026-07-24 14:11 IST.

</details>

kavin-114:
## Merged `develop` (13fd41d): conflict resolution notes

Three files conflicted. Both features that landed after this branch cut had modified files this PR rewrites wholesale:

| File | develop side | this branch |
| --- | --- | --- |
| `PlanBillingPane.vue` | #403 downgrade UI, +75 | full frappe-ui rewrite |
| `SettingsDialog.vue` | whitelabel branding, +30 | 455 to 152 lines |
| `GeneralPane.vue` | whitelabel branding, +4 | full rewrite |

The two sides share almost no lines, so a hunk-level resolution would have produced a pane that is half `jv-*` and half frappe-ui with no conflict marker to warn anyone. Resolved instead by taking this branch's rewrite for all three files and re-implementing develop's feature deltas in the surviving idiom.

### What was ported

**Downgrade UI, re-expressed in frappe-ui.** The scheduled-downgrade notice mirrors the cancellation notice already in this pane, and the plan grid reuses the already-migrated upgrade grid rather than a second layout. Resume stays the pane's only solid button, so "Keep current plan" is `variant="subtle"`; the non-revocable case stays a plain Desk text link (anti-pattern 15). The script block carried over unchanged: `downgradePlans`, `scheduledDowngrade`, `scheduledDowngradeNotice`, `doCancelDowngrade`.

**`agentName`** threaded into the notify-toggle help text and the AI models pane description, the two places whitelabel touched that this branch had rewritten. The AI models description moved into the pane here rather than living in the dialog's `DESCRIPTIONS` map.

**BrandingPane** registered in `PANES` and as a single `NAV` entry with a `FeatherIcon`, rather than porting develop's inline-SVG rail button.

### One thing a plain merge would have shipped broken

`BrandingPane.vue` arrived `jv-*`-styled and depended on the shared `jv-settings-head` this PR deletes, so it would have rendered **with no title at all**, and its plain-block wrapper meant `.jv-settings-body`'s `flex:1; overflow-y:auto` never became a scroll region, so the asset pickers would clip silently once they grew. Those are defects 3 and 4 from this PR's earlier review pass reproducing on new code. It now gets the same treatment `AiModelsPane` got: a self-supplied header plus a `flex-col` wrapper. Its body stays legacy and is deferred alongside AI models.

### Evidence

- `pre-commit run --all-files`: **PASS**, all 13 hooks (ruff, prettier, eslint, merge-conflict check), exit 0.
- `npm run build`: **PASS**, clean in 50.78s. Worth stating explicitly because the local-CI report above cannot cover it: its test leg is a `bench run-tests` gate, which SKIPs on a worktree, and this is a frontend-only diff where the build is the gate that carries the signal.
- Diff against this branch's pre-merge state is 87 lines in `PlanBillingPane.vue` and 5 each in the other two, so no formatting churn rode along.

One trap worth recording: `npx prettier` resolves to 3.9.6 while this repo pins **2.7.1** via pre-commit, and the v3 defaults reformat unrelated `defineAsyncComponent` calls. Run the pinned hook, not `npx`.

### Still unverified

Nothing here has been rendered in a browser. The ported downgrade UI's happy path remains blocked by #408: `site.jarvis` is not onboarded against the admin v2 plane, so `get_account` returns 417 and this pane shows its error state. That gap predates this merge and is unchanged by it.

kavin-114:
## Review of `13fd41d` (post-merge), plus CI status

### CI: hosted is green again

Hosted Actions is running on this repo once more, and this PR is green on all six checks at `13fd41d`:

| Check | Result |
| --- | --- |
| `lint` | pass (29s) |
| `coverage` | pass (18s) |
| `tests (1)` .. `tests (4)` | pass (4-way shard) |

That supersedes the earlier local-CI comment on this PR, which was posted while Actions was refusing jobs for billing. Local corroboration also passes (lint clean, coverage 91% against an 85% floor). The local run reports 47 test failures, but it verifies every one of them reproduces on base `818ea6f`, so they are pre-existing local-environment artifacts and not attributable to this branch. Hosted CI is the authority and it is clean.

So there is no gate failing here. The finding below is a visual regression that no CI gate on this repo can see.

### Blocking: the dialog stopped supplying the `jv-` palette, so unmigrated panes lose their theming

This PR's rewrite of `SettingsDialog.vue` dropped the two bindings the old dialog carried on its root:

```
:class="{ 'jv-dark': dark }"
:style="paletteVars"
```

Those are not decoration. The `jv-` custom properties are deliberately **not** on `:root`. `main.css` says so directly, in the comment on the brand tokens:

> Deliberately on `:root` and NOT in theme.js's paletteVars: jv-* vars only resolve inside a subtree that bound paletteVars (the chat/settings/onboarding roots) ... where those vars do not exist.

`ConfirmDialog.vue` documents the same rule, and names this exact file as the example to follow:

> The jv- palette (`--surface`, `--text`, `--red`, ...) is NOT global `:root` ... every jv- surface applies it to its own root via paletteVars + a jv-dark class (**see SettingsDialog**). This dialog is a shell sibling, so it must do the same or its `var(--...)` styling and dark backdrop resolve to nothing.

That comment now points at a file that no longer does it.

Two things make this bite rather than degrade quietly:

1. The dialog portals through `DialogPortal` into `<body>`, which this PR documents in the `main.css` z-index comment. So it is not a descendant of `ChatView`'s root and cannot inherit the palette from it.
2. Of the 161 `var(--...)` usages in `assets/settings.css`, **zero** supply a fallback value. Each one resolves to nothing rather than to a sane default.

Blast radius is every surface in the dialog this PR intentionally left on legacy markup:

| Component | `jv-` usages |
| --- | --- |
| `LlmPoolEditor.vue` (inside AI models) | 345 |
| `BrandingPane.vue` | 38 |
| `AiModelsPane.vue` body | 7 |

Dark mode is where this shows worst: the surrounding dialog chrome is frappe-ui and themes correctly, while the pane inside it loses its backgrounds, borders and text colours. `AiModelsPane`'s own header comment still asserts "The dialog root supplies paletteVars + .jv-dark", which is now stale and would mislead the next person.

The fix mirrors what `ConfirmDialog` already does, on the dialog's content root:

```js
import { useJarvisTheme } from "@/theme";
const { effectiveDark: dark, paletteVars } = useJarvisTheme();
```

```html
<div class="relative flex h-[calc(100vh-8rem)] flex-col sm:flex-row"
     :class="{ 'jv-dark': dark }" :style="paletteVars">
```

Worth stating plainly: this is a static-analysis finding, traced through the variable definitions, the two comments above and the portal behaviour. I have not rendered it in a browser, so please confirm visually before and after. If it turns out the palette reaches the pane by some route I missed, this drops entirely.

### Smaller notes, none blocking

- **A cancelled subscription with a scheduled downgrade renders both notices stacked.** admin-v2 #98 prevents *starting* a downgrade while pending-cancel, but not cancelling after one is scheduled, so both `v-if` blocks can be true at once. Both statements are true, so this is a layout nit rather than a correctness bug.
- **`.dialog-overlay { z-index: 60 }` depends on a frappe-ui internal class name.** It is the right call and the comment justifies going app-wide, but a frappe-ui upgrade that renames that class would silently restore the original stacking bug with nothing failing. A brief note in the CSS would help whoever bumps the dependency.
- **`AiModelsPane`'s stale comment** should be corrected along with the fix above.

### What I checked

Reviewed the net diff against develop: 16 files, +1274/-1379. The three new primitives (`KvRow`, `ToggleRow`, `SettingsPane`) are clean, correctly propped and well documented. Every `v-for` in the changed panes carries a `:key`. The `v-model.number` risk called out in the PR description is properly handled: `saveLimit` and `saveModelLimit` coerce at the boundary with `Math.max(0, Math.round(Number(...) || 0))` rather than trusting the modifier, so a string draft cannot reach the server.

Disclosure: I resolved the develop merge in `13fd41d`, so my read of the downgrade port and the `BrandingPane` header in that commit is self-review and deserves a second pair of eyes more than the rest of the diff.

kavin-114:
### Local CI — `Aerele-RnD/jarvis` PR #405

Tested commit `39d76bfef64d` · ❌ **FAIL** · 59s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 40s |
| `tests` | ❌ FAIL | 1/2 | 15s |
| `coverage` | 💥 ERROR | 3/3 | 2s |

**`tests` — FAIL**

- failed at step **Record what this run actually tested against** (exit 127)

<details><summary>log excerpt</summary>

```
# job: tests
# repo: Aerele-RnD/jarvis
# sha: 39d76bfef64d500d39c91c668c3327f4d581bd83
checked out 39d76bfe -> /ci/workspace/app
OCI runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/ci/workspace//opt/frappe-bench") set in config.json failed: no such file or directory
```

</details>

**`coverage` — ERROR**

> unsupported action 'actions/download-artifact@v4'. Add a shim in ci/jci/engine.py rather than letting the run skip this step.


<details><summary>log excerpt</summary>

```
# job: coverage
# repo: Aerele-RnD/jarvis
# sha: 39d76bfef64d500d39c91c668c3327f4d581bd83
checked out 39d76bfe -> /ci/workspace/app
Python 3.11.15
Requirement already satisfied: coverage~=7.10.0 in /ci/cache/python/3.11/lib/python3.11/site-packages (7.10.7)
!! unsupported action 'actions/download-artifact@v4'. Add a shim in ci/jci/engine.py rather than letting the run skip this step.
```

</details>

---

kavin-114:
### Disregard the ❌ on the local-CI comment above: that is a harness bug, not this branch

The local docker runner failed to execute, so its verdict says nothing about this code. Breaking it down by job:

| Job | Result | What actually happened |
| --- | --- | --- |
| `lint` | **PASS** | 4/4 steps, 40s. Real result. |
| `tests` | **did not run** | The container never started. Not a test failure. |
| `coverage` | **did not run** | Harness lacks a shim for `actions/download-artifact@v4`. |

The `tests` job died on a path-joining bug in the runner, which concatenated an absolute path onto the workspace root:

```
OCI runtime exec failed: unable to start container process:
chdir to cwd ("/ci/workspace//opt/frappe-bench") set in config.json failed: no such file or directory
```

Note the doubled separator in `/ci/workspace//opt/frappe-bench`. Every shard step after it is reported as `skipped`, so no test was executed locally. Worth fixing in the runner, but it is unrelated to this PR.

### The authoritative signal is green

Hosted GitHub Actions is running on this repo again, and it is clean on this exact commit (`39d76bf`):

| Check | Result |
| --- | --- |
| `lint` | pass, 28s |
| `coverage` | pass, 13s |
| `tests (1)` | pass, 3m25s |
| `tests (2)` | pass, 2m57s |
| `tests (3)` | pass, 2m29s |
| `tests (4)` | pass, 2m52s |

Six of six, including the full four-way test shard on GitHub's own amd64 runners. That is the real gate and it supersedes both local runs on this PR.

### One thing hosted CI cannot tell you

The palette regression fixed in `39d76bf` was found by reading the diff, not by any gate, and no gate on this repo would have caught it: it is a visual defect in dark mode with a clean build and clean tests either way. It still wants a confirm in both themes, which is covered by the planned settings e2e pass rather than blocking here.

Navin-S-R (approved):

PR #410, opened 2026-07-23
fix(onboarding): show why setup failed instead of spinning

kavin-114:
Part 3 of 3. Pairs with jarvis-fleet-agent#154 (contract 1.19) and jarvis-admin-v2#102 (the sentences this renders).

## What happened

A customer paid, connected a free-tier ChatGPT account that was out of quota, saw a green dot reading "Account connected", clicked **Start chatting**, and sat on a "Setting up Jarvis" spinner. The backend knew the reason the entire time.

## Four defects

**The `detail` was computed and never rendered.** `is_ready_for_chat` returns a `detail` carrying admin's sentence. `waitUntilReady` read only `r.ready`; `readiness.js` never read `.detail`; and `suspensionNotice` forwarded `detail` **only** for `subscription_suspended`, returning `null` for `container_provisioning`, which is exactly this case. The one string that could have explained the failure was filtered out on the only path that would have shown it.

**The poll budgets were lopsided.** Fifteen minutes waiting for sync to go terminal, then `waitUntilReady(5, 800)`, about four seconds, for readiness, before printing a hardcoded "Still finishing setup" with no backend field consulted. Readiness now gets a proportionate budget via named constants.

**The green dot was fake.** `accountHealth()` returned `{level:"neutral"}` unconditionally in `singleMode`, which is always during onboarding, before looking at any signal, and the CSS painted `--neutral` and `--ok` the same green. An unverified account was pixel-identical to a working one. Green must now be earned by an explicit `verified`.

**Chat was silent too.** A customer who forced past onboarding while not ready landed in an empty room with no explanation.

## One judgment call worth reviewing

The existing "Renew" CTA is now scoped to `subscription_suspended` only. An out-of-quota **LLM** account has nothing to do with the customer's **Jarvis** subscription, so offering "Renew" there would have pushed someone toward a billing page to fix a problem billing cannot fix.

## Known limitation, flagged deliberately

`_admin_customer_facing_reason` recognises admin's sentence by shape (`"Your ... ."`) because there is no structured field on that wire: the customer bench only ever sees the flattened string `admin_client` assembles. A miss degrades to the previous generic message rather than breaking, but a structured field would be better. Worth a follow-up issue.

## Tests

Frontend `node --test`: **159 passed, 0 failed**. Targeted Python `test_unified_llm_config`: **123 tests, 0 failures**. The full app suite has pre-existing local-environment failures (missing `gevent`, Redis throttling) that reproduce identically on the unmodified parent checkout. CI is the authority.

Requires `npm run build` on deploy.


kavin-114:
## Merge readiness: hosted CI green at `f65c024`

No local substitution needed for this PR. Unlike `jarvis-fleet-agent` and `jarvis-admin-v2`, this repo's runners still execute, so the evidence here is real hosted CI rather than a developer machine. Re-verified today against the current head `f65c024`.

| Check | Result | Duration |
|---|---|---|
| `lint` | **pass** | 32s |
| `coverage` | **pass** | 16s |
| `tests (1)` | **pass** | 3m26s |
| `tests (2)` | **pass** | 2m52s |
| `tests (3)` | **pass** | 2m14s |
| `tests (4)` | **pass** | 2m51s |

All four shards green: [run 30001633581](https://github.com/Aerele-RnD/jarvis/actions/runs/30001633581).

Worth noting for whoever is tracking the billing problem: hosted Actions refuses to start jobs for `jarvis-fleet-agent` and `jarvis-admin-v2` (every job dies in under 3 seconds before any code runs), yet this repo's runners work normally. The exhaustion is therefore not uniform across the org and looks like a per-repo setting worth comparing.

The known limitation in the PR body still stands and is unchanged by this run.

Assigned to @kavin-114 for the merge.

kavin-114:
## Merged with `develop`, conflict resolved, hosted CI green at `71a6036`

This branch conflicted with `develop` once it was marked ready. Resolved, and hosted CI is green on the merged commit.

### The conflict was a feature collision, so neither side could simply win

`develop` landed **whitelabel branding** after this branch opened: the wizard's copy now interpolates `agentName` (`@/branding`, `window.agent_name` with a `"Jarvis"` default). Both conflict hunks in `OnboardingView.vue` were the same two strings this branch had already replaced with backend-provided sentences.

Taking either side alone would have dropped a feature. Taking mine would have reverted whitelabelling for those two strings; taking `develop`'s would have reverted the whole point of this PR. So both are kept: the backend sentence logic via `notReadyNote` / `syncStatusNote`, with the agent name threaded into the two fallback strings that moved into `steps.js`.

**Why the agent name is a parameter and not an import.** `steps.js` declares in its header that it holds no Vue and no API calls so it stays cheap to unit-test under `node --test`, where there is no `window` and no `@` alias. Importing `@/branding` there would have quietly broken that contract. Instead the name arrives as an argument defaulting to `"Jarvis"`, and `OnboardingView` passes its already-imported `agentName`. `GENERIC_NOT_READY_NOTE` remains an exported constant and is byte-identical for a non-whitelabeled tenant, so every existing assertion still holds.

### Verification

| Check | Result | Duration |
|---|---|---|
| `lint` | **pass** | 34s |
| `coverage` | **pass** | 13s |
| `tests (1)` | **pass** | 3m37s |
| `tests (2)` | **pass** | 2m29s |
| `tests (3)` | **pass** | 2m21s |
| `tests (4)` | **pass** | 2m11s |

All four shards green on the merged commit: [run 30077149135](https://github.com/Aerele-RnD/jarvis/actions/runs/30077149135). This is real hosted CI, not a local substitute.

Two tests added for the whitelabelled path, including the case that a real backend sentence is never rewritten with a tenant's agent name. The `steps.test.js` suite is 21/21. Worth flagging separately: `ci.yml` does not currently run the JS tests at all, so those 21 ran locally under `node --test` rather than in the run above.

Assigned to @kavin-114. Note this repo's `develop` ruleset allows **merge commits only**, so `--squash` will be rejected here.

venkat102 (approved):

PR #411, opened 2026-07-23
feat(catalog): consume the admin model catalog (backend only)

kavin-114:
Phase 1 PR 3 of 5. **Ships dark:** the frontend still reads its own local literals until PR 4, so nothing user-visible changes here.

Depends on admin-v2 #100 being deployed for the live path; until then this degrades to the bundled fallback, which is the behaviour CI exercises anyway.

## What

`admin_client.get_model_catalog()`, following the existing `get_preset_catalog` pattern (guest fetch, Redis cache, bundled fallback), plus the generated offline bundle.

## Two things that are deliberately not like the preset fetcher

**1. A 5 second timeout, not the 150 second default.** This is read on the CHAT HOT PATH: `chat/api.py:940` inside `send_message` and `:1456` inside `set_conversation_model`. `_post_guest` defaults to `DEFAULT_TIMEOUT_S = 150`. Presets are only read at onboarding time, so they never needed this.

**2. Negative caching.** Caching only successes means a *hanging* admin costs every single chat send a full timeout. A short-lived failure marker makes an outage cost one slow request per minute instead of one per chat turn.

## The bundle is generated, not hand-written

`jarvis/_model_catalog.py` is produced by `scripts/gen_bundled_catalog.py` from admin's `PROVIDER_SEED`. That matters more than it looks: `get_model_catalog()` never returns `[]` (it returns the bundle on any failure), and **CI always runs with admin unreachable**, so this bundle is what every test actually sees. A hand-trimmed copy would silently shrink the catalog during any outage and break `test_chat_api.py:812` and `test_subscription_models.py`.

All 15 providers, including the two override fields that exist because the same provider has different names/ids per surface: Kimi is `Moonshot (Kimi)` on the api-key surface but `Kimi (Moonshot)` on the subscription surface, and Google is `google` fleet-side but `gemini` frontend-side.

Regenerate after any `PROVIDER_SEED` change:
```
python apps/jarvis/scripts/gen_bundled_catalog.py \
  --seed apps/jarvis_admin_v2/jarvis_admin_v2/fleet/provider_catalog.py \
  --out  apps/jarvis/jarvis/_model_catalog.py
```

## Verification

9 new tests green (cache hit, negative caching, short timeout, bundle-mirrors-seed drift check); `test_preset_catalog` still green. ruff clean.

Spec: `model-catalog-selfupdate-spec.md` section 6.4.


kavin-114:
**PR 3 now complete** (Tasks 5, 6 and 7). The earlier push carried only Task 5, which is why the catalog had no caller.

## Added

- **Task 6:** `_subscription_models.py` now reads the catalog. `SUBSCRIPTION_MODELS` and `DEFAULT_MODEL` keep their names as lazily-populated Mappings, so `chat/api.py` and `oauth/api.py` are untouched apart from two words (below). The literals survive as the seed and degraded-mode floor.
- **Task 7:** `get_chat_ui_settings` gains `api_key_models`, replacing the frontend's hardcoded `STATIC_MODEL_SUGGESTIONS` in PR 4.

## Two words that are load-bearing

`chat/api.py` wraps both values in `dict()`:

```python
"subscription_models": dict(_SUBSCRIPTION_MODELS),
"default_models": dict(_DEFAULT_MODEL),
```

Without them the response is silently **wrong, not broken**. Frappe serialises with orjson, which cannot encode a non-dict `Mapping`, so it falls to `json_handler`, which tests `isinstance(obj, Iterable)` at `frappe/utils/response.py:238` before anything else and returns `list(obj)` -- the keys only, no exception. The picker would receive `["OpenAI", ...]` where it expects an object. There is now a regression test asserting each value decodes to a JSON object through the real `json_handler`.

Verified on the live endpoint:
```
subscription_models: py=dict wire=dict OBJECT=True
default_models:      py=dict wire=dict OBJECT=True
api_key_models:      py=dict wire=dict OBJECT=True
```

## The two-label split holds end to end

The same response carries Kimi under **both** its labels, correctly:
- `subscription_models` -> `"Kimi (Moonshot)"` (what `oauth/api.py:_coerce_subscription_model` looks up)
- `api_key_models` -> `"Moonshot (Kimi)"`

Emitting one where the other belongs would coerce a Kimi subscription's model to `""` and default the picker to gpt-5.5.

## Review fixes also included

- `get_model_catalog` genuinely never raises now: the Redis calls were outside the `try`, so a cache blip propagated into `send_message`. The guard logs via `frappe.logger()`, not `log_error`, because the latter writes a DB document and would raise from inside the handler exactly when infrastructure is unhealthy. A test caught that.
- Tests clear the real Redis keys via `addCleanup`, not just `setUp`; previously a passing test left a fake payload cached for its full 6 hour TTL.
- One pinned test was edited by design: `test_subscription_models.py:25` did `json.dumps(SUBSCRIPTION_MODELS)`, which stdlib json rejects for a Mapping. It now asserts on the `dict()`-wrapped copy **and** additionally locks the orjson path, making it a stronger canary than before.

## Verification

`test_model_catalog` 20, `test_subscription_models` 6, `test_chat_api` 63, `test_oauth_api` 64, `test_preset_catalog` 4. All green. ruff clean.

Navin-S-R:
Nice failure-mode design — short timeout + negative caching + never-raises on the hot path, and the `get_model_catalog` tests are thorough. A few things worth addressing before merge:

**1. `dict → Mapping` is a silent-serialization footgun (highest priority).**
Per your own test comment, frappe's orjson + `json_handler` serializes a *bare* `_LazyModelMap` to a JSON list of its **keys**, with no error. The two `chat/api.py` response sites are `dict()`-wrapped and `oauth/api.py` only does `.get()`, so it's correct today — but any future `return {..., "models": SUBSCRIPTION_MODELS}` silently ships keys-only. Prefer making `_LazyModelMap` self-serialize (a hook `json_handler` respects, or handle `Mapping` there) so a bare return is safe, rather than relying on every caller to remember `dict()`. At minimum, add a test that fails if the bare Mapping serializes to anything but its dict.

**2. The new `_LazyModelMap` / `_subscription_rows` logic is untested.**
`test_subscription_models.py` only tweaks the existing serialization test. Nothing covers: resolution from a live catalog, seed fallback when the catalog is empty, the Kimi `subscription_label` vs `label` mapping (the footgun the docstring spends a paragraph on), or `is_default` → `DEFAULT_MODEL`. A few tests with a mocked `get_model_catalog` would close this — the Kimi case especially.

**3. `_subscription_rows()` hand-rolls a cache on `frappe.local`.**
`getattr(frappe.local, "_jarvis_sub_models", …)` — `@frappe.request_cache` is the sanctioned idiom. Also `_LazyModelMap._data()` re-runs `_builder(rows)` on every access (only `rows` is cached, not the built map), so each hot-path `.get()` rebuilds the dict. Cheap to fix by memoizing the built result too.

**4. Bundle drift.** Nothing asserts `_model_catalog.py == gen(PROVIDER_SEED)` (cross-app, so jarvis CI can't), and `test_bundle_covers_every_provider_and_model` hardcodes `len == 15` — a seed addition that skips the generator ships stale silently. Worth a "regenerate the jarvis bundle" step on admin-v2 #100, or a periodic drift check.

**Nits:** version-pinned facts in docstrings ("Verified 2026-07-22 against v7.2.35…") will rot in source — move to the PR/commit. `gen_bundled_catalog.py`: `open(path).read()` without a context manager; the AST fallback assumes a pure-literal seed.

Security/perf look clean — no SQL building, only display fields cross the guest wire, hot path is O(1) Redis apart from the rebuild in #3.

kavin-114:
## Local CI: FAILING

Hosted Actions could not run (org billing exhausted, jobs are refused before any code executes), so these gates were run on a developer machine instead.

| Gate | Result | Detail |
|---|---|---|
| `tests` | **FAIL** | Ran 2086 tests in 124.900s Ran 1568 tests in 98.922s \| coverage 91% (floor 92%) |

Branch `feat/model-catalog-consumer` at `3d27682`, run 2026-07-24 13:04 IST.

</details>

kavin-114:
## Local CI interpretation: FAIL is pre-existing, and the coverage floor is misreported

The automated report says FAILING. Both halves need context.

### Tests: not caused by this branch

27 failures, **none catalog-related** (they are socketio-backend, wiki link-mining and shift/envelope tests). Verified causally rather than by inspection: ran one of the failing modules on both sides.

```
this branch:      test_realtime_handlers -> Ran 19 tests, FAILED (errors=7)
upstream/develop: test_realtime_handlers -> Ran 19 tests, FAILED (errors=7)
```

Identical. This branch is also 0 commits behind `upstream/develop`, so it is not a staleness artifact either.

### Coverage: the floor in the report is not this repo's floor

The report gates at 92% and we measure 91%. But this repo's own CI gates at:

```yaml
coverage report --fail-under=85
```

(`.github/workflows/ci.yml:344`). At 91% this passes the real gate with room to spare. The 92 in the local report does not match what hosted CI enforces here.

Some of the 1-point dip is likely mine: this PR adds `scripts/gen_bundled_catalog.py` (94 lines), a developer tool that no test imports, and there is no coverage omit config in this repo. Happy to add an omit or a smoke test for it if reviewers prefer, but at 91 vs a floor of 85 it is not blocking.

### Gates

| Gate | Result | Detail |
|---|---|---|
| lint | **PASS** | pre-commit clean |
| tests | FAIL, identical to develop | 27 failures, zero catalog |
| coverage | 91% | vs this repo's real floor of **85** |

Catalog tests specifically: `test_model_catalog` 20, `test_subscription_models` 6, `test_chat_api` 63, `test_oauth_api` 64, `test_preset_catalog` 4. All green.

**Not claiming "CI green"** since hosted Actions cannot run. The precise claim is: no new failures relative to develop, and coverage clears the repo's actual gate.

kavin-114:
## Correction: the "Local CI: FAILING" comment above is a tooling bug, not a problem with this PR

Posting this because a reviewer arriving at this PR sees `FAILING` and reasonably hesitates. **Hosted CI on this PR is green: 6 of 6 checks pass** (lint, coverage, and all four test shards) at head `3d27682`, in [run 30075665056](https://github.com/Aerele-RnD/jarvis/actions/runs/30075665056).

That is the authoritative signal here. This repo is the one whose GitHub Actions runners still work; the local runner exists only for `jarvis-admin-v2` and `jarvis-fleet-agent`, where hosted jobs are refused for billing. Running it against this repo added noise rather than signal.

Three separate defects in that local runner produced the misleading report. All three are now fixed:

1. **Wrong coverage floor.** It gated this repo at **92%**, but this repo's own workflow enforces `coverage report --fail-under=85` (`.github/workflows/ci.yml:344`). At 91% this PR clears the real gate with six points to spare. The floor is now 85, cross-checked against each repo's workflow: admin-v2 93 (`ci.yml:128`), fleet 88 (`ci.yml:82`).
2. **Base resolution assumed the wrong remote.** It looked for `origin/develop`, but this repo uses **`upstream`**. So on this repo specifically it could never resolve a base commit, and therefore could not tell a pre-existing failure from a new one. It now tries both remotes.
3. **The bench path did no baseline comparison at all.** Only the pytest path re-ran failures against the base, so a Frappe app got a bare `FAILING` with no classification, which is why the follow-up comment had to be written by hand. That comparison is now automatic.

The manual interpretation comment above reached the right conclusion by hand, and its evidence still stands: the 27 failures are socketio-backend, wiki link-mining and shift/envelope tests, identical on `upstream/develop` (`test_realtime_handlers`: 19 tests, 7 errors on both sides), with nothing catalog-related and the branch not stale.

No action needed on this PR. Flagging it only so the `FAILING` label does not block the merge.

Navin-S-R (approved):

kavin-114:
## Local CI re-run: PASSING (supersedes the earlier FAILING report)

Re-run with the corrected runner. The earlier `FAILING` comment was tooling, not this PR: it gated coverage at 92% when this repo's own workflow enforces `--fail-under=85` (`ci.yml:344`), and it resolved its base against `origin` when this repo uses `upstream`, so it could not classify a pre-existing failure at all. Both are fixed, and the bench path now runs the baseline comparison automatically.



| Gate | Result | Detail |
|---|---|---|
| `coverage` | **PASS** | 91% against a 85% floor |
| `tests` | **PASS** | Ran 2086 tests in 115.003s Ran 1568 tests in 82.274s \| all 30 failure(s) reproduce on base b7146ad, so they are PRE-EXISTING, not introduced here |

Branch `HEAD` at `3d27682`, run 2026-07-24 16:03 IST.

4. **Schema drift.** A bench app under review is loaded via `PYTHONPATH` against the site's EXISTING database. If the branch adds or alters a DocType and the site has not been migrated onto it, tests can fail here for reasons CI (which migrates a fresh site) would never see. Pure-Python branches are unaffected.

</details>

**Hosted CI remains the authority on this repo, and it is green:** 6 of 6 checks at `3d27682` in [run 30075665056](https://github.com/Aerele-RnD/jarvis/actions/runs/30075665056). The run above is corroboration only. Unlike `jarvis-admin-v2` and `jarvis-fleet-agent`, this repo's Actions runners are not blocked by the org billing problem, so a local run here was never needed in the first place.

PR #416, opened 2026-07-24
feat(catalog): frontend and desk read the admin model catalog

kavin-114:
Phase 1 PR 4 of 5, and the **only user-visible one**. Depends on #411 (merged).

This is the last hop. Everything before it shipped dark: the catalog was live and admin-editable, but the picker still read its own hardcoded lists, so adding a model in the desk changed nothing for customers. After this, it does.

## What changes

- **`pool.js`** — `SUB_MODEL_SUGGESTIONS` becomes `subModelSuggestions(payload)`, reading what the API serves. `FALLBACK_SUB_MODELS` remains as the degraded-mode floor.
- **New `get_model_catalog_ui` endpoint** — the pool editor and subscription card never fetched `get_chat_ui_settings`, so this is the plumbing that gap required. It also works in the onboarding wizard, which has no settings call at all.
- **`LlmPoolEditor.vue`** — `STATIC_MODEL_SUGGESTIONS` deleted; the stale `gpt-4o` OpenAI default fixed; all four `defaultSubscriptionModel` call sites now pass the catalog. Until every one of them did, an admin could not change a subscription default at all.
- **`jarvis_account.js`** — the desk page reads the catalog, restoring the xAI and Kimi suggestions it was missing entirely.

## Two things that look wrong but are deliberate

**The connect card still offers exactly two providers.** It gates on a non-empty `auth_profile_id`, NOT on `supports_subscription`. That flag is true for four providers, because cliproxy really does serve xAI and Kimi subscription models. But only OpenAI and Google Gemini can use the paste-back OAuth flow: Kimi is device-code, and admin's `push_oauth_blob` rejects a direct xAI blob. Listing four would ship connect buttons that cannot succeed. The list is data-driven now; it just resolves to two.

**Kimi appears under two different labels in the same response.** `subscription_models` keys it `Kimi (Moonshot)`, `api_key_models` keys it `Moonshot (Kimi)`. Both are correct for their surface. Emitting one where the other belongs coerces a Kimi subscription's model to an empty string and silently defaults the picker to gpt-5.5.

## Also fixes a pre-existing crash

`DirectSubscriptionCard.vue` read `status.value.provider`, but there is no local `status` ref, so it resolved to the browser global `window.status` (a string). `.value` on that is undefined, so the computed **threw every time the paste-back screen rendered** — the entire xAI bare-code flow from #356. Every other reference in the file already used `props.status`. Present identically on develop; found while wiring the catalog in.

## Verification

Local CI, full run on this exact commit `ff22de79`:

| job | result |
|---|---|
| lint | PASS |
| tests-shard1 | 919 tests, 0 failing, 0 errors |
| tests-shard2 | 923 tests, 0 failing, 0 errors |
| tests-shard3 | 951 tests, 0 failing, 0 errors |
| tests-shard4 | 881 tests, 0 failing, 0 errors |
| coverage | **92%** (floor 85), all 4 shards combined |

**3,674 tests, zero failures.** Plus 68 frontend unit tests and a clean `npm run build`.

**This is not a GitHub Actions run.** Hosted Actions is billing-blocked for the org, so this is `jarvis-ci` executing the repo's own `ci.yml` in docker on linux/arm64. It sets no check on this PR. Two structural divergences: architecture (arm64 vs amd64), and credential scope (one token here vs two on GitHub), so token-scope regressions are invisible locally.

Getting this run to happen at all required six fixes to `jarvis-ci`, which could not previously produce a passing run for this repo on **any** commit, including untouched develop. Those ship separately.

Spec: `model-catalog-selfupdate-spec.md` section 6.4.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -182,6 +182,11 @@ export const getDoctypeFields = (doctype) =>
 // --- LLM pool / models config (System-Manager only, gated server-side) ---
 export const getLlmConfig = () => call("jarvis.onboarding.get_llm_config");
 export const getPresetCatalog = () => call("jarvis.onboarding.get_preset_catalog");
+// Admin-managed model catalog slice (api_key_models, subscription_models,
+// default_models) for the pool editor + subscription card. Independent of
+// get_chat_ui_settings so it also works in the onboarding wizard, which never
+// calls that. See jarvis.chat.api.get_model_catalog_ui.
+export const getModelCatalogUi = () => call("jarvis.chat.api.get_model_catalog_ui");
 export const getLlmSyncStatus = () => call("jarvis.onboarding.get_llm_sync_status");
 export const saveLlmPool = (models, preset = null, routingMode = "failover") =>
 	call("jarvis.onboarding.save_llm_pool", {
--- a/frontend/src/components/DirectSubscriptionCard.vue
+++ b/frontend/src/components/DirectSubscriptionCard.vue
@@ -208,14 +208,25 @@ onMounted(async () => {
 // Built-in fallback: subscription providers offered for a fresh DIRECT connect
 // before the catalog fetch lands or if it fails. Model lists mirror
 // jarvis/_subscription_models.py (codex/gemini-cli catalog).
-const SUB_PROVIDERS = [
+const FALLBACK_SUB_PROVIDERS = [
 	{ provider: "OpenAI", models: ["gpt-5.5", "gpt-5.4"] },
 	{ provider: "Google Gemini", models: ["gemini-2.5-pro", "gemini-2.5-flash"] },
 ];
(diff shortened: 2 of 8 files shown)


kavin-114:
### Local CI — `Aerele-RnD/jarvis` branch feat/model-catalog-frontend

Tested commit `ff22de79c915` · ✅ **PASS** · 26m03s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 55s |
| `tests-shard1` | ✅ PASS | 9/9 | 8m34s |
| `tests-shard2` | ✅ PASS | 9/9 | 6m13s |
| `tests-shard3` | ✅ PASS | 9/9 | 4m12s |
| `tests-shard4` | ✅ PASS | 9/9 | 5m58s |
| `coverage` | ✅ PASS | 5/5 | 6s |

---

Navin-S-R (approved):

PR #440, opened 2026-07-26
fix(agents): JF-017 — enforce the delegate capability contract at the bench

Vigneshsekar:
### The defect

A valid, session-bound plugin call from a marketplace-agent delegate run reached
`jarvis.api._run_tool` **without ever being compared against the agent's declared
`tools_allow`**. Before this change `tools_allow` appeared nowhere in `jarvis/api.py`,
`jarvis/tools/_delegate_write_caps.py`, or the `Jarvis Agent Run` doctype.

Consequences:

* The container-side `tools.allow` (rendered into `openclaw.json` by fleet-agent) was the
  only thing bounding a delegate's tool surface. That is **configuration, not
  authorization** — a compromised container or plugin, or anyone holding a leaked per-run
  session bearer, could call **any** registered tool the run's `run_as_user` happened to
  have Frappe permissions for. An auditor whose manifest declares four read tools could
  drive `delete_doc`.
* The write guard read the **current, mutable** `Jarvis Agent Listing.nature` / `.writes`
  at call time, so editing a listing silently re-authorised an **in-flight** run — an
  auditor could be widened into an operator mid-run, with no relaunch.

### Root cause

The run's capability contract was never **snapshotted** and never **enforced bench-side**.
Authority was read from mutable state (the listing) or from a place the bench does not
control at all (the container's config).

### The fix

**Snapshot at launch.** `agent_scheduler._launch_audit` now stamps the whole contract onto
the `Jarvis Agent Run` row in the same insert as the other launch provenance:

| field | source at launch |
|---|---|
| `tools_allow_json` | the bundled registry's declared `tools_allow`, verbatim |
| `capability_nature` | the listing's `nature` at that instant |
| `capability_writes_json` | the listing's `writes[]` at that instant |
| `capability_contract` | `snapshot` |

All four join `_IMMUTABLE_LAUNCH_FIELDS` on the controller, so the run's authorization is
at least as immutable as its provenance (PP-5's `bundle_version` / `preparation_mode` /
`initiating_human`).

**Enforce from the snapshot.**

* `api._dispatch_from_session` resolves the run from the caller's `session_key` and refuses
  any tool outside the snapshot **before** `_run_tool`, returning a `CapabilityDeniedError`
  envelope and writing a structured audit line. Only `jarvis__`-prefixed manifest entries
  can satisfy a bench tool name, so the container-side entries (`exec`, `canvas`,
  `message`) never grant a bench tool. Non-delegate sessions (ordinary chat, macros,
  direct-Python) resolve to `None` and are completely untouched.
* `_delegate_write_caps` reads `nature`/`writes` from the same snapshot, so a listing edit
  changes nothing for an in-flight run — in **either** direction (it cannot widen it, and
  it cannot silently break it either).

**Fail closed.** An empty or unparseable `tools_allow_json` authorises nothing. A run whose
`capability_contract` marker is blank holds **no** authority at all — the snapshot columns
are only trusted when the marker says a launch wrote them.

**Migration.** `v2_07_agent_run_capability_snapshot` stamps every pre-existing run
`legacy`: no `tools_allow` gate, and the write caps keep resolving against the live
listing — exactly the regime those runs launched under, so a deploy cannot kill a run in
flight. The grace is bounded by the patch's own `Patch Log` timestamp: an unstamped run
created **before** the cutover (one that raced the deploy) is treated as legacy, one
created **after** it is refused. New runs always carry a real snapshot.

`Jarvis Agent Run.session_key` also gains a search index — the contract is now resolved
from it on every plugin tool call.

### Tests

New `jarvis/tests/test_platform_capability_contract.py` (36 assertions across 5 classes),
driving the **real** `api._dispatch_from_session` path rather than a stand-in. Each case
fails if the guard is removed:

* **negative** — a registered but undeclared tool (`get_doc`, `delete_doc`) over a valid
  delegate session is refused with `CapabilityDeniedError` and never reaches `_run_tool`;
  the control case proves the same call succeeds for the same user off a delegate session,
  so the refusal is the contract, not the user's permissions;
* **mid-run listing edit** — re-naturing the auditor's listing to Operator with a write
  contract, widening an operator's `writes[]`, and widening the bundled registry all leave
  an in-flight run's authority unchanged; emptying `writes[]` mid-run does not revoke it;
* **write caps from the snapshot** — auditor refused every write, operator confined to its
  snapshotted doctypes;
* **legacy run** — a run stamped `legacy` still executes undeclared tools and still
  resolves its write caps from the live listing;
* **cutover boundary** — blank contract before the recorded cutover is legacy; after it (or
  with no cutover recorded) is refused; blank contract never trusts stray snapshot columns;
* **launch + immutability** — an end-to-end `_launch_audit` stamps the declared contract,
  the controller refuses any later edit of it, and the launched run then refuses an
  undeclared tool over the real dispatch path.

`jarvis/tests/test_platform_write_caps.py` fixtures now stamp the launch snapshot (that is
what the gate reads) and gained a `legacy=True` variant.

Ran locally: `ruff check` clean on every touched file, `ruff format` clean, `compileall`
clean, plus a standalone 20-case exercise of the resolution/denial logic.

**CI green** — lint + all four `run-parallel-tests` shards + coverage. All 31 new tests
executed (shard 3) alongside the existing `test_platform_write_caps`, `test_agent_identity`,
`test_agents_marketplace`, `test_app_learning_agent`, `test_api` and
`test_api_chat_session_header` suites.

### Known behaviour change worth a look

A run whose agent has **no entry in the bundled registry** (e.g. a Deprecated listing whose
registry entry was removed) snapshots an empty `tools_allow` and is therefore refused every
tool. That is the intended fail-closed outcome — such a bundle has no SKILL body to push
either — but it turns a partially-working deprecated run into a hard refusal, and the run
then sits `running` until the stale-run reaper fails it. Refusing at **launch** instead
would be better UX; left out of scope here.

### DIFF
--- a/frontend/src/views/ChatView.vue
+++ b/frontend/src/views/ChatView.vue
@@ -705,7 +705,7 @@
 						     e.g. a delegate refused on its very first call. Rendered here
 						     because the Activity accordion cannot hold it. -->
 						<div
-							v-if="m.role === 'tool' && !m.action_outcome"
+							v-if="m.role === 'tool' && orphanToolFailures[m.name]"
 							class="jv-toolfail"
 							role="alert"
 						>
@@ -729,10 +729,13 @@
 									{{ toolLabel(m.tool_name) }} didn't run
 								</div>
 								<div class="jv-toolfail-msg">
-									{{ toolFailParts(m).message }}
+									{{ orphanToolFailures[m.name].message }}
 								</div>
-								<div v-if="toolFailParts(m).hint" class="jv-toolfail-hint">
-									{{ toolFailParts(m).hint }}
+								<div
+									v-if="orphanToolFailures[m.name].hint"
+									class="jv-toolfail-hint"
+								>
+									{{ orphanToolFailures[m.name].hint }}
 								</div>
 							</div>
 						</div>
@@ -4350,20 +4353,37 @@ const modelsByProvider = computed(() => {
 const currentTitle = computed(
 	() => store.conversations.find((c) => c.name === currentId.value)?.title || "New chat"
 );
-// A FAILED tool call with no assistant turn ahead of it in the thread. The
-// Activity accordion is anchored to an assistant message, so a tool row that
-// arrives before any assistant text has nowhere to hang and renders NOWHERE —
-// which is exactly the shape an autonomous delegate produces when its very first
-// call is refused. Those rows are surfaced inline instead; everything else keeps
-// the accordion behaviour untouched.
-const orphanToolErrors = computed(() => {
-	const out = new Set();
+// A FAILED tool call with no assistant turn ahead of it in the thread, keyed by
+// message name → the copy to render. The Activity accordion is anchored to an
+// assistant message, so a tool row that arrives before any assistant text has
+// nowhere to hang and renders NOWHERE — which is exactly the shape an autonomous
+// delegate produces when its very first call is refused. Those rows are surfaced
+// inline instead; everything else keeps the accordion behaviour untouched. The
+// server writes plain, customer-facing text into error.message / error.hint for
+// this surface, so both are shown as-is.
+const orphanToolFailures = computed(() => {
+	const out = {};
 	let cur = null;
 	for (const m of messages.value) {
 		if (m.role === "user") cur = null;
 		else if (m.role === "assistant") cur = m.name;
-		else if (m.role === "tool" && !cur && !m.action_outcome && m.tool_status === "error")
-			out.add(m.name);
+		else if (m.role === "tool" && !cur && !m.action_outcome && m.tool_status === "error") {
+			let v = m.tool_result;
+			if (typeof v === "string") {
+				try {
+					v = JSON.parse(v);
+				} catch (e) {
+					v = null;
+				}
+			}
+			const err = (v && v.error) || {};
+			out[m.name] = {
+				message:
+					(typeof err.message === "string" && err.message.trim()) ||
+					"This step couldn't be completed.",
+				hint: (typeof err.hint === "string" && err.hint.trim()) || "",
+			};
+		}
 	}
 	return out;
 });
@@ -4373,7 +4393,7 @@ const visibleMessages = computed(() =>
 		// inline in the thread; every OTHER role=tool row belongs to the collapsed
 		// Activity accordion, not the main thread — except an orphaned FAILURE,
 		// which the accordion cannot hold.
-		if (m.role === "tool") return !!m.action_outcome || orphanToolErrors.value.has(m.name);
+		if (m.role === "tool") return !!m.action_outcome || !!orphanToolFailures.value[m.name];
 		if (m.role !== "user" && m.role !== "assistant") return false;
 		// Hide a blank streaming placeholder. The live "Working on it…" indicator
 		// below the thread already renders the assistant logo + status for the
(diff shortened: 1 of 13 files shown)


Vigneshsekar:
## Round-2 remediation (UX P0-2 / P1-2 / P1-3, adversarial P2-1 / P2-5)

Pushed as `195ce983`. The reviews' core complaint was that this fix fails **silently and dishonestly**: a run whose contract authorises no bench tool is refused on every call — `record_agent_run` included — so it can never finalize itself, and three hours later the stale-run sweep stamps *"run exceeded max duration"* on it. That message is false, and it sends the reader after a timeout that never happened.

**1. Refuse at LAUNCH (root cause, supersedes the reaper-message fix).** `_launch_audit` now resolves the bundled registry's declared surface **before** it inserts anything and throws when it is empty — no conversation, no run row, no bearer, and a message the human can act on (*"This agent's bundle declares no tools… nothing was started."*). The contract builder takes the already-resolved list so the registry is still read exactly once.

**2. Fail the run at the first refusal (adversarial P2-1).** `tool_denial` now returns a structured refusal with a `fatal` flag, set when the contract authorises **no bench tool at all** (blank contract, empty/unparseable snapshot, or a surface that is container-side only). The dispatcher terminalizes the run immediately through a new `agent_scheduler.fail_run`, which is the reaper's own failure tail factored out (`_terminalize_failed`) plus a compare-and-set under the run-row lock — a concurrent `record_agent_run` finish is never overwritten. The reaper now calls the same helper, so there is one run-failure path, not two.

**3. Deploy-window grace (adversarial P2-1).** `_is_pre_patch` widens to `patch_cutoff + LEGACY_GRACE_SECONDS` (6h, documented at the constant). Migrate finishing and the last worker restarting are not the same instant, and *both* deploy orders produce blank-contract rows in that window — they are deploy artefacts, not forgeries. Bounded, and shorter-lived than the runs it covers.

**4. Plain-language chat copy (UX P1-2).** The transcript receipt and the delegate's envelope are now written separately. Chat gets *"This agent tried to use a tool it isn't allowed to use, so the step was skipped."* plus the remedy in `hint`; the contract wording (*"…not in the capability contract snapshotted for this run at launch"*) stays in the audit trail and in what the delegate itself reads.

**5. Denied calls always render (UX P1-3).** `ChatView.vue`'s Activity accordion is anchored to an assistant message, so a tool row with no preceding assistant text rendered **nowhere** — exactly the shape an autonomous delegate produces when its first call is refused. Orphaned error-status tool rows now render inline (message + remedy). Everything else keeps the accordion behaviour untouched.

**6. Vocabulary (adversarial P2-5).** `CapabilityDeniedError` is a real class in `jarvis/exceptions.py` (deliberately *not* a `PermissionDeniedError` — no administrator can grant the tool mid-run) with an `_ERROR_HINTS` entry. Note: the openclaw plugin relays a failed tool call to the model as `"<code>: <message>"` and drops every other envelope field, so the delegate's *"retrying, renaming the tool or changing the arguments will not help"* instruction is in the **message**; `hint` keeps its house role as the human-facing remedy.

### Tests
12 new, 43 total in `jarvis.tests.test_platform_capability_contract` — all green, plus every neighbouring module that touches the changed code:

`test_platform_capability_contract` 43 · `test_platform_write_caps` 12 · `test_platform_launch_provenance` 4 · `test_agent_identity` 7 · `test_agents_marketplace` 36 · `test_app_learning_agent` 59 · `test_agent_dashboard_wire` 7 · `test_platform_min_apps` 14 · `test_api` 28 · `test_skill_tools` 29 · `test_platform_writeback` 30 · `test_platform_activation` 13 · `test_part4_security` 22 · `test_actions_api` 27 · `test_action_error_detail` 18 · `test_registry` 5 · `test_exceptions` 2 — **356 backend tests, 0 failures**. Frontend: `vite build` clean, `vitest run` 43/43.

### Deliberately not done
A cron-loop skip for a tool-less bundle was written and then **reverted**: it reordered the scheduler's documented skip reasons and broke three tests in `test_app_learning_agent`. The launch throw covers both entry points; the scheduler's existing handler records a failed run and retries next hour. Filed as a follow-up below.

PR #441, opened 2026-07-26
fix(plugin-auth): JF-014 — reject unsigned plugin requests by default

Vigneshsekar:
## Defect (JF-014, High)

`jarvis/_plugin_auth.py` validated the Phase-2 HMAC signature **only when at least one
of the three signature headers was present**:

```python
if sig or nonce or ts_str:
    ...validate (partial headers -> 400)...
else:
    frappe.logger().info("plugin_auth: legacy bearer-only request ...")   # accepted
```

A `call_tool` request that carried **none** of `X-Jarvis-Signature` /
`X-Jarvis-Nonce` / `X-Jarvis-Timestamp` was therefore accepted on bearer + session
alone, with nothing but an info log to show for it. Consequences:

- A captured `agent_token` replayed **arbitrary unsigned tool bodies** against any
  session key, for as long as the token lived — the exact blast radius the signature,
  the nonce dedup and the ±60s timestamp window exist to close.
- The check was trivially downgradable: partial headers were rejected, so an attacker
  simply sent **zero** signature headers instead of two.
- Since the IP allowlist was removed (2026-06-19), the signature was the only layer
  standing between a leaked bearer and full tool execution as the session's user.

## Root cause

A migration-era compatibility fallback. When the signature shipped, no plugin build
emitted one, so the bench accepted unsigned requests during rollout — and the sunset
was never scheduled or executed. Every managed container in the field now runs a
plugin build that signs **every** call (`signRequest` in the plugin's
`frappe-client.ts` is on the unconditional path in `callFrappeTool`), so the fallback
no longer protects anything we ship; it only widened the attack surface.

## Fix

1. **Default is ENFORCE.** No signature headers now rejects with the same fail-closed
   shape as partial headers — `400` / `InvalidArgumentError` — plus an Error Log row
   (`plugin_auth: unsigned request rejected`) carrying the caller IP and session for
   triage. Partial headers keep their existing reject and message.
2. **One documented escape hatch** for self-hosted benches running an un-upgraded
   plugin: the site_config key `jarvis_plugin_allow_unsigned`. Read with an explicit
   `0` default; only an explicit on-literal (`1`, `true`, `yes`, `on`) re-opens the
   path. Absent key, `0`, `""`, `None`, `False`, junk values, or a conf read that
   raises all mean ENFORCE. (Deliberately a site_config key, not a Check field —
   the v16 `get_single_value`-on-Check gotcha makes "NULL means default" unreadable,
   and a downgrade switch should be an explicit, greppable operator act.)
3. **Observability before the flag disappears.** Every request the hatch lets through
   WARN-logs the conf key, the caller, the session and the `2026-10-01` removal date,
   and bumps a per-day Redis counter `jarvis:plugin_unsigned_allowed:<YYYY-MM-DD>`
   (7-day TTL) so an operator can watch the stragglers drain to zero before turning
   the flag off. The counter is best-effort: a cache failure never fails a request
   auth has already decided to allow.
4. **The escape hatch does not relax the partial-header rule** — it re-enables
   *unsigned* requests, never half-signed ones.
5. Module docstring documents the rollout story, the flag and the sunset.

No change to the signed path, the bearer/expiry/session checks, or the rate limiter.

## Tests

New `TestJF014UnsignedPathRetired` in `jarvis/tests/test_api.py`:

- `test_unsigned_request_rejected_by_default` — the enforcement guard: delete the
  reject branch and this test fails.
- `test_off_or_junk_flag_values_keep_enforcement` — fail-closed across
  absent / `0` / `"0"` / `""` / `None` / `False` / `"false"` / `"no"` / `"maybe"`.
- `test_partial_signature_headers_rejected_regardless_of_flag` — flag off *and* on.
- `test_unsigned_allowed_when_flag_on_and_warns_with_sunset_date` — accepted, and the
  WARN names the conf key and the removal date.
- `test_legacy_allowed_requests_are_counted` — the per-day counter increments.
- `test_signed_request_still_accepted_with_flag_off` — no regression on the path
  every shipped plugin build uses.

The pre-existing plugin-auth tests posted bearer-only requests, which is no longer a
valid request shape. They now sign the way the plugin does — a shared
`sign_plugin_headers` helper mirrors the plugin's canonical string
(`session | sha256(body) | nonce | timestamp`, HMAC-SHA256 keyed by the gateway
token) with a fresh nonce per call. Tests whose subject is the *reject* path
(wrong token, missing session) are untouched: they fail before the signature step.

Run locally on the branch: `ruff check` + `ruff format --check` on the three touched
files, `compileall`, and a standalone harness that drives `validate_plugin_request`
with a faked `frappe` across all 24 accept/reject permutations above (all pass).
`jarvis.tests.test_api` and `jarvis.tests.test_api_chat_session_header` need a bench
run; full-suite CI covers both.

## Operator notes / rollout

- Managed fleet: no action. Every pod's plugin signs.
- A self-hosted bench on an old plugin build starts getting `400 InvalidArgumentError
  "signature headers required"` on tool calls, with one Error Log row per attempt.
  The fix is to upgrade the plugin; the stopgap is
  `bench --site <site> set-config jarvis_plugin_allow_unsigned 1`, which is scheduled
  for removal on 2026-10-01.

### DIFF
--- a/jarvis/_plugin_auth.py
+++ b/jarvis/_plugin_auth.py
@@ -45,10 +45,17 @@ unsigned bearer-only requests). It exists only for self-hosted benches
 whose openclaw plugin has not been upgraded yet, it WARN-logs the
 sunset date on every request it lets through, and it bumps a per-day
 Redis counter (``jarvis:plugin_unsigned_allowed:<YYYY-MM-DD>``) so an
-operator can confirm the field is quiet before turning it off. Requests
-rejected while the flag is off land in the Error Log under
-"plugin_auth: unsigned request rejected". The flag itself is scheduled
-for removal on 2026-10-01 (``_UNSIGNED_SUNSET_DATE``).
+operator can confirm the field is quiet before turning it off. While the
+hatch is on, ``unsigned_escape_hatch_status`` surfaces that count as an
+advisory row in the self-host "Test connection" checks (jarvis.selfhost)
+so the operator sees it in the product, not only in the logs. The flag
+itself is scheduled for removal on 2026-10-01 (``_UNSIGNED_SUNSET_DATE``).
+
+Requests rejected while the flag is off land in the Error Log under
+"plugin_auth: unsigned request rejected", throttled to one row per
+session per minute: an un-upgraded plugin retries every tool call, so an
+unthrottled audit would fill the Error Log with thousands of copies of
+one fact. The reject itself is never throttled.
 """
 
 from __future__ import annotations
@@ -67,7 +74,10 @@ _NONCE_REDIS_PREFIX = "jarvis:plugin_nonce:"
 _ALLOW_UNSIGNED_CONF_KEY = "jarvis_plugin_allow_unsigned"
 _UNSIGNED_SUNSET_DATE = "2026-10-01"
 _UNSIGNED_COUNTER_PREFIX = "jarvis:plugin_unsigned_allowed:"
-_UNSIGNED_COUNTER_TTL_S = 7 * 24 * 60 * 60
+_UNSIGNED_COUNTER_DAYS = 7
+_UNSIGNED_COUNTER_TTL_S = _UNSIGNED_COUNTER_DAYS * 24 * 60 * 60
+_UNSIGNED_AUDIT_GATE_PREFIX = "jarvis:plugin_unsigned_audit:"
+_UNSIGNED_AUDIT_GATE_TTL_S = 60
 
 
 class PluginAuthError(Exception):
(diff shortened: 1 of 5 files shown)


Vigneshsekar:
## Review remediation (round 1) — 52a0627e

Addresses the UX and adversarial review findings against this PR. **Enforcement semantics are unchanged**: unsigned is still a fail-closed 400 `InvalidArgumentError`, the escape hatch is still the only way through, and it still enforces on absent / `0` / junk / a raising conf read.

**UX P0-1 — remedy in the reject message.** `jarvis/api.py` returns the `PluginAuthError` envelope before `_dispatch_from_session`, so the failure never becomes a chat message, and there is no Settings surface for the conf key — the 400 is the operator's *only* signal. It now carries the remedy:

> `signature headers required; provide all of X-Jarvis-Signature/X-Jarvis-Nonce/X-Jarvis-Timestamp. Upgrade the openclaw plugin to a signing build, or run `bench --site <site> set-config jarvis_plugin_allow_unsigned 1` as a stopgap until that flag is removed on 2026-10-01.`

**Adversarial P2-7 — throttle the reject audit.** A bench with an un-upgraded plugin rejects *every* tool call, so each one wrote an Error Log row. The row is now gated to one per session per minute (Redis `SETNX` on `jarvis:plugin_unsigned_audit:<session>`, 60s TTL). The reject itself is never throttled, and the gate **fails open** when the cache is unreachable so a Redis outage can't make the audit trail quieter than it should be.

**UX P1-1 — operator visibility while the hatch is on.** `unsigned_escape_hatch_status()` reports the conf key, the removal date and how many unsigned calls were allowed across the counter's 7-day window; `jarvis.selfhost.config_checks` renders it as a `plugin_signature_enforced` advisory row in the existing self-host "Test connection" list (Onboarding SPA, Jarvis Account page and the Jarvis Settings dialog all already render advisory rows — no new page). The row appears **only** while the hatch is on, so an enforcing bench grows nothing. The count is read back from the same raw keys the allow path `incr`s, and says "call count unavailable" rather than a false `0` when the cache is down. Being advisory, it never flips the connection verdict.

**Tests.** `TestJF014UnsignedPathRetired` gains the remedy-text guard, the throttle (5 rejects → 1 row; a second session gets its own row; the fail-open path), and the escape-hatch status cases. `jarvis/tests/test_selfhost.py` gains `TestUnsignedPluginAdvisory` for the row's copy, its absence while enforcing, the unknown-count path, advisory-ness, and a guard that a status-lookup failure can never break "Test connection".

**Not addressed here (out of scope for this PR, flagged for the ledger):** adversarial P2-6 is a wording correction to the wave report, not to this code — the module docstring already states plainly that the HMAC is keyed with the same secret as the bearer, so JF-014 buys replay-resistance and freshness, not protection against `agent_token` compromise.

PR #443, opened 2026-07-26
(For Logesh Review) fix(mobile): JF-016 — per-device, revocable mobile credentials

Vigneshsekar:
## Defect (JF-016, verified on `develop`)

`jarvis/mobile/auth.py:get_mobile_token` returned the user's **account-wide**
Frappe `api_key`/`api_secret` to the phone.

* Not device-scoped — two handsets share one credential.
* Logging out revoked nothing; the credential outlived the session forever.
* Killing a lost phone meant rotating a secret that **every other API
  integration the user owned** was also using.
* No expiry, no device inventory, no way to see or cut off a single device.

## Root cause

Pairing minted (or reused) the *account* credential instead of a *device*
credential. There was no per-device identity to revoke, so revocation had to be
account-wide by construction.

## The fix

**New `Jarvis Mobile Device` doctype** — user, client-reported label/platform,
public `token_id`, `secret_hash`, `enabled`, `last_used`,
`revoked_at`/`revoked_by`/`revoked_reason`. Only
`HMAC-SHA256(key=token_id, msg=secret)` is at rest (HMAC rather than
`sha256(token_id + ":" + secret)` — plain concatenation is ambiguous, `("a",
"b:c")` and `("a:b", "c")` would collide). The plaintext secret is returned
exactly once, at pairing, and never stored, so a database dump yields no usable
credential. Perms: read-`if_owner` for Jarvis User, full for System Manager;
`in_create` so nothing hand-creates a row in Desk.

**Token wire format `jmd:<token_id>:<secret>`.** Three colon-separated segments
is the integration point, not cosmetics. Core's
`frappe.auth.validate_auth_via_api_keys` does
`api_key, api_secret = auth_token.split(":")`, which raises `ValueError` on a
three-segment token — swallowed by that function's
`except (AttributeError, TypeError, ValueError): pass`. So core cleanly
*declines* the token and `validate_auth_via_hooks` gets to run. A two-segment
`key:secret` format would instead have raised `AuthenticationError` for an
unknown key **before any app hook could run**; there is a negative-control test
for exactly that, so nobody "simplifies" the format later.

**`auth_hooks` entry `jarvis.mobile.device_auth.authenticate_device_token`** —
constant-time compare, refuses a disabled row or a disabled user, restores
`form_dict` across `frappe.set_user` (core's api-key path does the same, and
skipping it drops every argument of the request), yields to an already
established cookie session exactly as core does, and **fails closed**: it simply
does not set a user, and `validate_auth` turns a presented-but-unusable
Authorization header into a 401 on its own. Cost for every non-mobile request is
one header read plus a prefix test.

**Realtime is covered.** Frappe's socket.io middleware forwards the client's
`Authorization` header to `/api/method/frappe.realtime.get_user_info`, an
ordinary HTTP request, so the same hook authenticates the polling handshake.
`frappe.set_user` leaves `session.data.user_type` empty and core already falls
back to a cached `User.user_type` read for precisely that case — covered by a
GET-shaped test.

**`last_used`** is stamped at most once per 5 minutes per device (one write per
authenticated request would be a hot-path disaster), with its own commit on
read-only requests because Frappe rolls those back at request end. Best-effort:
it can never raise into auth.

**Revocation + inventory** — `revoke_mobile_device`, `revoke_all_mobile_devices`
(with `keep_current`), `list_mobile_devices`. Session user only, own rows only;
another account's device is indistinguishable from a nonexistent one, so the
endpoint cannot be used to probe someone else's inventory. Pairing is throttled
per user (per user, not per IP, so an office behind one NAT is not one budget),
live devices per user are capped with the oldest pruned, and every mint, prune
and revocation is audit-logged; the durable trail is the row itself plus its
Version history.

## Back-compat (no forced migration)

* The global `api_key`/`api_secret` is **no longer read or written** here and
  nothing revokes it — existing API consumers are completely untouched, and
  ordinary Frappe token auth keeps working for them.
* The response still carries `api_key`/`api_secret` because shipped app builds
  join them into `Authorization: token <api_key>:<api_secret>`. `api_key` is
  `jmd:<token_id>` and `api_secret` the device secret, so that join reproduces
  the device token exactly — an un-updated app upgrades onto a revocable
  credential with **no client change**. Newer builds read `device_token` +
  `device`.

## Tests

`jarvis/tests/test_mobile_device_auth.py` — 29 tests, all green:

* pairing issues a distinct token per device; the legacy `key:secret` join
  reproduces the device token; the response never carries the global secret and
  leaves the global keypair byte-identical; label/platform bounded; Website user
  and Guest refused; pairing throttled; device cap prunes the oldest.
* core's api-key path declines a device token **and** the two-segment negative
  control still raises in core; `parse_token` rejects a Frappe keypair.
* hook: valid token authenticates; tampered secret, unknown id, non-`token`
  scheme and disabled user all fail closed; cookie session wins; `form_dict`
  survives; GET realtime-handshake shape authenticates.
* revoke-one kills exactly that device and leaves the user's other device, the
  other user's device and the global keypair working; idempotent; cross-user and
  unknown both refused identically; revoke-all; revoke-all sparing the caller;
  inventory is own-rows-only and carries no hash.
* storage: no plaintext anywhere in the row, nothing in `__Auth`, digest matches
  the keyed HMAC, salt binding unambiguous; `last_used` throttling in both
  directions.

Guard suites re-run green against this branch: `test_whitelist_annotations`,
`test_part4_security` (endpoint-gating sweep), `test_role_gates`,
`test_chat_endpoint_gating`.

**This branch adds a DocType, so the review bench needs `bench --site <test
site> migrate` before `run-tests`.**

## Client

`Aerele-RnD/jarvis_mobile` branch `fix/jf016-per-device-mobile-tokens` stores
the device token + id, revokes on logout, and clears a stale credential before
sign-in.

---

# Round 2 — adversarial review remediation (P1-6, P1-7, P2-11, P2-12)

Verdict was **SHIP-WITH-P1S**: the mechanism checks out, but a stolen phone
could outlive its own revocation, so the fix did not deliver the revocation it
promised. Four changes.

## 1. A device token may not mint a sibling (P1-6)

`get_mobile_token` now refuses a request that is itself authenticated by a
device token:

> Sign in with your password to pair this device.

The original justification — "shipped app builds re-mint while holding a
possibly stale token" — was wrong, and the client in the paired PR refutes it:
`mintToken()` is called from exactly one place, `login()`, and `login()` begins
by clearing the stored token. Every real pairing is authenticated by the
password just typed. Without the guard, a stolen token mints siblings that
survive revoking the phone they came from, and "kill the lost handset" means
nothing.

## 2. Sign-out-everywhere always includes the caller (P1-7 scenario B)

`keep_current` is **gone** — from the endpoint and from
`device_auth.revoke_all`. Sparing the caller made one POST an instant-lockout
primitive: evict every legitimate device, keep your own. A caller who nukes
their own access along with everybody else's is acceptable; one who preserves it
is not.

Removing the parameter is back-compatible: `frappe.get_newargs` drops kwargs a
whitelisted function does not declare, so a caller still sending
`keep_current=1` gets the (now correct) full sweep rather than a TypeError.

The endpoint stays callable **from a device-token session** on purpose — that is
what makes "sign out everywhere" a real answer on the phone the user still
holds, which is what the UX review needed for the lost-phone story.

## 3. Revoke-one from a device session may target only itself

Not separately filed, but it falls straight out of #2: with revoke-all fixed and
revoke-one unrestricted, the same eviction just runs one row at a time. So a
request authenticated by a device token may revoke only that device. Selective
remote revocation is a password-session act — the Jarvis web app / Desk, where
the session is authenticated by a password the thief does not have.

`list_mobile_devices` stays open to device sessions: the phone's Devices screen
is built on it, it carries no secret, and a caller holding a device token
already has the user's whole API surface.

P1-7 scenario A (mint 10, wait, mint 10 more, let `_prune` evict the victim) is
closed by #1 — there is no minting from a device token at all.

## 4. Hygiene (P2-11, P2-12)

* **Revoked rows are no longer immortal.** New daily job
  `jarvis.mobile.device_auth.prune_revoked_devices` deletes **disabled** rows
  past `REVOKED_RETENTION_DAYS = 90`, batched at 500 per pass. Rows disabled
  outside the module (a System Manager unticking `enabled` in Desk) carry no
  `revoked_at`, so the sweep falls back to `modified`. A live credential is
  never touched.
* **Docstring truth.** The controller claimed "SHA-256 of `<token_id>:<secret>`"
  where the code is HMAC — corrected, since that file is the first thing a
  reviewer reads.
* **auth_hooks ordering (P2-11)** is now recorded in `device_auth`'s module
  docstring: `get_hooks` iterates `installed_apps` order, so an app whose auth
  hook decides from `frappe.session.user` — helpdesk's `authenticate`, under
  `block_endpoints` — must install after jarvis. Not reachable on any bench
  today; invisible from either app alone, hence written down.

Not addressed, still open by design: **token expiry** (P2-12) — the motivation
lists it as a defect this fix does not close.

## Tests

`jarvis.tests.test_mobile_device_auth` is now **37 tests (was 29), all green**.
New:

* a device token cannot mint a sibling — and no row is created;
* a password session still pairs (the negative control for that guard);
* a device token cannot revoke a sibling — the victim's device survives;
* a device token *can* revoke itself (the ordinary sign-out path);
* a password session can revoke any of its own devices;
* revoke-all from a device session takes the caller with it;
* revoke-all from a password session kills every device;
* the inventory is readable from a device session and marks the caller;
* old revoked rows are deleted, recent ones and live ones are not;
* rows disabled outside the module are not immortal.

Guard suites re-run green against this branch: `test_whitelist_annotations`
(1), `test_part4_security` (22), `test_role_gates` (6),
`test_chat_endpoint_gating` (1). `ruff check` + `ruff format --check` clean on
every touched file; no whole-file reformatting.

## Client

The paired `jarvis_mobile` PR (#7) grows the Devices section and the
signed-out toast that the UX review blocked on. Merge **this** first.

### DIFF
--- a/jarvis/hooks.py
+++ b/jarvis/hooks.py
@@ -383,6 +383,17 @@ scheduler_events = {
 	],
 }
 
+# ---------------------------------------------------------------------------
+# Per-device mobile credentials (JF-016)
+# ---------------------------------------------------------------------------
+# The Jarvis mobile app authenticates with a REVOCABLE per-device token
+# (`Authorization: token jmd:<token_id>:<secret>`) instead of the user's
+# account-wide Frappe api_key/api_secret. Core's api-key parser declines a
+# three-segment token, so this hook — which Frappe runs right after it — is
+# what resolves the token to a user. Cheap for every other request: one header
+# read plus a prefix test. See jarvis/mobile/device_auth.py.
+auth_hooks = ["jarvis.mobile.device_auth.authenticate_device_token"]
+
 # Python type annotations on whitelisted endpoints
 # ------------------------------------------------
 # Auto-update controller files with type annotations + require all
--- a/jarvis/hooks.py
+++ b/jarvis/hooks.py
@@ -374,6 +374,10 @@ scheduler_events = {
 		# bench's month-to-date per-user + per-model usage rollup to admin. Self-
 		# gating (skips self-hosted / unconfigured / not-onboarded); never raises.
 		"jarvis.chat.usage_push.push_usage_rollup",
+		# JF-016 hygiene: revocation only flips `enabled`, so the mobile-device
+		# table is append-only without this. Deletes DISABLED rows past the
+		# 90-day retention window; live credentials are never touched.
+		"jarvis.mobile.device_auth.prune_revoked_devices",
 	],
 	"weekly": [
 		# Wiki v2 health check: deterministic lint over Active pages
--- a/jarvis/jarvis/doctype/jarvis_mobile_device/jarvis_mobile_device.json
+++ b/jarvis/jarvis/doctype/jarvis_mobile_device/jarvis_mobile_device.json
@@ -1,0 +1,139 @@
+{
+ "actions": [],
+ "allow_rename": 0,
+ "autoname": "hash",
+ "creation": "2026-07-26 09:00:00.000000",
+ "doctype": "DocType",
+ "engine": "InnoDB",
+ "field_order": [
+  "user",
+  "device_label",
+  "platform",
+  "enabled",
+  "last_used",
+  "token_id",
+  "secret_hash",
+  "revoked_at",
+  "revoked_by",
+  "revoked_reason"
+ ],
+ "fields": [
+  {
+   "fieldname": "user",
+   "fieldtype": "Link",
+   "label": "User",
+   "options": "User",
+   "reqd": 1,
+   "read_only": 1,
+   "set_only_once": 1,
+   "search_index": 1,
+   "in_list_view": 1,
+   "in_standard_filter": 1,
+   "description": "Owner of the paired device. The token authenticates AS this user."
+  },
+  {
+   "fieldname": "device_label",
+   "fieldtype": "Data",
+   "label": "Device",
+   "length": 140,
+   "in_list_view": 1,
+   "description": "Human-readable name the phone reported at pairing (e.g. \"Pixel 8\"). Client-supplied, so display-only - never trusted for any decision."
+  },
+  {
+   "fieldname": "platform",
+   "fieldtype": "Data",
+   "label": "Platform",
+   "length": 40,
+   "in_list_view": 1,
+   "description": "Client-reported platform (ios / android / web / unknown). Display-only."
+  },
+  {
+   "fieldname": "enabled",
+   "fieldtype": "Check",
+   "label": "Enabled",
+   "default": "1",
+   "in_list_view": 1,
+   "in_standard_filter": 1,
+   "description": "0 = revoked. The auth hook refuses a disabled row, so flipping this kills exactly this one device."
+  },
+  {
+   "fieldname": "last_used",
+   "fieldtype": "Datetime",
+   "label": "Last Used",
+   "read_only": 1,
+   "in_list_view": 1,
+   "description": "Stamped by the auth hook at most once per throttle window (see jarvis/mobile/device_auth.py), so it is approximate by design - one write per request would be unacceptable on the hot path."
+  },
+  {
+   "fieldname": "token_id",
+   "fieldtype": "Data",
+   "label": "Token ID",
+   "length": 64,
+   "reqd": 1,
+   "unique": 1,
+   "read_only": 1,
+   "set_only_once": 1,
+   "no_copy": 1,
+   "description": "Public half of the device token (`jmd:<token_id>:<secret>`). Not a secret: it only identifies the row to look up."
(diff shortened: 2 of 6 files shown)


Vigneshsekar:
@l0gesh29 , We usually use the api key and secret. Instead of it, this is just a proposed idea. Accept it only if it works well. Otherwise, ignore and close it.

l0gesh29 (approved):

PR #449, opened 2026-07-26
fix(chat): share one model-pin allowlist between the two write paths

kavin-114:
## What

`send_message`'s inline `model_override` check validated only against `_SUBSCRIPTION_MODELS`, but `set_conversation_model` validates against that **unioned** with the enabled LLM-pool rows. A subscription/pool tenant stores `llm_provider=""`, so the subscription list is empty and `send_message` (the **PWA new-chat** path) rejected every pinned model even though the desktop picker already accepted it.

Fix: extract the union into one `_allowed_pin_models()` helper used by **both** paths, so they cannot drift again.

## Scope (narrowed after review)

An earlier draft also added a direct-mode stale-pin fallback in `turn_handler`. A high code-review found it **silently substitutes a model on the macro path** (macros set `model_override` directly, bypassing the allowlist), so it was **dropped**. A correct version must distinguish "removed from the catalog" from "valid but not pooled", and only matters once a model is actually removed — deferred.

## Tests + CI

`TestAllowedPinModels`: pool-row union, disabled-row exclusion, subscription inclusion. All existing `set_conversation_model` tests unchanged.

Local CI (this repo's real ci.yml in docker) on commit `d90b6bd8`: **PASS — lint 4/4, all four shards 9/9, coverage 5/5**. Report posted as a comment. Green **locally only** (hosted Actions is billing-exhausted; the comment sets no check and runs linux/arm64 vs GitHub's amd64).

### DIFF
--- a/jarvis/chat/api.py
+++ b/jarvis/chat/api.py
@@ -79,6 +79,22 @@ from jarvis._subscription_models import (
 
 _ALLOWED_THINKING = {"", "low", "medium", "high"}
 
+
+def _allowed_pin_models(settings) -> set[str]:
+	"""Every model id a conversation may be pinned to for this tenant: the provider's
+	subscription allowlist unioned with every enabled row of the LLM pool
+	(Jarvis Settings.models). The pool matters because a subscription/pool customer
+	stores llm_provider="", for which _SUBSCRIPTION_MODELS yields [] -- so checking
+	only that rejects every pin. Single source of truth so the two write paths
+	(set_conversation_model and send_message) cannot drift apart again.
+	"""
+	allowed = set(_SUBSCRIPTION_MODELS.get(settings.llm_provider, []))
+	allowed |= {
+		(m.model or "").strip() for m in (settings.models or []) if m.enabled and (m.model or "").strip()
+	}
+	return allowed
+
+
 # Curated dashboard canvas theme keys (lowercase) the builder may forward in the
 # send context — a literal allow-list (not passthrough), mirrors the DocType
 # `theme` Select + frontend/src/lib/dashboardThemes.js.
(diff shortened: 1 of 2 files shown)


kavin-114:
### Code review (high) response

The review flagged that the branch's original **direct-mode stale-pin fallback** in `turn_handler` was unsafe: macros write `conv.model_override` directly (macros.py:345/398) without going through the pin allowlist, so the fallback would **silently substitute** a macro's chosen model. It also introduced an allowlist divergence (the resolver added `llm_model`, the validators did not). Both confirmed.

**Resolution:** dropped bug #2 entirely (`turn_handler` is now byte-identical to develop). This PR is reduced to bug #1 only — the `_allowed_pin_models()` helper shared by `send_message` and `set_conversation_model`, which is behavior-preserving for the desktop path and fixes the PWA rejection. The stale-pin re-check is deferred; it only matters once a model is actually removed from the catalog and needs a design that distinguishes removed-from-catalog from valid-but-not-pooled.

The cleanup note (pool-mode branch hand-rolls its own enabled set) is left as-is: that set is intentionally different (pool rows only, no subscription/llm_model), so it is a distinct check, not drift.

kavin-114:
### Local CI — `Aerele-RnD/jarvis` PR #449

Tested commit `3fd066e39718` · ✅ **PASS** · 48m35s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 48s |
| `tests-shard1` | ✅ PASS | 9/9 | 25m59s |
| `tests-shard2` | ✅ PASS | 9/9 | 8m49s |
| `tests-shard3` | ✅ PASS | 9/9 | 5m39s |
| `tests-shard4` | ✅ PASS | 9/9 | 7m07s |
| `coverage` | ✅ PASS | 5/5 | 9s |

---

kavin-114:
### Local CI — `Aerele-RnD/jarvis` PR #449

Tested commit `d90b6bd8cde4` · ✅ **PASS** · 34m44s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 1m14s |
| `tests-shard1` | ✅ PASS | 9/9 | 10m27s |
| `tests-shard2` | ✅ PASS | 9/9 | 8m57s |
| `tests-shard3` | ✅ PASS | 9/9 | 6m13s |
| `tests-shard4` | ✅ PASS | 9/9 | 7m40s |
| `coverage` | ✅ PASS | 5/5 | 8s |

---

PR #464, opened 2026-07-27
fix(chat): tell the truth about voice clips a send left behind

Vigneshsekar:
Implements the BEHAVIOR SPEC of `fix-phase-20260726/REVIEW-UX-VOICE-CHIPS.md` (findings 1–5), with one directed amendment (below).

## The problem

The owner's report ("chips don't clear after send") is real, but the delete-after-send mechanism is **not** the bug — `captureSentInPayload` → `acknowledge()` releases done clips correctly and stays untouched here. The gap is that a **failed** clip is deliberately never released by a send (its words never made it into the payload) and **nothing told the user that**:

- the chip read exactly like a stuck leftover — nothing distinguished "not sent yet" from "sent, minus this chunk";
- the sent message (and its permanent history entry) read as fluent, complete prose while silently missing a chunk of dictation — no marker, no confirmation;
- Retry's tooltip promised the words would "drop back into the message, where they belong", but post-send the composer was already cleared, so they landed as unexplained text in an empty draft;
- the leave/close warning claimed audio "may be lost" for a clip that had terminally failed and was **durably mirrored** — factually wrong, and it trains users to click through the warning that matters.

## What changed

**`frontend/src/utils/voiceChunkQueue.js`**
- `markSentWithoutClip(seq)` — flags a terminal `failed` clip whose `⟦clip N⟧` placeholder rode out in an accepted send. `snapshot().failed` entries now carry `sentWithout`.
- `hasUnfinishedReason()` → `null | "live" | "unresolved"` — separates a genuine loss risk (pending/inflight, unpersistable, or a transcript that exists only in a volatile un-sent draft) from a merely-unresolved, confirmed-durable clip. `hasUnfinished()` is kept as the coarse predicate on top of it, unchanged in meaning.

**`frontend/src/views/ChatView.vue`**
- Send-time gap confirmation, **before** the strip and the POST: *"Send without part of what you said?"* / "Send anyway" / "Review first". Cancel aborts the send with the placeholder and chip still in place. `fromMain` only (a `resendFailed` re-sends a payload the user already chose to send with the gap).
- Flags the carried gaps via `markSentWithoutClip` in the accepted-send branch, beside the existing release, so both describe the same payload.
- Chip label and Retry tooltip branch on `sentWithout` ("Clip N was missing from your last message" / "Transcribe and add as a follow-up message"). Pre-send copy unchanged.
- `_replaceGapPlaceholder` toasts when it falls back to appending (token gone — the message already left), so recovered text never appears unexplained.
- The leave guard arms on `_voiceGuardReason() === "live"` for both `beforeunload` and `onBeforeRouteLeave`. Live copy is verbatim unchanged.

**`design.md` §4.5** — the two voice-capture surfaces (chat composer chunked-with-chips vs. single-shot `useAudioRecorder`) and their different retention models, plus the copy/tooltip/leave-warning rules, so a third contract isn't invented by accident (finding 5, P2, doc-only as specified).

## Amendment applied (directed)

Spec item 5 offered new non-loss-framed copy for the `"unresolved"` case. Per the direction given, `"unresolved"` does **not** block navigation at all — no confirm, no `beforeunload` arming; the chips and the recovery banner already carry it. The hard guard with the current copy stays only for `"live"`.

## Out of scope, untouched (and fenced by a test)

`captureSent` / `captureSentInPayload` / `acknowledge` for done clips, the recovery-banner flow, and the retained ("was edited out") chip copy.

## Tests

- `voiceChunkQueue.test.js`: 33 → 38 (5 new walks: sent-without flagging + snapshot shape + the fenced done-clip release; the reason split; unpersistable-is-always-live; a retried sent-without clip committing as a follow-up; a failed clip whose mirror write hasn't confirmed yet is live). **38/38 pass.**
- `voiceChipLifecycle.test.js` (new, the `pumpFence.test.js` source-assertion precedent — ChatView cannot be mounted in this app): confirm-before-strip-and-POST ordering, flagging inside the accepted-send branch, chip copy branching, the retry toast, guard-on-reason, plus a decision-parity walk against the real queue and a fence on the release path. **8/8 pass.** Run in CI by `jarvis/tests/test_voice_chip_lifecycle_client.py`.
- Unregressed: `voiceSendGlue.test.js` 11/11, `clipMirror.test.js` 7/7, `eventFence.test.js` 8/8.
- `npx vite build` green; `pre-commit run --files …` (pinned prettier, eslint, ruff) green.

Bench modules to run: `jarvis.tests.test_voice_chunk_queue_client`, `jarvis.tests.test_voice_chip_lifecycle_client`, `jarvis.tests.test_voice_send_glue_client` (they subprocess the node suites above).

### DIFF
--- a/design.md
+++ b/design.md
@@ -547,6 +547,31 @@ three times. Recipe:
 - A `FormControl` paste field (never a bare `<textarea>`), inline `ErrorMessage` for
   expired/invalid-nonce errors, countdown hint in `text-p-sm text-ink-gray-5`.
 
+### 4.5 Voice capture — two surfaces, two retention models (say which one you are)
+
+Jarvis has **two** voice-recording surfaces with deliberately different contracts, and a user who
+learns one carries its expectations to the other. State which one a new surface implements.
+
+- **Chat composer dictation** (`useChunkedRecorder` + `utils/voiceChunkQueue.js`, ChatView): long
+  takes chunked into self-contained clips, transcribed independently, **audio retained** in an
+  IndexedDB mirror until its text is part of an accepted send. Anything unresolved is **visible and
+  actionable as a chip** — a failed clip (Retry/Download/✕), a clip edited out before sending
+  (Restore/Download/✕), a clip whose audio could not be saved (Download/Retry) — plus a
+  previous-session recovery banner. Nothing is ever removed silently except the success path
+  (chips clear on the send that carried their words; no toast — that would be noise on the common
+  case).
+- **Single-shot capture** (`useAudioRecorder`, `VoiceRecorder.vue` — wiki nudge, Business tab):
+  record → transcribe → text, **no retention, no chips, no recovery**. A failure is a toast and a
+  re-record, and the audio is gone.
+
+Rules for either: chip copy must distinguish "not sent yet" from "already sent without this"
+(identical copy for both is what makes a correctly-retained clip read as a stuck one); an action's
+tooltip must promise only what it can do *now* (Retry cannot edit a message that has already been
+sent — it builds a follow-up); and a leave/close warning may claim loss **only** when something is
+genuinely at risk — durably mirrored, re-offerable clips must not arm it (a warning that cries
+wolf gets trained away). If a new surface wants the chip model, reuse the queue rather than growing
+a third contract.
+
 ---
 
 ## 5. Do / Don't — current Jarvis anti-patterns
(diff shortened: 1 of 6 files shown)


Vigneshsekar:
Superseded by #467 (at-once dictation). Its substance survives there — the retention/acknowledge lifecycle, the sentWithout honest-copy state, and the hasUnfinishedReason leave-guard split were all carried over; the per-clip gap-token machinery this PR polished is deleted wholesale by the redesign. Closing to keep the merge queue unambiguous — the voice line is admin-v2#123 → jarvis#463 → jarvis#467.

PR #465, opened 2026-07-27
fix(chat): client-side near-silence gate for dictation (stacks on #464)

Vigneshsekar:
> **Stacks on #464** (`fix/voice-chip-lifecycle`) — this branch is based on it because #464 owns the latest `voiceChunkQueue.js` / `ChatView.vue` voice glue. **Merge #464 first**; this PR's diff against `develop` will then be only the silence gate.

## Why

Desktop chunked dictation is moving to `whisper-large-v3-turbo` via OpenRouter's transcription endpoint (#463). That model **deterministically hallucinates on near-silent audio** — a pure-silence 15 s webm transcribes as `"Thank you."` 3/3 on today's probe.

The chunked recorder rotates a clip every ~15 s *whether or not anybody spoke*, and pauses mid-dictation are routine. Without a gate, every pause injects a confident little phrase the user never said into the composer.

It cannot be filtered server-side:
- OpenRouter does **not** pass through `verbose_json` / `no_speech_prob` (probed: it 400s), so there is no confidence signal to threshold on.
- A server-side artifact blocklist was vetoed — it would eat the same phrases when a user *genuinely* dictates them.

So the gate is client-side, at the microphone: measured near-silence never leaves the browser.

## What

- **`frontend/src/utils/voiceSilenceGate.js` (new)** — `createRmsMeter(stream, deps)` runs a WebAudio `AnalyserNode` on a MediaStream, sampling peak RMS at 10 Hz (4096-sample window ≈ 85 ms, so ~85% of the clip's timeline is observed and a short word can't slip between two windows). `isNearSilent(peakRms)` is the decision: strictly `<` a conservative **0.005** peak-RMS threshold (~-46 dBFS). The peak is taken across the **whole clip**, so one word anywhere in it clears the gate — skipping requires the entire 15 s to be inaudible. WebAudio and the timers are dependency-injected so `node --test` drives the real graph code without a browser.
- **`useChunkedRecorder.js`** — attaches the meter to the **same** `MediaStream` the `MediaRecorder` is recording, resets the peak at each cycle boundary (per-clip peak) and stamps each rotated clip with its measured `peakRms` (alongside the existing `durationS`). The meter is torn down with the stream.
- **`ChatView.vue`** — the clip handoff gates before the queue: a clip measured as near-silence is resolved **without an API call**. It contributes no text, leaves no `⟦clip⟧` placeholder, creates no chip, is not a failure and does not arm the leave guard — it simply never happened. A `console.debug` breadcrumb notes the skip (no new telemetry).

## The invariant that outranks the feature

**Absence of measurement must never drop audio.** Every failure mode reports UNMEASURED, and an unmeasured clip is uploaded exactly as it is today:

- no `AudioContext` / `webkitAudioContext` at all
- a constructor or graph call that throws (the half-built context is closed, not leaked)
- an analyser read that throws
- a clip that ended before the first sample
- **a context the browser suspended** — its analyser reads all-zero, which is indistinguishable from silence; `peak()` re-checks `ctx.state` so a parked context can never be mistaken for a pause

The analyser chain also terminates in a **muted** (`gain = 0`) node connected to the destination, so a browser that only renders a *pulled* graph cannot feed the analyser zeros and gate an entire session. Gain 0 means nothing is audible and there is no mic→speaker feedback path.

Every tie goes to keeping the audio: the comparison is strictly `<`, so a clip sitting exactly on the threshold is transcribed.

## Tests

25 new walks in `frontend/src/utils/voiceSilenceGate.test.js` (`node --test`, the `voiceChunkQueue.test.js` precedent), run in CI by the new `jarvis/tests/test_voice_silence_gate_client.py`:

- the decision table and the threshold boundary (at-threshold transcribes, one ULP under skips)
- every unmeasured/degrade path above, including the suspended-context trap
- the per-cycle peak reset (a loud clip must not shield the next clip's silence)
- **end to end against the real `voiceChunkQueue`**: a measured-silent clip makes no transcribe call, produces no chip, no placeholder, no failure and no armed guard; a loud clip runs the normal lifecycle; an unmeasured clip is always transcribed; a dropped pause doesn't disturb the seq ordering of the clips around it
- source assertions fencing the ChatView/recorder wiring the harness stands in for (gate strictly before `enqueue`, meter on the same stream, peak read before the next cycle resets it, meter torn down with the stream)

`voiceChipLifecycle.test.js` gains the chip consequence: a gated pause leaves the chip row exactly as it found it.

Run locally, all green: `voiceSilenceGate` + `voiceChipLifecycle` + `voiceChunkQueue` + `voiceSendGlue` = **83/83 pass**; `vite build` ✓ (1m 26s); pinned `pre-commit run prettier` + `ruff` + `eslint` pass.

## Risk notes

- The threshold is a judgement call. It errs low on purpose: too low merely restores today's behaviour (transcribe, risk a hallucination); too high deletes speech. Worth one real-mic pass confirming a normal pause measures under 0.005 and quiet/distant speech measures over it.
- The muted destination sink is the deliberate belt against a non-pulled graph; it keeps an output path open while recording.

### DIFF
--- a/design.md
+++ b/design.md
@@ -547,6 +547,31 @@ three times. Recipe:
 - A `FormControl` paste field (never a bare `<textarea>`), inline `ErrorMessage` for
   expired/invalid-nonce errors, countdown hint in `text-p-sm text-ink-gray-5`.
 
+### 4.5 Voice capture — two surfaces, two retention models (say which one you are)
+
+Jarvis has **two** voice-recording surfaces with deliberately different contracts, and a user who
+learns one carries its expectations to the other. State which one a new surface implements.
+
+- **Chat composer dictation** (`useChunkedRecorder` + `utils/voiceChunkQueue.js`, ChatView): long
+  takes chunked into self-contained clips, transcribed independently, **audio retained** in an
+  IndexedDB mirror until its text is part of an accepted send. Anything unresolved is **visible and
+  actionable as a chip** — a failed clip (Retry/Download/✕), a clip edited out before sending
+  (Restore/Download/✕), a clip whose audio could not be saved (Download/Retry) — plus a
+  previous-session recovery banner. Nothing is ever removed silently except the success path
+  (chips clear on the send that carried their words; no toast — that would be noise on the common
+  case).
+- **Single-shot capture** (`useAudioRecorder`, `VoiceRecorder.vue` — wiki nudge, Business tab):
+  record → transcribe → text, **no retention, no chips, no recovery**. A failure is a toast and a
+  re-record, and the audio is gone.
+
+Rules for either: chip copy must distinguish "not sent yet" from "already sent without this"
+(identical copy for both is what makes a correctly-retained clip read as a stuck one); an action's
+tooltip must promise only what it can do *now* (Retry cannot edit a message that has already been
+sent — it builds a follow-up); and a leave/close warning may claim loss **only** when something is
+genuinely at risk — durably mirrored, re-offerable clips must not arm it (a warning that cries
+wolf gets trained away). If a new surface wants the chip model, reuse the queue rather than growing
+a third contract.
+
 ---
 
 ## 5. Do / Don't — current Jarvis anti-patterns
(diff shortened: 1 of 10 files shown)


Vigneshsekar:
Superseded by #467 (at-once dictation). The silence-gate meter, fail-open rules, and its test suite were carried into that branch verbatim (voiceSilenceGate.js, re-pointed at the whole take). Closing to keep the merge queue unambiguous — #467 is the one to merge, stacked on #463.

PR #481, opened 2026-07-27
fix(dashboards): clear the composer, ask before building, keep the canvas

Vigneshsekar:
The three owner-reported Dashboards builder defects (2026-07-27). Persona half: **Aerele-RnD/jarvis-persona#170** (`fix/dashboards-clarify-first`) — deploy together; the skill and the injected context state the same rule.

---

## D1 — the composer kept the message it had just sent

`send()` **did** clear the draft. The clear did not survive. The pane bound `:disabled="sending"` on a frappe-ui `FormControl` textarea, and that component patches `disabled` before `value` and listens on `change` as well as `input`. So on the Enter path the browser blurred the focused, dirty box, `change` fired, and the **pre-clear DOM value** was re-emitted straight back into `draft`.

Fixed by adopting `Composer.vue`'s documented pattern rather than patching around it:

- a raw `<textarea v-model="draft">` — `v-model`, not `:value` + a manual emit, so Vue's own `vModelText` keeps handling IME composition;
- the textarea is **never** disabled; `sending` gates only the send button and `send()`;
- `autoGrow` is driven by a `flush:"post"` watcher on `draft`, so a programmatic clear actually shrinks the box (four scattered `nextTick(autoGrow)` calls go away).

`TriggerChatPane.vue` is the fork this pane came from and carried byte-identical wiring — fixed too. Its Send button also now gates on `sending` (it only had `:loading`, so a double-submit was reachable).

Restore-on-server-reject is unchanged, on both the `ok:false` and the throw path.

## D2 — it built off the first message instead of asking

**Server (`turn_handler._prepend_doc_context`).** The client sends the same `{"page":"dashboards"}` context on every turn and has no notion of a turn index, so first-build state is derived **server-side** — from the same `Jarvis Chat Message.canvas` field `persist_canvases` stamps and `get_conversation` already hands the UI. No canvas in this conversation, and not revising a saved document ⇒ this turn asks. The context then instructs exactly ONE `jarvis-ask` block covering the 2-4 decisions that shape a build: data scope, the breakdown, static-vs-live **only when the Data toggle has not settled it**, theme **only when none is selected**. Explicit escape hatches keep it from becoming an interrogation: skip anything the request already answers, build immediately when it answers everything, never ask twice. Once a canvas exists the wording is byte-for-byte today's iterate text.

**No new context key** — `send_message`'s allow-list is untouched, so there is no new API surface and nothing a stale frontend can smuggle. The helper fails to `True` (build, i.e. today's behaviour) on any DB error, so an unreadable state can never re-interrogate the user every turn.

**Frontend.** The pane *deleted* ` ```jarvis-ask ` blocks, so a clarifying turn rendered as an empty bubble — the mechanism existed but was unreachable here. ChatView's option cards are now a shared **`<AskCard>`** over a shared **`@/lib/chatAsk`** (parse / readiness / answer formatting), used by both surfaces. Answers post into the same conversation as an ordinary user message.

**Extraction note (flagged for judgment):** the brief left "shared component vs. importing ChatView internals" to the doer. I extracted the component — ChatView keeps no copy, so there is nothing to drift. It cost a real ChatView diff (−479 lines: the template block, ~110 lines of draft/selection logic, and ~195 lines of `jv-ask*` CSS, all moved verbatim into `AskCard.vue`). The cards look identical because the CSS moved with them and the `jv-*` tokens resolve by cascade; the dashboards pane binds `paletteVars` on the card (the `SkillDetail.vue` precedent) since those vars are not global. Both hosts key the card by message name — the draft lives in the component, and `spec` is a computed that re-parses on every stream tick, so a `spec` watcher would have wiped half-made picks.

## D3 — the canvas vanished, silently

- **Tab switch.** Visiting Saved unmounted `DashboardChatPane`, whose `onBeforeUnmount` tears down the socket listener — and the agent's `kind:"canvas"` frame is a **one-shot** publish, never replayed. A build that landed while the user was on Saved was lost outright. The builder tab is `v-show` now (Saved becomes its own `v-if`), so the pipeline survives.
- **Navigation.** The route component unmounts and `builderHtml` is gone. Nothing is actually lost: `get_conversation` returns each message's parsed `canvas` list, and `get_canvas` replays any past message for the conversation's owner. The pane now emits the newest canvas frame from every transcript load (`lib/dashboardRestore.js`), so returning to the page is lossless — with **zero new server state**. A live frame always beats a replay, re-checked after the `get_canvas` round trip; a failed replay is silent (it is background work nobody asked for). Side benefit: the `?nosocket` / headless-QA path now gets a canvas it could never receive before.
- **Explicit discard.** "New dashboard", the pane's "New chat", and opening another dashboard for editing all ask first — *"Discard this unsaved dashboard? Its chat stays in your conversations."* — and only when there really is unsaved work (canvas html that is no longer the saved document's). "New chat" now **asks the page** instead of clearing itself first, so cancelling no longer leaves an emptied chat beside a surviving canvas.
- **Found while proving the resume loop:** the pane never followed the sticky conversation slot when the page repointed it (`loadEdit` resuming a saved dashboard's thread), so the transcript on screen could belong to a different dashboard than the one the next send went to. Fixed with a guarded watcher (`"" -> id` is the pane's own first send and is left alone).

### Flagged for Fable's ruling

`loadEdit()` gets the discard confirm as the brief specifies, but today it only ever runs at mount, when the canvas is empty — so it is a no-op guard rather than a live prompt. The `?edit=` deep-link is ordered so the explicit target wins over a transcript replay (`editSeed`, read synchronously at setup, before the pane mounts). The in-progress build's *chat* is not destroyed by that path — only the single sticky slot's pointer is repointed, and the conversation remains in the user's chats.

---

## Gates run

| Gate | Result |
|---|---|
| `vite build` | ✓ built (clean, after every edit) |
| `vitest run` (5 files) | **51 passed** — incl. new `ask-card.test.js` (8, mounts and drives the cards) |
| `node --test` new suites | `chatAsk.test.js` **18/18**, `dashboardRestore.test.js` **17/17** |
| `node --test` pre-existing suites | 11 files, **162 passed, 0 failed** |
| bench `test_dashboard_clarify_context` (new) | **11 OK** |
| bench `test_chat_ask_client` / `test_dashboard_builder_ux_client` (new) | **1 OK each** |
| bench `test_dashboards_api` | **39 OK** |
| bench `test_canvas_embeds` / `test_canvas` | **13 / 9 OK** |
| bench `test_turn_handler_report_context` | **14 OK** |
| bench `test_untrusted_fence` | **21 OK** |
| bench `test_turn_handler_thinking` | **5 OK** |
| bench `test_chat_asks` | **32 OK** |
| bench `test_agent_dashboard_wire` | **7 OK** |
| bench `test_dashboard_tools` / `test_dashboard_permissions` | **8 / 17 OK** |
| ruff check + ruff format | clean (`turn_handler.py` needed no reformat — no whole-file churn) |
| pinned prettier + eslint (pre-commit) | clean |

Bench modules ran on `patterntest.localhost` with the branch on `PYTHONPATH` (verified `frappe.get_app_path("jarvis")` resolved to the worktree). Canonical checkouts untouched.

Not merged.

### DIFF
--- a/frontend/src/components/chat/AskCard.vue
+++ b/frontend/src/components/chat/AskCard.vue
@@ -197,10 +197,12 @@ function isPicked(i, opt) {
 }
 function submit() {
 	if (!ready.value) return;
-	const text = askAnswerText(props.spec, sel.value, other.value);
-	sel.value = {};
-	other.value = {};
-	emit("submit", text);
+	// Emit and keep the picks. The card is keyed by message, so it unmounts the
+	// moment the next assistant message arrives — there is nothing to reset. And
+	// a host is allowed to REFUSE the text (a send already in flight, a dictation
+	// still transcribing): clearing first would silently swallow the answers and
+	// leave the user staring at an empty, disabled card.
+	emit("submit", askAnswerText(props.spec, sel.value, other.value));
 }
 </script>
 
--- a/frontend/src/components/chat/AskCard.vue
+++ b/frontend/src/components/chat/AskCard.vue
@@ -1,0 +1,402 @@
+<template>
+	<div class="jv-ask" :class="{ 'jv-ask--form': isForm }">
+		<div v-for="(q, qi) in spec.questions" :key="qi" class="jv-ask-q">
+			<div class="jv-ask-qt">
+				<span class="jv-ask-num">{{ qi + 1 }}</span
+				>{{ q.q }}
+			</div>
+			<!-- yes/no, with optional custom labels (e.g. Approve / Reject) -->
+			<div v-if="q.type === 'yesno'" class="jv-ask-opts">
+				<button
+					v-for="(lbl, li) in q.options.length === 2 ? q.options : ['Yes', 'No']"
+					:key="li"
+					class="jv-ask-opt"
+					:class="{ on: isPicked(qi, lbl) }"
+					@click="toggleSingle(qi, lbl)"
+				>
+					<span v-if="isPicked(qi, lbl)" class="jv-ask-tick">✓</span>{{ lbl }}
+				</button>
+			</div>
+			<!-- single / multi choice -->
+			<div v-else-if="q.type === 'single' || q.type === 'multi'" class="jv-ask-opts">
+				<button
+					v-for="(opt, oi) in q.options"
+					:key="oi"
+					class="jv-ask-opt"
+					:class="{ on: isPicked(qi, opt) }"
+					@click="q.type === 'multi' ? toggleMulti(qi, opt) : toggleSingle(qi, opt)"
+				>
+					<span v-if="isPicked(qi, opt)" class="jv-ask-tick">✓</span>{{ opt }}
+				</button>
+			</div>
+			<!-- date / datetime / free text fields -->
+			<input
+				v-else-if="q.type === 'date'"
+				type="date"
+				class="jv-ask-field"
+				:value="sel[qi] || ''"
+				@input="pickSingle(qi, $event.target.value)"
+			/>
+			<input
+				v-else-if="q.type === 'datetime'"
+				type="datetime-local"
+				class="jv-ask-field"
+				:value="sel[qi] || ''"
+				@input="pickSingle(qi, $event.target.value)"
+			/>
+			<input
+				v-else-if="q.type === 'text'"
+				type="text"
+				class="jv-ask-field"
+				:value="sel[qi] || ''"
+				@input="pickSingle(qi, $event.target.value)"
+				placeholder="Type your answer…"
+				@keydown.enter.prevent
+			/>
+			<!-- link: search a record of the given DocType -->
+			<div v-else-if="q.type === 'link'" class="jv-ask-link">
+				<input
+					type="text"
+					class="jv-ask-field"
+					:value="link[qi] && link[qi].q != null ? link[qi].q : sel[qi] || ''"
+					@input="onLinkSearch(qi, q.doctype, $event.target.value)"
+					@focus="onLinkSearch(qi, q.doctype, (link[qi] && link[qi].q) || '')"
+					:placeholder="'Search ' + (q.doctype || 'records') + '…'"
+					@blur="closeLink(qi)"
+					@keydown.enter.prevent
+				/>
+				<div
+					v-if="link[qi] && link[qi].open && (link[qi].items || []).length"
+					class="jv-ask-linkmenu"
+				>
+					<button
+						v-for="(it, ii) in link[qi].items"
+						:key="ii"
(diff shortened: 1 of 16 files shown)


Vigneshsekar:
**Review remediation pushed (`59619343`).** Every P1 and P2 from the UX and adversarial panels is addressed on this branch.

P1
- **Interview re-fired every turn until a canvas persisted.** A second probe (`_dashboard_has_asked`) looks for an assistant ```jarvis-ask block; once the ask has gone out the context tells the model to use the answers and build — wording that no longer contains "do NOT build". Closes the failed-publish trap and the self-hosted (`canvas_available:false`) infinite loop. The first-turn arm is scoped to build requests, so a greeting or a data question gets a normal answer.
- **Restore lost the editing identity** (Save inserted a duplicate `Jarvis Dashboard`). The editing target is sticky per user; a remount resumes it through `loadEdit`.
- **Blank canvas on tab return.** `DashboardCanvas` exposes `rebuild()`; the page re-drives it (`flush:"post"`) when the builder tab comes back, so an iframe that loaded at 0×0 behind `display:none` is not left with dead charts.

P2
`_dashboard_has_canvas` now filters to html items (a generated image no longer suppresses the interview) · the restore replays the recorded builder canvas message, not "newest html in the conversation" · `?edit=` is a watcher, so its discard confirm is reachable and the seed clears when the confirm settles · `AskCard` keeps its picks until the host takes the text · the theme clarify branch is deleted (the picker always supplies one) · both probes `log_error` before failing open · `TriggerChatPane` renders `AskCard` · failed restores are latched per message id · the Auto data-mode copy states the new meaning · the discard confirm offers "Open its chat" · the three Enter-to-send handlers ignore an IME composition commit · a repoint from the pane's own in-flight send keeps its Thinking indicator · the clarify-context test no longer commits and deletes its messages with the conversation (the orphan rows earlier runs left on patterntest are cleaned).

Gates re-run on the branch: 357/357 node (`src/**/*.test.js`), 52/52 vitest, `vite build` clean, pinned prettier/ruff/eslint via pre-commit, and on patterntest.localhost `test_dashboard_clarify_context` (20), `test_dashboards_api` (39), `test_canvas_embeds` (13), `test_dashboard_builder_ux_client`, `test_chat_ask_client`, `test_turn_handler` (7), `test_turn_handler_report_context` (14), `test_turn_handler_thinking` (5) — all OK, and the suite now leaves zero rows behind. Five deliberate mutations (watcher removed, `flush:"pre"`, sticky target not cleared, AskCard clearing again, IME guard removed) each fail the fence suite, so the new tests assert effect rather than presence.

Pairs with jarvis-persona#170.

PR #274, opened 2026-07-10
Frappe design language: design.md + Settings (Account & billing) and onboarding UI restyle

Vigneshsekar:
## What this PR does

1. **Adds `design.md` at the repo root** — the Frappe design language distilled from the actual code of **frappe-ui (0.1.278)**, **Frappe CRM**, **Frappe Helpdesk** and **Frappe Cloud (press)**: semantic tokens, type/space/radius/shadow scales, button/dialog/form/badge/banner recipes, the settings-dialog and billing/plan/onboarding patterns, a Do/Don't table of the anti-patterns that were live in Jarvis, and the `jv-*` legacy-palette mapping used for this port.
2. **Restyles the shell Settings dialog** (with the **Account & billing** group as the centerpiece) and the **onboarding wizard + intro tour** to that language — **appearance only, functionality frozen**.
3. **Ships the rendered evidence**: `docs/ui-preview/index.html` is a before/after gallery (52 screenshots, light + dark) captured from the real running app — open it locally, or skim the key pairs below **before merging**.

## Key before / after

| | Before (main) | After (this PR) |
|---|---|---|
| **Plan & billing** | before | after |
| **Plan & billing (dark)** | before | after |
| **General pane** | before | after |
| **Onboarding — plan** | before | after |
| **Onboarding — pay (dark)** | before | after |
| **Onboarding — details** | before | after |

Full gallery (all panes/steps, both themes): [`docs/ui-preview/index.html`](https://github.com/Aerele-RnD/jarvis/blob/ui/frappe-design-language/docs/ui-preview/index.html) *(check out the branch and open in a browser — images are committed alongside)*.

## What changed, concretely

**Settings dialog** — pane-owned header (title + one-line description; duplicate in-pane headings removed), raised-card active nav item, sentence-case rail groups (the boxed "System Mgr" tag is gone — the rail and server already gate the group), flat near-black primary buttons with zero hover-lift, one retry pattern, near-black switches, raised-chip segmented controls, quiet ghost close button.

**Plan & billing pane** (the flawed one) — status badge inline with the plan name, feature list with check glyphs, upgrade cards restructured (name/price header + full-width subtle **Upgrade** button — no more `Upgrade →` link-buttons), *Manage plan & billing* is now a plain text link with an external-link icon.

[...]

PR #497, opened 2026-07-27
refactor(onboarding): migrate OnboardingView to frappe-ui (part B)

kavin-114:
## Summary

Part B of jarvis-admin-v2#65 stage 2 ("Consistent Frappe UI with onboarding and
settings page") - migrates `frontend/src/views/OnboardingView.vue` (2639 lines,
66 legacy classes measured with `grep -oE 'class="[^"]*"' OnboardingView.vue |
grep -oE '\bjv-[a-z0-9-]+' | sort -u`, and only that command) to frappe-ui.
Part A (`OnboardingGate`, `Banner`, `SetupNeuralNet`) is #487, already merged.

This is the paid signup funnel - the highest-stakes surface in the app.
**Presentation only. No behavior, copy, or flow-control change.** The script
section of the file is untouched except for one new import line (verified by
diffing `<script setup>...</script>` against `develop` after every commit).

66 legacy classes -> 2 (`jv-ob-root`, and the `jv-dark` class binding), both
deliberately kept and documented in the template - see "Traps" below.

## Commit sequence (each builds + lints clean on its own)

1. **Shared page chrome** - root/main/center/wrap/brand header/step rail/panel
   shell to Tailwind. Every step's own content is untouched in this commit
   (byte-identical to `develop`); the six sections still reference their
   pre-existing `jv-ob-*` classes, whose CSS still ships unchanged here.
2. **Plan step** - cards, checkmark icon (-> FeatherIcon), Retry/Continue
   buttons (-> frappe-ui `<Button>`). Introduces the shared step-chrome
   classes (`ob-screen`/`ob-body`/`ob-head`/`ob-foot`/`ob-back`/`ob-link`/
   `ob-note`) every later step adopts - copied 1:1 from the still-present
   `jv-ob-*` originals (same padding, same `var(--...)` colors), so this is a
   pure rename for those properties, not a restyle.
3. **Details step** - every `<input>` -> frappe-ui `<FormControl
   variant="outline">`. JvCombo (out of scope, not a plain input) gets a new
   `.ob-details-form :deep(.jvc-field)` rule matched to FormControl's own
   `variant="outline"` recipe so it still looks like its siblings.
4. **Review & Pay step** - spinner -> `<LoadingIndicator>`; review card
   restyled per design.md's read-only-row pattern; the payment-gateway
   segmented chooser's hardcoded `#eef0f3`/`#e3e6ea` grays (the one hex-color
   violation in this file, predating this PR) -> `bg-surface-gray-2`/
   `border-outline-gray-1`; lock icon -> FeatherIcon; every button -> real
   `<Button>` with `:loading`/`loading-text` instead of manual text-swaps.
5. **Connect + Self-host steps** - FormControl/Checkbox/FeatherIcon
   conversions; retires the last cross-cutting legacy CSS (the global
   `jv-ob-root` focus-visible rule, `jv-ob-btn-primary`, dead `jv-ob-err*`).
   **Does not touch the "Start chatting" button** - see next commit.
6. **The Start Chatting CTA, alone** - see below.

## The one deliberate accent: commit 6

The task brief for this migration (based on design.md's anti-pattern table and
part A's precedent) expected a gradient/pulsing-glow CTA and decorative
background orbs to strip. I checked before touching anything:
`grep -iE 'orb|glow|gradient|pulse|blur' OnboardingView.vue` and a `git log -S`
trace both come up empty for this file **today**. That work already happened -
PR #294 (2026-07-14, well before this migration) flattened the "Start
chatting" button's CSS to a plain solid near-black fill and removed the
pulsing-glow keyframe; only the class name `jv-ob-btn-grad` survived as a
leftover label on since-flattened CSS (its own code comment already said so).
There are no decorative orbs anywhere in this file's history since #274/#294's
"Frappe design language" restyle. **So there was nothing to strip.**

What commit 6 *does* do, kept isolated as instructed so it can be reverted
alone: migrates that last hand-rolled button (`jv-ob-btn jv-ob-btn-grad`) to a
real frappe-ui `<Button variant="solid">`, matching every other footer button
in this file. The visible delta is small but real - the button's metrics move
from the old 36px/16px-padding/8px-radius hand-rolled recipe to frappe-ui
`Button`'s default `sm` size (28px/`h-7`/`rounded`). If Kavin wants the old
36px sizing back, revert this commit alone; every other commit stands
independently.

## Traps checked (from the task brief)

1. **`SetupNeuralNet`'s `paletteVars` dependency.** `readColors()`
   (`onboarding/SetupNeuralNet.vue`) reads `--text-3`/`--surface`/`--surface-3`
   via `getComputedStyle` on its own root, which only resolve because they
   inherit from `paletteVars` bound on an ancestor - there is no `:root`
   fallback. **The binding survives**: `<div class="jv-ob-root" :class="{
   'jv-dark': dark }" :style="paletteVars">` is untouched on
   `OnboardingView`'s root element throughout all 6 commits. Verified by:
   (a) grepping the final file - the binding is still there, unmoved; (b) the
   production build succeeds and the dev server transforms the module
   cleanly; (c) `SetupNeuralNet.vue` itself was not touched (`git diff
   --stat` shows only `OnboardingView.vue` across all 6 commits). I could not
   visually confirm the canvas animation renders with resolved colors (no
   browser tool, and reaching the connect step needs a not-yet-onboarded
   workspace - see below); the coordinating session's full-flow capture is
   the visual check for this.
2. **A second, related trap I found that wasn't in the brief**:
   `frontend/src/index.css` has a *global* rule keyed to the literal string
   `.jv-ob-root` (`.jv-ob-root :where(input...):focus { outline: revert; ...
   }`, added to suppress `@tailwindcss/forms`' blue focus ring specifically on
   this page). `LlmPoolEditor.vue` (Connect step) and `JvCombo.vue` (Details
   step) - both out of scope for this migration - render raw `<input>`
   elements that depend on this rule, and also read the same legacy
   `var(--...)` custom properties directly in their own scoped CSS (182 and
   12 usages respectively), which only resolve inside a `paletteVars`-bound
   subtree. This is a second, independent reason `jv-ob-root` couldn't be
   renamed even if I'd wanted to shed it entirely - it's why 2 of the 66
   original classes remain, both documented in a template comment on the root
   element explaining exactly why.
3. **Copy**: frozen, verified. Every heading/paragraph/button-label/error
   string was checked present verbatim post-migration (script diff is one
   import line; template text-content spot-checked exhaustively against
   `develop` for every step). The one wording-adjacent change: "(optional)"
   suffixes on Details-step labels move from a separately dimmed `<span>`
   into the `FormControl` `label` string itself (e.g. `"Contact number
   (optional)"`) - same words, same order, no rewrite; `FormControl`'s label
   prop is a single string with no sub-styling hook. Passing `required` on
   the work-email field is a new (positive) side effect of using the real
   component - `FormLabel` renders the standard red-asterisk + sr-only
   "(required)" design.md mandates; the field was already functionally
   required (unchanged JS validation in `onDetailsSubmit`), this only makes
   it visible.
4. **Behavior**: frozen. Flow control, step order, poll budgets, and error
   handling are byte-identical to `develop` - the only script change across
   all 6 commits is the one new frappe-ui import line.
5. **Banner call sites**: all 9 untouched, confirmed - `Banner.vue`'s
   type/title/message props and slots are unchanged since #487, so these
   inherit the info-color fix for free as expected.

## What I could and couldn't verify

**Could:**
- `npm run build` succeeds on every one of the 6 commits, and on the final
  state: **5180 modules transformed**, matching #487's own count (no new
  dependencies - `Button`/`FormControl`/`FeatherIcon`/`LoadingIndicator` are
  already bundled).
- `pre-commit run --files frontend/src/views/OnboardingView.vue` passes
  (prettier 2.7.1) on every commit; eslint's pre-commit hook is scoped to
  `types_or: [javascript]` in this repo's own config and reports "no files to
  check" for a `.vue` file - that's this repo's existing, correct behavior,
  not something this PR bypasses.
- Every one of the 131 Tailwind utility classes this migration introduces was
  cross-checked against the actual compiled CSS output
  (`jarvis/public/frontend/assets/*.css`) to confirm each one really
  generated a rule - this caught and fixed a real bug during development
  (several `max-w-<N>`/`min-h-<N>` bare-number classes I'd used don't exist in
  this Tailwind preset's scale and were silently dropped; fixed to bracket
  arbitrary-value syntax, e.g. `max-w-[620px]`, re-verified against the
  compiled CSS afterward).
- The dev server (`npm run dev`, port 8082) transforms
  `/src/views/OnboardingView.vue` (script, template, and scoped-style
  sub-requests) with HTTP 200 for both `site.jarvis` and `jarvis.proxy` Host
  headers - the module graph compiles cleanly end to end. No browser tool was
  used at any point (curl only, per instructions).

**Could not:**
- Visually confirm any step actually rendering, in either theme. Per the task
  brief, the wizard only renders for a workspace that has NOT onboarded, and
  none exists on this bench right now (`jarvis.proxy` is onboarded;
  `site.jarvis`'s `is_ready_for_chat` reason is `llm_credentials`, which is
  deliberately excluded from the not-onboarded set). I did not mutate any
  site's state to force it, per instructions. The coordinating session
  captures the full funnel in both themes for review before this merges -
  this PR should be treated as "compiles clean, traps checked, copy/behavior
  frozen" evidence, not "seen rendered."

## Deliberately left alone

- `JvCombo.vue` and `LlmPoolEditor.vue` - out of scope, not touched (`git diff
  --stat` across all 6 commits shows only `OnboardingView.vue`).
- `jv-ob-root` and the `jv-dark` binding - kept, documented, load-bearing (see
  Traps above).
- The Razorpay wordmark's own navy (`#0b2a6b`) in the payment-gateway
  chooser - kept as a hardcoded hex on purpose, matching the brand-asset
  exception (it's a partner's own mark, not app chrome).
- No bugs found in the underlying logic were fixed - this PR is presentation
  only, per instructions.

## CI

**No local CI evidence on this PR.** Hosted GitHub Actions is the real gate
again (billing was restored 2026-07-24); the `jarvis-ci` local harness is
known-broken for the `jarvis` repo and was not used here, per instructions.

Claude-Session: https://claude.ai/code/session_01BXaYUhmfKv15hAw4KY5tQF

### DIFF
--- a/frontend/src/views/OnboardingView.vue
+++ b/frontend/src/views/OnboardingView.vue
@@ -592,10 +592,10 @@
 							 save_llm_pool; this step is the post-save readiness handoff.
 							 v-show (not v-if) so the editor stays MOUNTED while its own save()
 							 is still awaiting. ===== -->
-						<section v-else-if="state.step === 'connect'" class="jv-ob-screen">
-							<div class="jv-ob-body">
+						<section v-else-if="state.step === 'connect'" class="ob-screen">
+							<div class="ob-body">
 								<div v-show="state.finishing">
-									<div class="jv-ob-head">
+									<div class="ob-head">
 										<h1>Setting up {{ agentName }}</h1>
 										<p>
 											{{
@@ -604,19 +604,23 @@
 											}}
 										</p>
 									</div>
-									<div class="jv-ob-setup-net">
+									<!-- min-height (not h-full) is load-bearing: SetupNeuralNet's
+										 canvas fills via absolute+inset-0, and percentage heights
+										 don't resolve against a min-height parent - see its own
+										 comment. Don't change this to a fixed h-*. -->
+									<div class="relative mt-2 min-h-[380px] flex-1">
 										<SetupNeuralNet :dark="dark" />
 									</div>
 								</div>
 								<div v-show="!state.finishing">
-									<div class="jv-ob-head">
+									<div class="ob-head">
 										<h1>Give {{ agentName }} a brain</h1>
 										<p>
 											Pick which AI powers {{ agentName }}. You can change
 											this anytime in Settings → AI models.
 										</p>
 									</div>
-									<div class="jv-ob-connect">
+									<div class="mx-auto max-w-[640px]">
 										<LlmPoolEditor
 											ref="poolRef"
 											:editable="true"
@@ -632,38 +636,29 @@
 										:message="state.finishNote"
 									>
 										<template #action>
-											<button
-												class="jv-ob-btn jv-ob-btn-primary"
+											<Button
+												variant="solid"
+												:label="`Continue to ${agentName}`"
 												@click="forceContinue"
-											>
-												Continue to {{ agentName }}
-											</button>
+											/>
 										</template>
 									</Banner>
 								</div>
 							</div>
-							<div v-if="!state.finishing" class="jv-ob-foot">
+							<div v-if="!state.finishing" class="ob-foot">
 								<!-- No Back on a reconciled resume: signup/payment already completed
 									 in a previous session, so there is no local pay/review context to
 									 go back to (re-running startSignup there would double-sign-up). -->
 								<button
 									v-if="!state.reconciledConnect"
-									class="jv-ob-back"
+									class="ob-back"
 									:disabled="savingConnect"
 									@click="goBack"
 								>
-									<svg
-										width="14"
-										height="14"
-										viewBox="0 0 24 24"
-										fill="none"
-										stroke="currentColor"
-										stroke-width="1.5"
-										stroke-linecap="round"
-										stroke-linejoin="round"
-									>
-										<path d="m15 18-6-6 6-6" /></svg
-									>Back
+									<FeatherIcon
+										name="chevron-left"
+										class="h-3.5 w-3.5 text-ink-gray-5"
+									/>Back
 								</button>
 								<span v-else></span>
 								<!-- Always rendered; disabled until the editor reports a savable config,
@@ -681,9 +676,9 @@
 						<!-- ===== Self-host (reached via the quiet Plan-step link; logic
 							 unchanged, field names/args match test_connection /
 							 save_self_hosted verbatim) ===== -->
-						<section v-else-if="state.step === 'selfhost'" class="jv-ob-screen">
-							<div class="jv-ob-body">
-								<div class="jv-ob-head">
+						<section v-else-if="state.step === 'selfhost'" class="ob-screen">
+							<div class="ob-body">
+								<div class="ob-head">
 									<h1>Connect your openclaw</h1>
 									<p>
 										Point {{ agentName }} at <b>your own</b> openclaw server.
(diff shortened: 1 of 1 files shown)


kavin-114:
## Self-review pass (manual — the code-review workflow tool isn't available in this session, so this is a direct, careful manual review of the full diff rather than the automated finder/verifier pipeline)

Checked: correctness of every `v-model` binding against original state fields
(all 9 match, no typos), keyboard/focus-visible coverage for every raw
interactive element left after the migration (9 elements: 5 back-buttons, 1
quiet link, 1 plan-card radio group, 2 provider-chooser buttons — all
confirmed covered, either by a component's own built-in focus ring or an
explicit `focus-visible:ring`/`.ob-back:focus-visible` rule), no `dark:`
variants, no `!important`, no stray inline `:style` objects beyond the two
documented exceptions (`paletteVars` and the Razorpay wordmark's brand
navy), and Banner/JvCombo/LlmPoolEditor integration points untouched.

**One real bug found and fixed** (commit `ef066989`, pushed): every
`<FormControl id="...">` I'd added (7 call sites: the 5 Details-step text
fields plus the 2 Self-host fields) had a broken label. `FormControlProps`
doesn't declare an `id` prop, so my explicit `id` fell through `$attrs` onto
the inner `TextInput`'s `v-bind` spread and won there (Vue's attribute-merge
order), while `FormLabel` stayed wired to FormControl's own separately
generated `useId()` value — the two never agreed, so `<label for="fc-N">`
pointed at nothing and the real `<input id="jv-ob-email">` had no
programmatic label. Confirmed via `grep -rl "<FormControl"` (30 files) that
no other call site in the app passes an explicit `id` — this was a bug my
new code introduced, not an established pattern. Fix: drop the `id` prop
entirely (nothing in `<script setup>` ever read those DOM ids — they were
copy-pasted from the pre-migration `<input id="jv-ob-email">` markup, whose
`<label for>` pairing `FormControl` now owns internally) and let FormControl
self-associate. The Company field's `<label for="jv-ob-company"> +
<JvCombo id="jv-ob-company">` pairing is a *different*, correct case —
`JvCombo` does declare a real `id` prop and wires it to its own input — and
was left alone.

No other correctness/security/a11y issues found. This PR has no backend/
DocType/permission surface (presentation-only Vue template + scoped CSS), so
most of the usual Frappe review checklist (SQL, permissions, migrations,
concurrency) doesn't apply here.

PR #498, opened 2026-07-27
fix(chat): stop sending jarvis-pool to an openclaw-direct BYO api-key pool

kavin-114:
## The defect

`jarvis-fleet-agent` recently shipped **openclaw-direct**: a `/llm-pool` whose every enabled
model is a BYO **api-key** credential is now rendered with each credential as its own provider
*inside* openclaw, native `agents.defaults.model.{primary,fallbacks}` failover, and **no
bifrost/cliproxy sidecar at all** (`pool_render.is_byo_direct`). A pool containing any
subscription still takes the Bifrost+cliproxy path, because cliproxy is what serves an OAuth blob.

The customer app had not caught up.

1. `pool_serialize.compute_proxy_active` returned `True` for **any** 2+-model pool, regardless of
   credential type. For a 2x api-key pool that is now wrong: no proxy is deployed.
2. `chat/turn_handler.py:_session_model_for` therefore sent the Bifrost-only placeholder
   `POOL_VIRTUAL_MODEL = "jarvis-pool"` on every default ("Auto") turn.
3. Against an openclaw-direct container that model id matches nothing, so the turn fails with
   `model_not_found`. Worse: because that is an *explicit bad-id rejection* rather than an upstream
   failure, openclaw's native failover does **not** engage (`next=none`).

**Customer impact:** someone who adds two BYO API keys via Settings -> AI models gets a correctly
provisioned container whose default chat is dead.

Verified live on `jarvis.proxy` today with a GLM + ollama api-key pool: an "Auto" turn returned
"Something went wrong" and the container log showed
`model=claude-sonnet-4-6 provider=zai_coding-0 error=... model_not_found`. Explicitly picking a
real model in the UI worked (12.1s reply, plus a real ERP tool call), and forced failover worked
(`candidate_failed reason=auth 401` -> `candidate_succeeded ollama-1/qwen2.5:3b`). Only the
default/"Auto" path was broken.

## The fix

`proxy_active` was answering two different questions at once. Split them:

| helper | question | drives |
|---|---|---|
| `compute_pool_mode(settings)` (**new**, the old body verbatim) | "is this config a POOL?" | sync leg (`/llm-pool` vs `/llm-creds`), chat-readiness gate, installed-apps resync leg, converged-marker choice, chat pin validation |
| `compute_proxy_active(settings)` (**narrowed**) | "is a Bifrost+cliproxy sidecar actually DEPLOYED?" | the `jarvis-pool` placeholder, Monitor usage, the Connection card |

`compute_proxy_active` is now exactly "at least one enabled model is a `subscription`" — the
inverse of the fleet's `pool_render.is_byo_direct`. It still **implies** `compute_pool_mode` (a
subscription forces pool mode), so it is strictly narrower and nothing that meant "pool" changed
behaviour for any existing tenant.

`_session_model_for` then picks the name by what is actually in front of the container:

* proxy deployed -> `jarvis-pool` (unchanged; Bifrost's catch-all rule expands it)
* openclaw-direct pool -> the pool's **primary real model**
* not a pool -> unchanged (`settings.llm_model`)

The primary is the first enabled row by `(order, provider, model)`, matching the fleet's
`llm_proxy.render._common.ordered`, emitted as a **bare** model id (`glm-4.6`). Bare is the form
this app already uses everywhere (`llm_model`, `model_override`, the Settings model picker) and it
is exactly what the live container accepted when a model was picked by hand. We deliberately do not
reconstruct the fleet's `<provider>-<idx>/<model>` qualified ref: that id scheme belongs to the
fleet's renderer and a copy here would rot silently the moment its indexing changed.

### Call sites moved from `proxy_active` to `compute_pool_mode`

| file | why |
|---|---|
| `jarvis_settings._on_update_unified_llm` | routes the sync; pushing a 2-key pool's `models[0]` through `/llm-creds` would cut the container to one credential |
| `jarvis_settings._enqueued_sync_via_admin_pool` | the run-time re-validate would otherwise **skip the push** for every BYO pool |
| `jarvis_settings.reconcile_pending_llm_sync` | `_stamp_converged_ok(is_pool=...)` picks `llm_pool_synced_at` vs `llm_direct_synced_at` |
| `onboarding._reconcile_pending_applying` | same marker choice, from the UI poller |
| `account.is_ready_for_chat` | a BYO pool stamps `llm_pool_synced_at`; gating it on the direct marker would strand it at `llm_provisioning` forever |
| `installed_apps_sync._pool_active` | the single-model resync leg would knock a pool container down to one credential |

`oauth/api.get_direct_subscription_status` keeps reading `proxy_active` — under the new meaning it
is *more* correct there (`is_single_subscription_pool` and the Connection-card gate both want
"sidecar deployed").

### Two smaller wrongs fixed for free

* `account.get_llm_usage` no longer round-trips to admin for Bifrost usage a BYO pool has no
  Bifrost to produce.
* `account.get_llm_connection_status` no longer offers a cliproxy auth profile that does not exist,
  so `ConnectionPane` stops rendering a misleading "Not connected" for a working BYO pool.

The run-time skip status string is renamed `no longer proxy-valid` -> `no longer pool-valid` (it is
surfaced in the onboarding banner); the three frontend mentions were updated with it.

### Interaction with the known "direct-mode stale pin" bug

There is a separate known issue where a session can carry a stale explicit model pin. This change
does not touch it and does not widen it. The one thing worth noting: `_resolve_model_and_provider`'s
pool branch — the branch that **validates** a pin against the enabled model names and drops a stale
one — is now keyed on `compute_pool_mode`, not `proxy_active`. Had it stayed on `proxy_active`, a
BYO pool would have fallen into the unvalidated direct branch and a stale pin naming a removed model
would have been forwarded verbatim. Keying on pool mode keeps that validation, and
`test_session_model_for_stale_pin_resets_to_the_pool` now pins that a stale pin on a BYO pool resets
to the pool's primary.

## Tests

New / changed in `jarvis/tests/test_unified_llm_config.py`:

* `test_compute_proxy_active_false_for_a_pure_api_key_pool` — **fails without the fix:**
  `AssertionError: True is not false`
* `test_session_model_for_byo_pool_with_no_pin_names_a_real_model` — **fails without the fix:**
  `AssertionError: 'jarvis-pool' != 'gpt-4o'` (the literal defect)
* `test_session_model_for_stale_pin_resets_to_the_pool` — same failure without the fix
* `test_api_key_pool_is_not_proxy_active` — **fails without the fix:** `AssertionError: 1 != 0`
* `test_subscription_pool_is_proxy_active` / `test_session_model_for_subscription_pool_with_no_pin_names_the_pool`
  — the Bifrost path still gets `jarvis-pool`; both pass before **and** after (no regression)
* `test_session_model_for_single_model_tenant_is_unchanged` — a one-model config is not a pool
* `compute_pool_mode` coverage (2 models / preset / lone subscription / single model)
* `compute_proxy_active` coverage incl. a preset pool and a **disabled** subscription row
* `pool_primary_model` coverage: lowest `order` wins, provider-then-model tie-break, disabled rows
  skipped, blank when not a pool, junk `order` does not raise (the module's never-raises contract)
* `test_pool_active_is_derived_from_models_not_the_proxy_flag` in `test_installed_apps_sync.py`

Fixture hardening: `TestFT5ChatWorkerPoolAwareness._pool_of_two` used to **force** `proxy_active=1`
in the DB after saving an all-api-key pool, which is exactly what hid this defect — it now lets
`on_update` derive the flag. `_set_pool_gate_state` and the three `reconcile_pending_llm_sync` tests
faked pool state with `db_set("proxy_active", 1)` on an *empty* `models[]`; they now seed real rows,
which also makes them order-independent. The two `v1_10` backfill-patch tests explicitly write the
legacy flag, because that patch grandfathers pre-change tenants and deliberately reads the stored
field (at migrate time `on_update` has not re-derived it). The patches themselves are untouched.

Local run (`bench --site site.jarvis run-tests --module ...`), all green: `test_unified_llm_config`
(83 + 55), `test_llm_pool_endpoints` (18), `test_installed_apps_sync` (14), `test_settings` (9 + 4),
`test_settings_on_update` (37), `test_llm_monitor` (5), `test_oauth_api` (57 + 7),
`test_subscription_accounts_roundtrip` (6), `test_usage_per_model` (10), `test_dev` (3),
`test_account` (19), `test_turn_handler` (4 + 7), `test_chat_worker` (47), `test_chat_api` (61 + 5),
`test_turn_handler_thinking` (5), `test_turn_handler_report_context` (14). Plus
`node --test frontend/src/onboarding/steps.test.js` (25).

One pre-existing local failure, `test_onboarding_sync.test_start_signup_persists_api_key_secret_on_verification_response`,
reproduces identically on the unmodified branch — it is `site.jarvis` DB pollution (a leftover
`jarvis_admin_customer_password` in `__Auth`), not this change.

## Not verified

* Not re-run against the live container. The fix is derived from the fleet's renderer
  (`pool_render.build_pool_providers` + `templates/openclaw.json.j2`) plus today's live evidence
  that a bare model id resolves in an openclaw-direct container; an end-to-end "Auto" turn on
  `jarvis.proxy` should confirm it after merge.
* The tie-break on `provider` is the *stored/wire* provider id. The fleet ties on
  `bifrost_provider()`, which synthesizes a per-model name for custom-endpoint vendors. The two can
  only disagree for rows that already share an `order`, and since we send a bare model id (which
  openclaw resolves against every registered provider) a disagreement would pick a different-but-valid
  model rather than break the turn.

---

## Review follow-ups (second commit)

Two confirmed findings from a review pass, both fixed in `9f1a90e4`:

1. **Two pre-existing tests were order-dependent, not just stale.**
   `test_pending_applying_flips_to_ok_when_admin_reports_ready` and
   `test_proxy_active_pool_is_ready_for_chat` faked a pool tenant with
   `db_set("proxy_active", 1)` on an **empty** `models[]`. Once the sync leg and the readiness gate
   derive pool mode from `models[]`, that is no longer a pool tenant. The first one **fails in
   isolation** (verified: `AssertionError: None is not true` — it stamps `llm_direct_synced_at`
   instead of `llm_pool_synced_at`); it passed in the suite only because an alphabetically earlier
   test in the same class leaves committed rows behind. Both now seed real rows, which also makes
   them order-independent — the sharded runner sorts files alphabetically, so an order-dependent
   test is a latent CI break.

2. **`BillingMeteringPane.vue` read `proxy_active` as "is this a pool".**
   For a BYO api-key pool it rendered `Mode: Direct` directly above an "Active pool" card listing
   both of that tenant's models, and the usage card said *"This tenant runs a single model
   (direct)"*. The mode label now derives from the enabled-model count
   (`Pool (proxied)` / `Pool (direct failover)` / `Direct`), and the metering copy says what is
   actually true of both no-sidecar shapes.

`GeneralPane.vue` and `ConnectionPane.vue` were checked and are correct as-is: their `isProxy` is
about "does a sidecar / auth profile exist", which is exactly the narrowed meaning.

### Reviewed and deliberately not changed

* **Tie-break divergence (latent, unreachable today).** `pool_primary_model` ties on the *wire*
  provider id; the fleet ties on `bifrost_provider()`, which synthesizes `compat-<model>` for the
  openai_compat/zai family. For two enabled rows sharing the same `order` the two can pick opposite
  primaries (e.g. `groq` vs `openai_compat`). Not reachable through the product: `buildSaveModels`
  in `LlmPoolEditor.vue` and `presetToModels`/`seedRowsFromConfig` in `pool.js` always re-derive
  `order: i` from array position, so duplicate `order` values cannot be saved from the UI. Left
  alone rather than importing the fleet's Bifrost-specific naming into the customer plane for a case
  the UI cannot produce — the docstring now states the limit explicitly.
* **The `v1_10` / `v2_00` backfill patches still read the raw `proxy_active` field.** That is
  correct: they grandfather pre-change tenants, and at migrate time `on_update` has not re-derived
  the flag, so the stored value is exactly the old-rule value they were written against.


kavin-114:
### Local CI — `Aerele-RnD/jarvis` branch fix-byo-direct-pool-default-model

Tested commit `00aacaf6a747` · ✅ **PASS** · 24m57s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 52s |
| `tests-shard1` | ✅ PASS | 9/9 | 6m32s |
| `tests-shard2` | ✅ PASS | 9/9 | 6m19s |
| `tests-shard3` | ✅ PASS | 9/9 | 5m29s |
| `tests-shard4` | ✅ PASS | 9/9 | 5m36s |
| `coverage` | ✅ PASS | 5/5 | 5s |

---

kavin-114:
### Local CI — `Aerele-RnD/jarvis` branch fix-byo-direct-pool-default-model

Tested commit `9f1a90e4aba5` · ❌ **FAIL** · 339m25s

|---|---|---|---|
| `lint` | ❌ FAIL | 2/3 | 15m07s |
| `tests-shard1` | ✅ PASS | 9/9 | 77m48s |
| `tests-shard2` | ✅ PASS | 9/9 | 143m02s |
| `tests-shard3` | 💥 ERROR | 0/0 | 15m42s |
| `tests-shard4` | ✅ PASS | 9/9 | 87m34s |
| `coverage` | ✅ PASS | 5/5 | 7s |

**`lint` — FAIL**

- failed at step **actions/setup-node@v4** (exit 1)

<details><summary>log excerpt</summary>

```
# job: lint
# repo: Aerele-RnD/jarvis
# sha: 9f1a90e4aba511c929b2c86226047aca37336940
checked out 9f1a90e4 -> /ci/workspace
Python 3.11.15
curl: (18) Transferred a partial file
xz: (stdin): Unexpected end of input
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
  Error: failed to download archive for 20.20.2
```

</details>

**`tests-shard3` — ERROR**

> docker pull --quiet redis:7 failed: Error response from daemon: pull access denied for redis, repository does not exist or may require 'docker login'


<details><summary>log excerpt</summary>

```
# job: tests-shard3
# repo: Aerele-RnD/jarvis
# sha: 9f1a90e4aba511c929b2c86226047aca37336940
!! docker pull --quiet redis:7 failed: Error response from daemon: pull access denied for redis, repository does not exist or may require 'docker login'
```

</details>

---

kavin-114:
### Local CI — `Aerele-RnD/jarvis` branch feat/jv-spinner

Tested commit `44da2d36af51` · ✅ **PASS** · 27m36s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 36s |
| `tests-shard1` | ✅ PASS | 9/9 | 6m11s |
| `tests-shard2` | ✅ PASS | 9/9 | 6m22s |
| `tests-shard3` | ✅ PASS | 9/9 | 5m34s |
| `tests-shard4` | ✅ PASS | 9/9 | 8m42s |
| `coverage` | ✅ PASS | 5/5 | 6s |

---

PR #510, opened 2026-07-28
feat(tools): non-blocking holiday advisory on employee-date writes

venkat102:
## What

Customer ask: whenever Jarvis touches an employee record on a date-related thing, it should **check that employee's holiday list and warn (never block)**.

When a Jarvis *write* touches a curated HR doctype on an employee's activity date that is a **holiday or weekly-off**, the tool result now carries a non-blocking `warnings` list (e.g. *"Priya: 2026-08-15 is a holiday (Independence Day) on their holiday list."*). Detection is deterministic code — not a persona instruction.

Pairs with `jarvis-persona` PR (a one-line nudge telling the agent to relay `warnings`). **Merge this first / same window.**

## Scope

Curated `doctype → (employee field, date span(s))` map — deliberately tight, so we don't warn on identity dates (DOB/DOJ/relieving), payroll periods, or comp-off (working a holiday is the point):
- **Attendance** (attendance_date), **Attendance Request** (from/to), **Employee Checkin** (time), **Shift Assignment** (start/end), **Leave Application** (from/to).

Warns on **both named holidays and weekly-offs**, labelled distinctly. Ranges name each affected day. Reuses ERPNext's holiday-list resolver (`get_holiday_list_for_employee`), then queries `Holiday` rows for the `weekly_off` flag.

## Guarantees

- **Never blocks, never raises into the write path** — any internal failure returns no advisory.
- **Silent no-op** without erpnext / holiday list / employee, or for any non-curated doctype.
- **Byte-identical output for every non-HR write** — `attach` adds nothing when there are no advisories.

## Surface (no plugin change)

`warnings` rides inside the tool result `data`. Confirmed live on every path: on a direct/auto-apply write it reaches the agent intact; on a gated write the park-time preview carries it to the model *before* confirmation and the human's receipt chip shows it. (Deferred follow-up: the post-confirm continuation text is name-only — a pre-existing limitation, not part of this feature.)

## Review

Cyclic Fable review to green (2 cycles). Cycle 1 found three defects, all fixed + mutation-verified:
- **F1** — the holiday list was resolved as of *today*, not the activity date; a **backdated attendance** (the flagship attendance-regularization case) under HRMS silently checked the wrong year's list. Now `as_on=<activity date>`.
- **F2** — `attach()` could **crash the write** if a doctype had a Custom Field named `warnings`; guarded (piercing the never-break-a-write invariant).
- **F3** — the never-raise handler could swallow the write's *own* message; now truncates only what the advisory pushed.

## Tests

`test_holiday_advisory` **14** + `test_holiday_advisory_wiring` **5**, and the `test_create_doc`/`test_create_docs`/`test_update_doc` regressions green on `jarvis-test.localhost`. Every security/correctness guard mutation-verified.

## Deferred (non-blocking, noted in review)

- Gated-write post-confirm continuation text (name-only) — pre-existing.
- A range straddling a Holiday-List-Assignment boundary resolves against the start-date list (same limitation HRMS's own helpers have; fail-quiet).
- Advisory volume cap for very long spans ("first N + more").

### DIFF
--- a/jarvis/tests/test_holiday_advisory.py
+++ b/jarvis/tests/test_holiday_advisory.py
@@ -1,0 +1,106 @@
+from unittest.mock import patch
+
+import frappe
+from frappe.tests.utils import FrappeTestCase
+
+from jarvis.tools import _holiday_advisory as ha
+
+HL = "JV Holiday Advisory Test"
+NAMED_DATE = "2026-08-15"  # Independence Day (weekly_off = 0)
+WOFF_DATE = "2026-08-16"  # a Sunday (weekly_off = 1)
+PLAIN_DATE = "2026-08-17"  # ordinary working day
+
+
+def _doc(doctype, **fields):
+	return frappe._dict(doctype=doctype, **fields)
+
+
+class TestHolidayAdvisory(FrappeTestCase):
+	@classmethod
+	def setUpClass(cls):
+		super().setUpClass()
+		frappe.set_user("Administrator")
+		if frappe.db.exists("Holiday List", HL):
+			frappe.delete_doc("Holiday List", HL, force=True, ignore_permissions=True)
+		frappe.get_doc(
+			{
+				"doctype": "Holiday List",
+				"holiday_list_name": HL,
+				"from_date": "2026-01-01",
+				"to_date": "2026-12-31",
+				"holidays": [
+					{"holiday_date": NAMED_DATE, "description": "Independence Day", "weekly_off": 0},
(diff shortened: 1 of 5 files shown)


venkat102:
Superseded by a skill-based approach (per navin): for a non-blocking advisory, a deterministic check on every create/update is unnecessary. Replacing with guidance in the hrms-hr persona skill (the agent checks the holiday list via the existing get_holidays_for_employee tool and warns). Closing.

PR #514, opened 2026-07-28
feat(chat): dual persona (Jarvis/Jara) + composer model/effort & persona pickers

l0gesh29:
## What

Adds a second persona, **Jara** (calmer, more jovial tone — identical tools, permissions and data), as a per-user choice, and moves the model/effort and persona selectors into the chat composer (ChatGPT / Claude-web style).

## Changes

**Backend**
- `Jarvis User Settings`: new `preferred_persona` Select (Jarvis default / Jara).
- `turn_handler._persona_clause`: folds the chosen persona into the trusted `[Context:]` line each turn via a fresh DB read. Jarvis/unset emits nothing, so the default turn stays byte-identical and pays zero extra tokens.
- `user_settings_api`: read + validated write of `preferred_persona`.

**Frontend**
- `ModelEffortPicker`: composer pill → model list + an **Effort** side-flyout. Opens upward and, being right-docked, leftward so it never clips the screen edge.
- `PersonaPill`: composer pill → Jarvis / Jara dropdown, each with its living mark.
- `Composer`: new `#right-toolbar` slot. Toolbar split — **left:** attach + persona; **right:** dictation mic + model/effort + send.
- Removes the model/effort control from the top bar and the interim settings-pane persona picker.

## Switching behaviour
Persona is read fresh from the DB on every turn, so switching takes effect on the **next message** — no reload, no reprovisioning.

## Testing
- `vitest`: 43/43 (composer snapshots updated for the new toolbar structure).
- Persona-preference unit tests in `test_user_settings.py`.
- SPA build green.

## Note
The companion `jarvis-persona` change (AGENTS.md "Personas" section defining Jara + version bump) ships separately and is reseeded via the Fleet Console.

---
## Deploy (corrected runbook)
Persona v0.63.0 must land FIRST:
1. From admin (jarvis_admin_v2): `fleet.content.push_content(cell=..., bundles=["persona"], restart=1)` — bumps `persona/current` to v0.63.0 and rolls tenant restarts. `restart=1` is mandatory (a pin bump is inert until the entrypoint re-runs the RO bind-mount symlinks). **`reseed_persona` is wrong here**: it is restart-only and re-applies whatever version is already pinned.
2. Verify `content_status` reports v0.63.0 on every host.
3. Deploy the bench: `bench --site <site> migrate` (preferred_persona + persona_enabled fields, patch v2_09) + SPA rebuild.

### DIFF
--- a/frontend/src/components/chat/CheckMark.vue
+++ b/frontend/src/components/chat/CheckMark.vue
@@ -1,0 +1,26 @@
+<template>
+	<!-- Selected-row check glyph, shared by the composer pills so the SVG is not
+	     forked (was an inline template SVG in PersonaPill and an h() render in
+	     ModelEffortPicker). Colour follows the surrounding text. -->
+	<svg
+		class="jv-check"
+		width="15"
+		height="15"
+		viewBox="0 0 24 24"
+		fill="none"
+		stroke="currentColor"
+		stroke-width="2.2"
+		stroke-linecap="round"
+		stroke-linejoin="round"
+		aria-hidden="true"
+	>
+		<path d="M20 6 9 17l-5-5" />
+	</svg>
+</template>
+
+<style scoped>
+.jv-check {
+	flex: none;
+	color: var(--text);
+}
+</style>
--- a/frontend/src/components/chat/Composer.vue
+++ b/frontend/src/components/chat/Composer.vue
@@ -227,62 +227,66 @@
 					</svg>
 				</button>
 			</slot>
-			<span
-				style="margin-left: auto; font-size: 11px; color: var(--text-3); margin-right: 4px"
-				>{{ busy ? "Stop" : "Enter ↵" }}</span
-			>
-			<button
-				v-if="busy"
-				@click="emit('stop')"
-				title="Stop generating"
-				style="
-					width: 32px;
-					height: 32px;
-					display: flex;
-					align-items: center;
-					justify-content: center;
-					background: var(--cta);
-					border: none;
-					border-radius: 8px;
-					cursor: pointer;
-				"
-			>
-				<svg width="13" height="13" viewBox="0 0 24 24" fill="var(--cta-fg)">
-					<rect x="6" y="6" width="12" height="12" rx="2.5" />
-				</svg>
-			</button>
-			<button
-				v-else
-				class="jv-sendbtn"
-				:class="{ ready: sendable }"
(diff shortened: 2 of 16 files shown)


l0gesh29:
### Cycle-3 review findings addressed (commits 7cd91f6, 346db4a)

Two Criticals from the cycle-3 review, fixed and verified against Frappe source:

- **C1 - kill switch failed CLOSED.** Both readers gated on `get_single_value(...) is None` to mean "unset = ON", but Frappe coerces an unset `Check` to `0` (never `None`), so an un-backfilled bench read the switch as OFF (pill hidden, clause silent). Replaced both with one canonical `tabSingles` row-existence probe (`turn_handler.persona_feature_enabled`): no row = ON, only a stored `0` = OFF. It never touches the meta, so it also can't throw in the migrate window (closes the old pill-vs-clause inversion, I8) and centralises the duplicated predicate.
- **C2 - shipped dead for new tenants.** NULL=ON fixes the read path, but a full `Jarvis Settings.save()` (onboarding LLM sync) coerces the unset Check to `0` via `get_valid_dict`, silently flipping it OFF. The backfill is restored, now **seed-when-absent** (never clobbers an admin's explicit `0`) with `update_modified=False`, mirroring `roles._seed_personalise_settings_defaults`.

Important/Minor also addressed: **I1** (fail-open clause logs at ERROR, not the prod-silent WARNING), **I6** (composer effort button opens on click instead of toggling shut after hover), **I7** (server omits `preferred_persona` on a read failure so the SPA keeps its pill; client already guards on the key), **I10** (`_s()` coercion at the admin limit endpoints now tested), the `_settings_payload` migrate-window guard, the backfill idempotency/`update_modified` Minors (subsumed by the seed-when-absent restore), persona-write rollback-to-confirmed + in-flight timeout, and the JarvisMark orb-highlight occlusion.

**I2/I3** (pill lags a mid-session admin flip): kept as-is and documented inline. The server clause re-reads the switch every turn, so behaviour is always correct; the pill is voice-only cosmetic lag that self-heals on reload.

Tests exercise the real mechanism with the `tabSingles` row **deleted** (fails on the old `get_single_value` code), the fail-open branches, the seed-when-absent/no-clobber backfill, and the full-save write path. Local: 49/49 in `test_user_settings`, SPA build clean.

venkat102 (approved):

Issue #299, opened 2026-07-14
Chat model picker is inert when the LLM proxy is active (the pin is silently ignored)


kavin-114:
## What happens

With a multi-model pool (`proxy_active=1`), picking a model in the chat header's model menu **does nothing**. The pin is saved, the pill shows it as active, and Bifrost ignores it.

## Evidence (live, tenant `jarvis-pool-ae275f`)

Pool order: `gpt-5.5` (subscription, primary) → `claude-sonnet-4-6` (api key, fallback).

Asked Bifrost explicitly for `claude-sonnet-4-6` — exactly what a chat pin sends:

```
asked=gpt-5.5             served=gpt-5.5             PRIMARY     error    upstream_credentials_exhausted
asked=claude-sonnet-4-6   served=claude-sonnet-4-6   FALLBACK#1  success
```

Bifrost **discarded the requested model** and substituted the routing rule's target. It only reached Claude because Claude is the *fallback*.

## Root cause

`llm_proxy` renders a single global routing rule that matches unconditionally:

```json
{
  "id": "rule-site-failover",
  "cel_expression": "true",        // <-- matches EVERY request
  "scope": "global", "priority": 100,
  "targets":   [ { "provider": "cliproxy-subs", "model": "gpt-5.5" } ],
  "fallbacks": [ "compat-claude-sonnet-4-6/claude-sonnet-4-6" ]
}
```

openclaw is configured with the virtual model `jarvis-pool`, so the rule exists to expand that into the failover chain. But because the expression is `true`, it *also* rewrites explicit model requests.

## Consequences

1. **The chat model picker is a lie in proxy mode.** `ChatView.selectModel()` saves `model_override`, the server validates it against the pool, the pill renders it — and routing ignores it. Only the pool ORDER (Settings → AI models) decides.
2. **Every turn burns a failed call on a dead primary**, even when the customer pinned the model that works. On a metered subscription that is a real cost, and it adds a round-trip of latency to every message.

## Proposed fix

Scope the rule to the pool's virtual model only:

```diff
- "cel_expression": "true"
+ "cel_expression": "request.model == 'jarvis-pool'"
```

Then:
- default turns (openclaw sends `jarvis-pool`) → full failover chain, unchanged;
- a **pinned** model → routed straight to its provider, no wasted primary call, pin means what it says.

A pinned model would then have no failover, which is the honest reading of "I explicitly want this model."

Needs verification that Bifrost's CEL context exposes the requested model under that name.

[...]

PR #517, opened 2026-07-28
fix(chat): tell openclaw when WE picked the model, so failover survives

kavin-114:
## The bug

A BYO api-key pool **never failed over**. Measured live on `jarvis.proxy` with two real vendor keys: a genuine 429 on the pool primary, with a healthy second model configured, produced

```
[model-fallback/decision] decision=candidate_failed
  requested=gemini-0/gemini-2.5-pro  reason=rate_limit  next=none
```

and surfaced the error to the customer instead of switching vendors.

This matters beyond one tenant: **openclaw-native failover is the entire premise of retiring the per-tenant CLIProxyAPI/Bifrost sidecars.** On the evidence above it looked like the premise was false.

## It was ours, and it is one field

Read out of the shipped 2026.6.8 bundle rather than inferred. `resolveEffectiveModelFallbacks`:

```js
if (!hasSessionModelOverride) return agentFallbacksOverride;   // no pin -> chain applies
if (!(modelOverrideSource === "auto" || ...)) return [];       // pinned -> NO chain
```

and the `sessions.patch` handler:

```js
entry.modelOverrideSource = patch.modelOverrideSource === "auto" ? "auto" : "user";
```

Naming a model makes the session model-**overridden**, and an override that does not declare itself automatic drops the agent's whole `model.fallbacks` chain: zero candidates, `fallbackConfigured` false, `resolveRunFailoverDecision` returns `surface_error`. **Omitting the field is not neutral — it reads as `"user"`.**

We began naming a model on every unpinned turn in #299/#498, for good reasons: openclaw remembers a session's model across turns, so "send nothing" left a stale pin live forever, and the `jarvis-pool` placeholder is a hard `model_not_found` against a sidecar-free container. That fix was correct. It just never told openclaw *who chose the model*.

## The change

`_session_model_for` returns `(model, provider, source)`; both transports forward it as `modelOverrideSource`.

`"user"` remains the default **on purpose** — someone who explicitly selects Gemini should get an error, not a silent switch to another vendor mid-conversation. Only a literal `"auto"` unlocks the chain, so a typo or a future source name fails closed.

Both "we chose it" cases report `auto`: an unpinned pool turn, and resetting a pin naming a model the customer has since deleted.

## Tests

Three new, and all fail against the current code (the first two raise on the old signature/arity):

- wire payload carries `modelOverrideSource: "auto"` when we picked the model
- an unrecognised source degrades to `"user"`, never `"auto"`
- `_session_model_for` reports `auto` for both auto paths and `user` for a real pin

The pre-existing exact-params test was asserting the old wire contract and is updated to the new one.

## Verification still owed

Re-run the live failover with the primary on `gemini-2.5-pro` (a guaranteed 429 on a free key) and confirm the log shows `candidate_succeeded` on the second vendor instead of `next=none`. I'll post that here before marking ready.

### DIFF
--- a/jarvis/chat/openclaw_client.py
+++ b/jarvis/chat/openclaw_client.py
@@ -689,15 +689,32 @@ class OpenclawSession:
 		session_key: str,
 		model_ref: str,
 		*,
+		source: str = "user",
 		timeout_s: float = CONNECT_TIMEOUT_SECONDS,
 	) -> None:
 		"""Point the session at ``model_ref`` (``"<provider>/<model>"`` or
 		bare ``"<model>"``) via ``sessions.patch``. chat.send has no per-turn
 		model param, so per-conversation overrides are applied to the session
-		before the send."""
+		before the send.
+
+		``source`` MUST be ``"auto"`` when WE picked the model rather than the
+		customer. openclaw drops the agent's entire model.fallbacks chain for an
+		overridden session unless the override declares itself automatic. From the
+		2026.6.8 bundle -- the sessions.patch handler:
+
+		    entry.modelOverrideSource = patch.modelOverrideSource === "auto" ? "auto" : "user"
+
+		and ``resolveEffectiveModelFallbacks`` returns ``[]`` for any non-"auto"
+		override. Omitting the field is therefore NOT neutral: it reads as "user" and
+		silently disables failover (see ``turn_handler._session_model_for``).
+		"""
 		self._request(
 			"sessions.patch",
-			{"key": session_key, "model": model_ref},
+			{
+				"key": session_key,
+				"model": model_ref,
+				"modelOverrideSource": "auto" if source == "auto" else "user",
+			},
 			timeout_s=timeout_s,
 		)
 
--- a/jarvis/chat/prepare.py
+++ b/jarvis/chat/prepare.py
@@ -135,7 +135,7 @@ def run_prepare(run_id: str, relay_target_id: str | None = None) -> dict:
 		return {"ok": False, "reason": "self_host"}
 
 	gateway_url = (settings.agent_url or "").replace("http://", "ws://").replace("https://", "wss://")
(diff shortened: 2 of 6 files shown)


kavin-114:
## Do not merge — this approach does not work on openclaw 2026.6.8

Caught in live testing on `jarvis.proxy`, **after** a clean `/code-review high` pass (which found nothing — a static review cannot see this).

Deploying this branch and sending a real chat turn produces:

```
[ws] ⇄ res ✗ sessions.patch 4ms errorCode=INVALID_REQUEST
     errorMessage=invalid sessions.patch params: at root: unexpected property 'modelOverrideSource'
```

**Every session patch fails, so every chat turn dies with "I couldn't reach the assistant."** This would have been a total chat outage, not an improvement.

### Why the unit tests missed it

They stub `_request`, so they assert the payload we *intend* to send and never exercise the gateway's schema validation. The wire contract is only observable against a real container.

### The diagnosis is still correct — the delivery mechanism is not

The root cause stands, and is confirmed from the shipped bundle:

```js
// resolveEffectiveModelFallbacks
if (!hasSessionModelOverride) return agentFallbacksOverride;
if (!(modelOverrideSource === "auto" || ...)) return [];
```

and `openclaw-tools-*.js` genuinely reads `patch.modelOverrideSource`. But that internal handler sits behind a **strict RPC schema** that rejects unknown properties, so the field is not reachable over `sessions.patch` in this build. The handler and the public schema disagree.

### Options to evaluate next

1. Whether `sessions.create` accepts the field (the session is created before the first patch).
2. Stop pinning entirely and instead *clear* the session override so the agent default — which keeps its fallback chain — applies. This needs a supported "reset to default" instruction, which is exactly what jarvis#299 lacked.
3. Upstream: the field is in the handler but not the schema, which looks like an openclaw bug worth reporting.

Leaving this draft open as the record of the investigation. The branch must not be merged as-is.

PR #522, opened 2026-07-29
feat(chat): sweep the openclaw session model pins we wrote ourselves

kavin-114:
## The defect this heals

openclaw stores a per-session model override. From the 2026.6.8 bundle,
`resolveEffectiveModelFallbacks`:

```js
if (!params.hasSessionModelOverride) return agentFallbacksOverride;
if (!(params.modelOverrideSource === "auto" || ...)) return [];
```

Any pin that is not flagged automatic returns **zero** failover candidates for that
conversation, permanently. The bench pinned sessions on unpinned "Auto" turns too
(`turn_handler._session_model_for` names the pool primary so openclaw does not reject the
`jarvis-pool` placeholder), and every such patch landed as `modelOverrideSource: "user"`.

Fixing the turn path does not heal what is already pinned. The override lives in the
container's session store and outlives restarts, re-renders and re-provisions. This PR is
the cleanup, not the fix.

## Where it lives, and why

The customer app, as a bench maintenance command. Attribution is the whole problem and it
is only possible here:

- Inside the container a customer's pick and a bench-written pin are byte-identical, and
  the gateway does not put `modelOverrideSource` on the wire at all.
- On the bench, `Jarvis Conversation.model_override` records what the customer actually
  chose. Empty means Auto, so a pinned session whose conversation is on Auto is ours by
  construction.

A fleet-agent endpoint can see the pins but cannot attribute them, so it would have to
unpin blindly and would silently revert deliberate customer picks.

## Remedy per class

| class | verb | why |
|---|---|---|
| conversation on Auto | **clear** | we wrote it; the customer asked for the default and its chain |
| conversation pins the same model | **keep** | deliberate pick, and openclaw disabling failover for it is the product behaviour we want |
| conversation pins a different model | **report** | bench and container disagree; the next turn re-patches it |
| `agent:<id>:main[:heartbeat]` | **clear** | the built-in poller resumes it every tick, so a pin there fails forever |
| bench throwaway (title/prewarm/polish) | **skip** | single-use, and `session_lifecycle` reaps them on the `updatedAt` a patch would bump |
| anything else with no conversation | **report** | unattributable, never touched |

Rewriting `modelOverrideSource` to `"auto"` was considered and rejected. It leaves a dead
provider as the chain primary so every turn still starts with a guaranteed failure; it
fabricates the auto-failover provenance openclaw's primary-probe machinery reads; and
`sessions.patch` rejects the field outright (`additionalProperties: false`, verified
against the pinned image).

## RPC, not a file edit

Clearing goes through `sessions.patch {"key": k, "model": null}`, which takes openclaw's
own isDefault branch and deletes `modelOverride`, `providerOverride`, `modelOverrideSource`
and the stale runtime fields together, then persists through the gateway's store writer.
Editing `sessions.json` host-side would race the gateway, which keeps the store in memory
and rewrites it on every session update, and would otherwise need the container stopped.

## Honest limits

A `sessions.list` row carries the **resolved** model, never the raw override, so "resolves
away from the agent default" is the only available signal. It cannot separate a live
override from a stale runtime record left by an earlier render. Measured on
`jarvis-pool-bf4097`: 6 conversation sessions resolve away from the default, 3 of which
hold a real override. The clear is an idempotent assertion, so a false positive costs a
telemetry reset, never a behaviour change.

## Safety

- Dry run is the default; `apply=True` is explicit.
- `MAX_CLEAR` caps one run at 200.
- Never patches a key the gateway did not just report (patching an absent key makes
  openclaw mint an entry with a fresh sessionId and drop the label).
- Skips sessions with a run in flight and sessions with no `sessionId`.
- `model_override` is re-read immediately before each patch, so a pick made between the
  listing and the apply is not reverted.
- The patch echoes the resulting entry: a surviving override, or a different `sessionId`
  (the lifecycle sweep freed the session mid-run), counts as an error, not a fix.

## Dry run on the live tenant

```
$ bench --site jarvis.proxy execute jarvis.chat.session_pin_sweep.run

default: anthropic-0/claude-sonnet-5 | scanned: 32 | planned: 26
counts: {'clear': 6, 'keep': 0, 'report': 0, 'skip': 20}
cleared: 0  raced: 0  errors: 0

  clear  openai_compat-0/claude-sonnet-5  conv=nqhgkivuov  conversation is on Auto
  clear  openai_compat-0/jarvis-pool      conv=ltcolnru7t  conversation is on Auto
  clear  zai_coding-0/claude-sonnet-4-6   conv=48khjmi2t3  conversation is on Auto
  clear  vllm/claude-sonnet-4-6           conv=oaa18i0chm  conversation is on Auto
  clear  vllm/claude-sonnet-4-6           conv=14qig8pcn4  conversation is on Auto
  clear  vllm/claude-sonnet-4-6           conv=oo7tnui5kk  conversation is on Auto
```

The tenant's session store was byte-identical before and after; no container was mutated.

## Usage

```bash
bench --site <site> execute jarvis.chat.session_pin_sweep.run
bench --site <site> execute jarvis.chat.session_pin_sweep.run --kwargs "{'apply': True}"
```

## Related, and worth checking before it merges

Branch `fix-auto-model-override-source` sends `modelOverrideSource: "auto"` as a
`sessions.patch` param. On openclaw 2026.6.8 that field is not in
`SessionsPatchParamsSchema`, which is `additionalProperties: false`, so the compiled
validator rejects the whole call:

```
{"key":"...","model":"claude-sonnet-5","modelOverrideSource":"auto"}
  => false  at root: unexpected property 'modelOverrideSource'
```

The bundle line that branch quotes lives in `buildDirectChildSessionPatch`, the subagent
spawn path, not in the `sessions.patch` handler. If that branch ships as written, every
model patch becomes an `INVALID_REQUEST`.

### DIFF
--- a/jarvis/chat/openclaw_client.py
+++ b/jarvis/chat/openclaw_client.py
@@ -668,6 +668,55 @@ class OpenclawSession:
 		res = self._request("sessions.list", {}, timeout_s=timeout_s)
 		return (res.get("payload") or {}).get("sessions") or []
 
+	def list_sessions_page(
+		self,
+		*,
+		limit: int = 100,
+		offset: int = 0,
+		timeout_s: float = CONNECT_TIMEOUT_SECONDS,
+	) -> dict:
+		"""One page of ``sessions.list`` with the WHOLE payload, not just the
+		rows: ``sessions`` plus ``defaults`` (the agent's configured model, i.e.
+		what an unpinned session resolves to) and ``hasMore`` / ``nextOffset``.
+
+		``list_sessions`` above drops the envelope and accepts the gateway's own
+		100-row cap. The model-pin sweep needs both halves: ``defaults`` is the
+		only way to tell a pinned session from an unpinned one over the wire
+		(the row carries the RESOLVED model, never the raw override), and a
+		tenant with a long history needs the paging cursor."""
+		res = self._request(
+			"sessions.list",
+			{"limit": limit, "offset": offset},
+			timeout_s=timeout_s,
+		)
+		return res.get("payload") or {}
+
+	def clear_session_model(
+		self,
+		session_key: str,
+		*,
+		timeout_s: float = CONNECT_TIMEOUT_SECONDS,
+	) -> None:
+		"""Drop the session's model override so the agent's configured default -
+		and, crucially, its ``model.fallbacks`` chain - applies again.
+
+		``sessions.patch`` with a NULL model takes openclaw's own isDefault
+		branch (2026.6.8 ``src/sessions/model-overrides.ts``), which deletes
+		``modelOverride``, ``providerOverride`` and ``modelOverrideSource``
+		together with the now-stale runtime model/contextTokens fields, then
+		persists through the gateway's own store writer. Rewriting
+		``sessions.json`` on disk instead would race that writer: the gateway
+		holds the store in memory and rewrites it on every session update.
+
+		NEVER call this for a key the gateway did not just report. Patching an
+		ABSENT key does not fail - openclaw creates the entry, mints a fresh
+		sessionId and drops the label."""
+		self._request(
+			"sessions.patch",
+			{"key": session_key, "model": None},
+			timeout_s=timeout_s,
+		)
+
 	def delete_session(
 		self,
 		session_key: str,
--- a/jarvis/chat/openclaw_client.py
+++ b/jarvis/chat/openclaw_client.py
@@ -696,7 +696,7 @@ class OpenclawSession:
 		session_key: str,
 		*,
 		timeout_s: float = CONNECT_TIMEOUT_SECONDS,
-	) -> None:
+	) -> dict:
 		"""Drop the session's model override so the agent's configured default -
 		and, crucially, its ``model.fallbacks`` chain - applies again.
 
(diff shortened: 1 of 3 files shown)


kavin-114:
### Local CI — `Aerele-RnD/jarvis` branch sweep-poisoned-session-model-pins

Tested commit `06734417545e` · ✅ **PASS** · 19m21s

|---|---|---|---|
| `lint` | ✅ PASS | 4/4 | 44s |
| `tests-shard1` | ✅ PASS | 9/9 | 5m12s |
| `tests-shard2` | ✅ PASS | 9/9 | 4m40s |
| `tests-shard3` | ✅ PASS | 9/9 | 3m51s |
| `tests-shard4` | ✅ PASS | 9/9 | 4m47s |
| `coverage` | ✅ PASS | 5/5 | 4s |

---

kavin-114:
Follow-up commit `67508746` fixes the three `tests (3)` errors on run 30441470958.

Root cause: develop removed self-hosted deployment support in `9f8d984a`, which deletes `jarvis/selfhost.py`. This branch predates that, so the merge is textually clean (the PR reads MERGEABLE) but the merge RESULT that CI tests has no `jarvis.selfhost` to resolve. Every one of the three was the same `AttributeError: module 'jarvis' has no attribute 'selfhost'` raised by `mock._patch.__enter__`, not an assertion failure.

- `run()` drops `from jarvis import selfhost` and the `aborted="self-hosted"` early return. Jarvis is managed-only, so that branch is unreachable.
- `test_run_on_a_self_hosted_bench_does_nothing` is DELETED. It asserted an abort reason that no longer exists.
- `test_run_defaults_to_dry_run` and `test_run_without_agent_url_aborts` keep every assertion and lose only the dead `patch(...)`. Both already exercised the managed path, which is now the only path.

The "Skipped entirely on a self-hosted bench" bullet is removed from the description for the same reason.

Unrelated, and pre-existing on develop: `jarvis/error_push.py::push_error_rollup` still opens with `from jarvis import selfhost`. It is inside the function's own `try/except Exception`, so on develop that scheduler entry now raises `ModuleNotFoundError` on every tick and silently degrades to `_log_push_failure_throttled()`. Worth a separate issue; deliberately not touched here.

PR #528, opened 2026-07-29
fix(account): a provider label is not a connection

kavin-114:
## The bug

`_has_llm_config` asked whether a provider **name** was written down, not whether a usable **credential** existed:

```python
return bool(provider or model or settings.get("models") or proxy_active)
```

`llm_provider` / `llm_model` are labels that `on_update` mirrors from `models[0]`. They are not proof of anything on their own, and a partial disconnect leaves them behind.

`jarvis.oauth.api.disconnect` (Disconnect chat subscription) clears only the OAuth side, writes `last_sync_status = "disconnected"`, and **deliberately keeps** `llm_provider` / `llm_model`. A workspace in that state has no usable credential and cannot answer a turn, yet reported `disconnected: False` — so the Connection badge showed a healthy state over a dead workspace, while the sync status beside it said "disconnected".

## The fix

It now asks for a credential:

- **`models[]`** — the pool holds its own keys and accounts. Counted even when every row is DISABLED, because a paused pool still *has* credentials and must not read as disconnected.
- **`proxy_active`** — derived from config at save time and reset to 0 when the config goes away, so a set flag is itself proof a pool exists.
- **A direct tenant's stored api key, or its live OAuth connection.**

A password-store read that raises is treated as "no credential" rather than propagating: an unreadable secret is not evidence of a connection, and the status endpoint must not fail because of one.

## Correction worth recording

The tenant that prompted this turned out **not** to be exhibiting the bug. It showed `last_sync_status: "disconnected"` with `disconnected: False`, but it has a real stored key and an empty pool, which is a legitimate direct-tenant shape where `disconnected: False` is correct. Its status string was stale.

The predicate flaw was real regardless, and is now pinned by tests.

## Why the tests use a stub

They are pure unit tests over a stub rather than the live Single, and that is deliberate. The first version drove the real `Jarvis Settings` and was wrong twice over:

1. It fought whatever another session happened to have configured on the shared test site. A model appeared mid-run and flipped the result.
2. Its setup had to delete a real `__Auth` row, because blanking a Password **column** leaves the secret in `__Auth`. That leaked a test key into every later test on the site and failed them out of order.

The predicate is pure logic over a settings-like object, so it is tested that way.

## Verification

- `test_llm_monitor`: 12 tests (7 new), green
- `test_disconnect_llm`, `test_role_gates`: green
- `test_account`'s single failure is **pre-existing**, confirmed by running it against the untouched checkout on the same database
- `ruff` check and format clean (pinned 0.14.10)

https://claude.ai/code/session_01WLXuBvzxvSt2WqHQRXdRBz

### DIFF
--- a/jarvis/account.py
+++ b/jarvis/account.py
@@ -331,19 +331,43 @@ def get_llm_connection_status() -> dict:
 
 
 def _has_llm_config(settings) -> bool:
-	"""True when this workspace still has an AI connection configured.
+	"""True when this workspace still has a usable AI CREDENTIAL.
 
-	The flat llm_provider / llm_model pair is the mirror on_update keeps of
-	models[0], so it covers pool and direct tenants alike. models[] is checked
-	too because a table whose rows are all DISABLED leaves the mirror blank while
-	the credentials are still stored - that is a paused pool, not a disconnected
-	workspace, and it must not read as one. proxy_active is checked for the same
-	reason from the other direction: it is DERIVED from the config at save time
-	(compute_proxy_active) and reset to 0 whenever the config goes away, so a set
-	flag is by itself proof that a pool exists, whatever the mirrors say."""
-	provider = (settings.get("llm_provider") or "").strip()
-	model = (settings.get("llm_model") or "").strip()
-	return bool(provider or model or settings.get("models") or getattr(settings, "proxy_active", 0))
+	It asks whether a credential exists, not whether a provider NAME is written
+	down, and that distinction is the whole point. llm_provider / llm_model are
+	labels that on_update mirrors from models[0]; they are not proof of anything
+	on their own, and a partial disconnect can leave them behind.
+
+	The concrete case: jarvis.oauth.api.disconnect (Disconnect chat subscription)
+	deliberately clears only the OAuth side, writes last_sync_status
+	"disconnected", and LEAVES llm_provider / llm_model in place. Testing the
+	labels therefore reported such a workspace as connected while its own sync
+	status said "disconnected" - two sources of truth disagreeing, and the
+	customer's Connection badge showing a healthy state over a workspace that
+	cannot answer a turn. Observed on a real tenant: models[] empty,
+	last_sync_status "disconnected", llm_provider still "OpenAI".
+
+	What counts as a credential:
+	  * models[] - the pool holds its own keys/accounts. Checked even when every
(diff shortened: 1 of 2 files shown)


kavin-114:
## Correction to this PR's own Verification section

The body above claims `test_llm_monitor` was "12 tests (7 new), green" and that the only failure was a pre-existing one in `test_account`. **Both claims were wrong.** CI was failing on:

```
FAIL test_llm_monitor.TestGetLlmConnectionStatus
     .test_direct_tenant_short_circuits_without_admin_call
AssertionError: '' != 'gpt-4o'
```

A pre-existing test in the very module this branch edits, not `test_account`.

### Why it failed

That test sets up a workspace holding only a label:

```python
frappe.db.set_single_value("Jarvis Settings", "proxy_active", 0)
frappe.db.set_single_value("Jarvis Settings", "llm_model", "gpt-4o")
...
self.assertEqual(out["default_model"], "gpt-4o")
```

No api key, no `models[]` rows. Under the old lenient `_has_llm_config` a bare label counted as a connection, so it passed. That is precisely the behaviour this branch changes, so **the test was asserting the bug** and I never updated it.

### Why I wrongly believed it was green

It passes on my local `site.jarvis`, which holds a real stored credential and a populated `models[]` -- either satisfies the new predicate regardless of the label. CI runs a fresh DB with neither. That is the CLAUDE.md rule I broke: *CI is the authority, never a green local bench-test.*

### The fix (`ce8c879f`)

`_has_llm_config` is now stubbed for that test rather than satisfied with a real key. The subject there is the short-circuit and the field remap; the predicate itself is covered exhaustively over a stub in `TestHasLlmConfig`. Writing a real Password field would leave the secret in `__Auth` when blanked, leaking a test key into every later test on the site -- the same hazard that drove these tests to stubs originally -- and would make the outcome depend on whatever `models[]` the shared test site happens to hold.

Also adds `test_a_leftover_model_label_alone_reports_disconnected`, pinning the new behaviour at the **endpoint** rather than only on the predicate: a workspace with a leftover mirror and no credential reports `disconnected: True` and must not leak the label out as a `default_model` the SPA would render as a healthy "Direct" state. That is the shape `jarvis.oauth.api.disconnect` leaves behind, and nothing covered it there.

### Now green

`lint` pass, `tests (1)` `(2)` `(3)` `(4)` all pass, `coverage` pass, on run 30536410290. Real hosted Actions, not a local run.

PR #540, opened 2026-07-30
feat(chat): GAP 3 — self-heal a wedged relay pump (force-takeover, flag-off)

venkat102:
## What

**Track A of the pump-recovery hardening (GAP 3): self-heal a wedged pump.** A relay pump can keep renewing its lease while no longer *draining* turns (e.g. a stuck socket/lane, a livelocked reactor). Because the lease looks live, `ensure_pump` refuses to revive it, so its turns queue forever. This adds a watchdog-driven, epoch-fenced **force-takeover** of such a pump.

Shipped **flag-off** (`jarvis_pump_force_takeover` absent/falsy = the A4a *detect-and-log* phase): the watchdog detects + records wedges but does **not** force-expire until the flag is enabled (A4b), after the false-positive rate is measured in prod.

## How (tasks A1–A5)

- **A1** — `lease_acquire` seeds `loop_heartbeat_ts` on a cold start but **carries it forward** on a clean-hop re-acquire (`COALESCE`), so the stamp tracks *progress*, not hop churn.
- **A2** — `loop_heartbeat_ts` is now a **progress stamp**: `drain_slice` advances it only when the slice did real drain work (promoted/dispatched a turn, applied a mux frame, resolved an ack/recovery) or was correctly capacity-blocked. The lease is still renewed every slice (liveness), decoupled from progress.
- **A3** — `force_expire_wedged`: one epoch-bumping CAS that fences the wedged pump (its cached-epoch renews/heartbeats/turn-writes then affect 0 rows → `lease_lost_exit`), vacates the lease, re-stamps in-flight turns, and resets the progress stamp so the successor gets a fresh window.
- **A4** — the watchdog detects a lease-live-but-progress-stale pump (`PROGRESS_STALE_S = 2×SOFT_HOP_BUDGET_S = 180s`); logs (throttled) + emits `pump_wedged`/`pump_force_expired` telemetry; force-expires + revives **only when the flag is on**. Fires at most once per shard per watchdog tick.
- **A5** — the `jarvis_pump_force_takeover` site_config rollback rung (default OFF).

No schema change — reuses the existing `loop_heartbeat_ts` field.

## Review

Ran through the full `/review-loop` (T3). **Cycle 1 caught a CRITICAL** worth calling out: `run_pump_hop` calls `lease_acquire` on *every* hop, so an early design that stamped `loop_heartbeat_ts=now` on acquire refreshed the progress stamp ~every 90s and the 180s wedge trigger **could never fire** — a dead feature whose tests validated an unreachable row. Fixed with the `COALESCE` carry-forward (A1 above) and regression-guarded by `test_a1_acquire_carries_progress_stamp_forward` (mutation-verified: the original overwrite reds it). The fix wave also wired per-event telemetry (the throttled log under-counted the A4a signal) and counted `_poll_pending` resolutions as progress. **Cycle 2: clean** — the wedge is now reachable, no false-kill of a healthy pump in normal operation, and no new defect from the fix.

**Deferred (documented, non-blocking, flag-off contained):** a freshly-*revived* pump inherits a stale stamp → a slightly elevated trip rate on revived shards (the A4a measurement expects this); a genuinely-silent stream >180s is an inherent wedge-detection false-positive to be tuned in A4a; the `force_expire_wedged`/`lease_acquire` re-stamp duplication could share a helper. None mutate state while the flag is OFF.

## Tests

All pump-related modules green on `jarvis-test.localhost`: `test_turn_state` (61), `test_pump` (62), `test_pump_pipeline` (37), `test_turn_recovery` (35), `test_chat_admission` (59), `test_pump_fence_shared_client` (3). Two guards mutation-verified (the carry-forward CRITICAL guard; the wedge-staleness gate).

## Rollout

Merge with the flag OFF (A4a). After the detect-log telemetry shows an acceptable false-positive rate, enable `jarvis_pump_force_takeover` (A4b). Rollback = flip the flag. Tracks B/C (the GAP 1 scheduler dead-man's-switch, control plane) and D (ops) follow in separate PRs.

### DIFF
--- a/jarvis/chat/pump.py
+++ b/jarvis/chat/pump.py
@@ -1559,12 +1559,11 @@ def drain_slice(ctx: PumpContext) -> str:
 	# regardless — the snapshot poll never blocks and mux.dispatch runs below.
 	_maybe_refresh_capacity(ctx)
 	_poll_snapshot(ctx)  # CDX-11: fold a resolved refresh (or fail-close a timed-out one) pre-promote
-	_promote_queued(ctx)
-	_dispatch_ready(ctx)  # OARF-5: issues chat.send acks, parks them (never blocks)
+	promoted = _promote_queued(ctx)
+	dispatched = _dispatch_ready(ctx)  # OARF-5: issues chat.send acks, parks them (never blocks)
 	_poll_pending(ctx)  # OARF-5: resolve done acks/recovery tails; ack-timeout deadlines
 
-	if ctx.mux is not None:
-		ctx.mux.dispatch(block_s=SLICE_BLOCK_S)
+	applied = ctx.mux.dispatch(block_s=SLICE_BLOCK_S) if ctx.mux is not None else 0
 	if ctx.lease_lost:
 		ts.lease_lost_exit(ctx.lease_lost)
 	# D5 §5-d: a dead socket ends the hop (the reader's Closing already failed the
@@ -1575,7 +1574,12 @@ def drain_slice(ctx: PumpContext) -> str:
 
 	_cancel_sweep(ctx)
 
-	_heartbeat_and_renew(ctx)
+	# GAP 3 A2: progress = the slice did real drain work (promoted / dispatched a turn
+	# or applied a mux frame) OR the pump is correctly capacity-blocked (fail-closed:
+	# gateway visibility UNKNOWN, where a takeover cannot help — not a wedge). Only a
+	# genuine "should be draining but isn't" leaves the progress stamp un-advanced.
+	progressed = promoted > 0 or dispatched > 0 or applied > 0 or not ctx.gateway_active_known
+	_heartbeat_and_renew(ctx, progressed)
 
 	if _idle_exit(ctx):
 		return "idle_exit"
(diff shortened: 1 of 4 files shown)


Vigneshsekar:
Review fix pushed (b75d86af): **the takeover CAS now re-asserts the full wedge predicate**, not just the epoch.

**The race:** `_pump_wedged` reads `loop_heartbeat_ts` staleness, then `force_expire_wedged` CASed on `pump_epoch=E` alone. A pump that resumed progress between that read and the UPDATE (still at epoch E) would be force-expired while healthy — the false-positive this flag's rollout is supposed to measure, baked into the mechanism itself.

**The fix:** `force_expire_wedged(target, epoch, stale_after_s)` folds `AND lease_expires_at > now AND loop_heartbeat_ts < now - stale_after_s` into the UPDATE's WHERE. A pump that heartbeated since the read, a lease that expired meanwhile (the ordinary `ensure_pump` revive path owns that shape), and a NULL stamp (SQL `NULL < x` never matches — same NULL-is-fresh guard as `_progress_stale`) all affect 0 rows. The watchdog passes `PROGRESS_STALE_S`.

**Proof:** the new regression test `test_a4_force_expire_reasserts_progress_under_cas` (heartbeat lands between detection and takeover → expect 0 rows) was replayed against the epoch-only CAS and fails there exactly (`True is not false`), passes with the fix. Also added `test_a4_force_expire_requires_live_lease`.

Suites on patterntest: turn_state 63 OK, pump 62 OK, pump_pipeline 37 OK, chat_admission 11 OK. ruff 0.14.10 check + format --check clean on the three changed files (hunk-confined).

Issue #535, opened 2026-07-29
prewarm/title/polish: sessions.list cannot distinguish 'run accepted, not started' from 'finished', so a reclaim can delete a run about to begin


kavin-114:
Follow-up to #530, which fixed this for the chat turn path but not for the page-load / prewarm path.

## What still happens

Captured on `jarvis-pool-bf4097` during the clean Phase B run on 2026-07-29, 41 seconds before the chat turn and on a **different runId**:

```
16:12:21.569  res ✓ sessions.create
16:12:21.654  res ✓ agent runId=e95590f74abd4e9a9b118324c2089ef3
16:12:21.954  res ✓ sessions.delete          <-- session deleted 300ms into a live run
16:12:23.776  embedded run agent end: runId=e95590f7... isError=true
16:12:23.812  failover decision: decision=surface_error reason=rate_limit
16:12:23.818  model fallback decision: decision=candidate_failed ... reason=rate_limit next=none
```

Same signature #530 fixed for turns: the session is deleted while its run is still in flight, so the orphaned run cannot resolve agent fallbacks and produces `surface_error` instead of `fallback_model`.

Confirmed #520, #522 and #530 are all merged and deployed on `develop` at the time of capture (`8e5f30d7`, `a76884fa`, `6e3794f2`). The chat turn in the same window was clean: exactly one agent run, no `sessions.create`/`sessions.delete`, no `EmbeddedAttemptSessionTakeoverError`. So #530 works, it just did not reach this caller.

## Why it matters

Identical reasoning to #525 and #531. A `next=none` from this path is **indistinguishable in the log** from a genuinely dead failover chain. That ambiguity is what produced five wrong root causes across 2026-07-28/29 and blocked the sidecar-retirement decision for two days.

The clean Phase B run met every PASS clause except "no spurious `next=none` before it", solely because of this. It is the last known source of that noise.

## Suspected scope

The triple fires on page load, which points at prefix prewarm (`jarvis/chat/prewarm.py`). #530 introduced `session_lifecycle.reclaim_throwaway_session`, which probes `sessions.list -> hasActiveRun` before deleting, and wired three call sites onto it. Either this path was missed, or it takes a different route to the delete. Worth checking whether anything else creates and tears down a session on page load.

## Note on the related overlap

[...]

Issue #531, opened 2026-07-29
llm: throwaway one-shots pin an explicit model, so they can never fail over and still emit a misleading next=none


kavin-114:
Follow-up to #525. That issue removed the *noise*; this is the remaining *ambiguity*.

## The problem

The throwaway one-shot gateway turns (auto-title `jarvis/chat/title.py`, pattern-polish `jarvis/learning/polish.py`, prefix prewarm `jarvis/chat/prewarm.py`) each pass an explicit `model` / `provider` on the `agent` RPC.

openclaw's `resolveEffectiveModelFallbacks` returns `[]` for any session model override not marked `modelOverrideSource: "auto"`, and that marker is **not in the `sessions.patch` schema** (the gateway rejects it with `INVALID_REQUEST`, see the closed PR #517). So an explicitly pinned one-shot is **structurally unable to fail over**.

Consequence: when a one-shot hits a genuine upstream failure it prints

```
decision=surface_error ... next=none
```

which is byte-identical to what a genuinely dead failover chain prints. Observed on two real prewarm runs that hit a Gemini `429 quota exceeded`.

## Why this matters out of proportion to its size

`next=none` has had **three separate causes** stacked on top of each other:

1. a `source:"user"` session pin on the main chat path — fixed by #520
2. a spurious first `agent` run from a session deleted mid-flight — fixed by #530
3. **this one, still open**

Each was mistaken for "openclaw native failover is broken". Across 2026-07-28/29 that produced five wrong root causes in a row, including a confident and completely incorrect conclusion that the `google-generative-ai` adapter could not fail over, which was later falsified by re-running openclaw's own resolver. The sidecar-retirement decision was blocked on it for two days.

**Until this is closed, a `next=none` in the logs cannot be trusted to mean anything**, and any future failover measurement stays ambiguous.

## Why it was left out of #530

Removing the override would regress the deliberate model pinning these features rely on: prewarm intentionally warms a specific model's prefix cache, and auto-title intentionally uses a cheap model rather than the pool primary. The fix is not simply "stop pinning".

## Options to evaluate

1. Whether openclaw offers a per-request way to pin a model that still permits fallbacks (the `sessions.patch` route is closed, but the `agent` RPC may accept something else).
2. Whether these one-shots should tolerate failover and simply not pin, accepting the cost.
[...]

PR #555, opened 2026-07-30
fix(chat): stop prefix prewarm from burning the tenant's own LLM quota

### LINKED ISSUE #548: prewarm: fires on login, every chat page load AND every 5 minutes, burning a paid/quota LLM request each time
## What happens

Prefix prewarm (`jarvis/chat/prewarm.py`) fires an `agent` run on **login, every chat page load, and every 5 minutes**. Each fire is a real upstream LLM request against the tenant's own credentials.

## Why it was noticed

It exhausted the free-tier Google quota on `jarvis.proxy` during 2026-07-29 testing. The resulting 429s were being read as test signal:

```
generate_content_free_tier_input_token_count, limit: 0, model: gemini-2.5-pro
```

Two knock-on effects, both already investigated at some cost:

- The 429s fed a `surface_error` on a pinned one-shot, which is part of what made `next=none` ambiguous (#531).
- #535 was filed on a misread of one of these prewarm runs. Its stated premise turned out to be false, though the investigation did surface a genuine defect.

So the quota exhaustion was not a harmless side effect. It actively muddied two other investigations.

## Why it matters in production

This is the customer's own API key and their own quota or bill. A tenant that opens the chat page repeatedly, or leaves a tab open, pays for prewarm requests that produce no user-visible output. On a free-tier or low-quota key it can **exhaust the quota that real chat turns need**, which is the failure mode observed here: prewarm consumed the budget and actual turns then 429'd.

The cost also scales with the wrong thing. Page loads and a 5-minute timer are not correlated with whether anyone is about to send a message.

## Worth checking

1. Whether all three triggers are needed. A 5-minute timer plus every page load looks like belt-and-braces for a cache that may not need either.
2. Whether the cooldown is effective. Two warms were observed passing the get-then-set cooldown check **6ms apart**, so the guard is racy and lets duplicates through.
3. Whether prewarm should be skipped entirely when the tenant's pool is on a metered or free-tier key, or made opt-in.
4. Whether the benefit is measured at all. If nobody has quantified the latency saved by a warm prefix cache, the cost side is currently unpriced.

## Related

- #531 pinned one-shots cannot fail over (prewarm is one of the three)
- #535 the reclaim window measured while investigating these runs
- #545 the fix for that window

kavin-114:
Fixes #548.

## The defect

`jarvis/chat/prewarm.py` fired a real upstream `agent` run on **three** triggers: login (`on_session_creation`), every chat-surface load (`list_conversations`), and a `*/5` cron. Each is billed to the tenant's own credential and produces nothing the user sees. It exhausted a free-tier Gemini key on `jarvis.proxy`, and the resulting 429s then landed on real chat turns.

## Why the volume was never defensible

A warm can never be cost-neutral, only cost-traded-for-latency:

| | first turn | later turns |
|---|---|---|
| no prewarm | full prefix | provider's cached rate |
| prewarm | cached rate | cached rate (plus the warm's own full prefix) |

So prewarming is always strictly more spend, and a warm that no turn follows is a wholly wasted prefix. Measured from `logs/jarvis.chat.latency.log*` on `jarvis.proxy`: **528 warms against 86 real turns**, six paid warm-ups per turn.

That means the only defensible trigger is one correlated with a user about to send a first message, firing at most once per arrival.

## Triggers kept and removed

**Removed: the `*/5` `keep_warm_if_active` cron.** Its gate ("any chat message in the last 30 minutes") was **inverted with respect to its own purpose**. Recent chat traffic is exactly what keeps the provider cache warm for free, so the cron spent up to seven paid warms per burst of activity precisely when warming was worthless, and no-opped on the idle benches where a cold prefix is actually possible. A timer cannot know a turn is coming, which is the premise of warming, so it has no correct interval, not a longer one.

**Removed: the `on_session_creation` login warm.** It billed a request for every login to the site, including desk and API logins by users who never open chat, to buy a sub-second head start on a multi-second prefill that the chat-surface load already begins.

**Kept: the chat-surface load.** The only trigger correlated with someone about to send a message.

## Correctness and volume on the trigger that stays

- **The cooldown race is fixed.** The get-then-set spanned a `Jarvis Settings` load (a DB round trip), which is the window two warms were observed passing 6ms apart on `jarvis-pool-bf4097`: a duplicate billed run, and the trigger for the #535 reclaim race. It is now a single `SET NX EX`, claimed before any slow work. It raises on a Redis outage, and callers turn that into "no warm" deliberately: failing closed is the right side for work that spends the customer's money.
- **A warm is skipped while the prefix is already hot** (a real turn within the last two minutes). Sixteen SPA call sites reach `list_conversations`, several of them after every send, so without this the remaining trigger still warms into sessions whose own turns keep the prefix warm.
- **`warm_session`** (the whitelisted endpoint) now goes through `enqueue_warm_if_due` instead of enqueuing unconditionally, so N calls can no longer become N short-queue jobs.
- **`jarvis_prefix_prewarm: 0`** in site_config is an operator kill switch (absence reads as on, same shape as `jarvis_pump_enabled`).

## On metered keys and opt-in

There is no mode in which the operator pays: `llm_auth_mode` is api_key / subscription / oauth and all three are the customer's credential. So "skip for metered keys" would mean "skip always", and a per-tenant consent toggle would be asking every customer to answer a question nobody has data for. The honest position is the one taken here: volume proportional to real use, plus an operator off switch. Revisit per-tenant opt-in once a benefit number exists.

## Is the benefit measured? No.

Recorded in the module docstring rather than implied away:

- The turn telemetry line records no warm/cold state and no cached-token count, so `first_delta_ms` cannot be split by prefix state **even in principle**.
- `fire_ms` times the WS round trip, not the LLM work, so it says nothing about what a warm bought.
- The Stage-A/B harnesses run against a `FakeGateway`, and their `warm_session` pre-creates an openclaw **session**, not a provider prefix cache.
- "Watch warm-turn token spend after this change" (commit `1017f3ab`, which tripled the frequency) was never done. There is no token-spend telemetry for warms at all.

Skips are now logged alongside fires on the latency logger, so the paid-warm rate is readable off one file.

One caveat on the retention numbers the 4-minute cooldown is justified against: ~5-10 min idle eviction is documented for OpenAI and 5 min is Anthropic's default ephemeral TTL, but Gemini's **implicit** cache has no documented TTL, and Gemini is the provider that actually blew the quota. The comment says so rather than presenting it as measured.

## Not regressed

The pin stays pinned (#531, #544): prewarm still passes an explicit model so it warms THAT model's cache, and still mints a `jarvis-pinned-prewarm:` run id. #545's `fired_at` plus 5.0s grace on `reclaim_throwaway_session` is untouched.

## Tests

Fail without the change:

- `test_two_concurrent_warms_fire_exactly_one` pins the race. It re-enters `warm_prefix` from the `Jarvis Settings` load, which is precisely the window the old get-then-set left open, so under the old code both callers reach `fire_agent`. Threads are not usable here because `frappe.local` is thread-local.
- `TestPrefixAlreadyWarm` covers the hot-prefix skip, that it arms the full cooldown rather than the short marker, and the exact window `_RECENT_TURN_S` means.
- `TestPrewarmKillSwitch` covers absent-means-on and every explicit falsy value.
- `TestPrewarmTriggerSurface` asserts on `hooks.py` so neither removed trigger can be quietly wired back up, and that the two entry points are deleted rather than merely unhooked.
- `test_warm_session_respects_the_cooldown` covers the endpoint.

https://claude.ai/code/session_015kFVKdpkN6uMTgZDCm2LJ5

### DIFF
--- a/jarvis/chat/api.py
+++ b/jarvis/chat/api.py
@@ -126,9 +126,13 @@ def list_conversations() -> list[dict]:
 	reuses the hidden empty row.
 	"""
 	require_jarvis_access()
-	# Chat page loaded: warm the openclaw prefix cache in the background
+	# Chat surface loaded: warm the openclaw prefix cache in the background
 	# (best-effort, debounced) so the first turn of a new chat skips the cold
-	# provider prefill. Never blocks or fails this read.
+	# provider prefill. Never blocks or fails this read. Since #548 this is the
+	# ONLY prewarm trigger, and it is deliberately a load-correlated one: a warm
+	# is a billed upstream request against the tenant's own quota. The several
+	# call sites that reload this list AFTER a send are absorbed by prewarm's own
+	# "the prefix is already hot" guard, not by anything here.
 	from jarvis.chat import prewarm
 
 	prewarm.enqueue_warm_if_due()
@@ -1661,14 +1665,19 @@ def set_conversation_model(conversation: str, model: str | None = None) -> dict:
 @frappe.whitelist()
 def warm_session() -> dict:
 	"""Fire-and-forget: warm this tenant's openclaw prefix cache so the next
-	new-chat first turn skips the cold prefill. Best-effort; always ok. The
-	chat UI calls this on open. Unconfigured benches no-op.
-	Runs in a background RQ job so the gunicorn web worker is not blocked."""
+	new-chat first turn skips the cold prefill. Best-effort; always ok.
+	Unconfigured benches no-op. Runs in a background RQ job so the gunicorn web
+	worker is not blocked.
+
+	Goes through ``enqueue_warm_if_due`` rather than enqueuing ``warm_prefix``
+	directly (#548). A warm is a billed upstream request against the tenant's own
+	quota, and enqueuing unconditionally let any authenticated Jarvis user turn N
+	calls to this endpoint into N short-queue jobs. The cooldown claim inside
+	``warm_prefix`` bounds the SPEND either way; this bounds the jobs too."""
 	require_jarvis_access()
-	frappe.enqueue(
-		"jarvis.chat.prewarm.warm_prefix",
-		queue="short",
-	)
+	from jarvis.chat import prewarm
+
+	prewarm.enqueue_warm_if_due()
 	return {"ok": True, "enqueued": True}
 
 
--- a/jarvis/chat/prewarm.py
+++ b/jarvis/chat/prewarm.py
@@ -1,12 +1,68 @@
-"""Best-effort pre-warming of the openclaw container's OpenAI prefix cache.
+"""Best-effort pre-warming of the openclaw container's provider prefix cache.
 
-The first turn of a fresh session pays a cold provider prefill over the
-large static system prefix (persona + skills + tool schema). That cache is
-prefix-keyed and container-wide, so one cheap throwaway agent turn warms it
-for every subsequent new chat in the same container. We never touch the
-user's real session or write chat rows. Always best-effort; never raises.
+The first turn of a fresh session pays a cold provider prefill over the large
+static system prefix (persona + skills + tool schema). That cache is
+prefix-keyed and container-wide, so one cheap throwaway agent turn warms it for
+every subsequent new chat in the same container. We never touch the user's real
+session or write chat rows. Always best-effort; never raises.
+
+WHAT A WARM COSTS (issue #548)
+------------------------------
+A warm is a real upstream ``agent`` run billed to the TENANT'S OWN credential.
+There is no mode in which the operator pays: ``llm_auth_mode`` is api_key /
+subscription / oauth and all three are the customer's. So a warm can never be
+cost-neutral, only cost-traded-for-latency:
+
+  without prewarm  the first turn pays a full prefix, later turns pay the
+                   provider's discounted cached rate
+  with prewarm     the warm pays a full prefix, and then EVERY turn pays the
+                   cached rate
+
+which is strictly more spend per cycle when a turn does follow, and one wholly
+wasted prefix when none does. This app fired 528 warms against 86 real turns on
+jarvis.proxy (six paid warm-ups per turn) and exhausted a free-tier Gemini key
+that real chat turns then 429'd on.
+
+So the only defensible trigger is one correlated with a user about to send a
+first message into a plausibly-cold container, firing at most once per arrival.
+Hence, as of #548:
+
+- ONE trigger: the chat-surface load (``enqueue_warm_if_due``, called from
+  ``list_conversations``). Two were removed.
+  * The ``*/5`` ``keep_warm_if_active`` cron. Its gate ("any chat message in
+    the last 30 minutes") was INVERTED with respect to its own purpose: recent
+    chat traffic is exactly what keeps the provider cache warm for free, so the
+    cron spent up to seven paid warms per burst of activity precisely when
+    warming was worthless, and no-opped on the idle benches where a cold prefix
+    is actually possible. A timer also cannot know a turn is coming, which is
+    the whole premise of warming.
+  * The ``on_session_creation`` login warm. It billed a request for every login
+    to the site, including desk and API logins by users who never open chat, to
+    buy a sub-second head start on a multi-second prefill that the chat-surface
+    load already starts.
+- A warm is SKIPPED while the prefix is already hot (``_prefix_recently_used``).
+  That is what stops the SPA's post-send list refreshes from billing warms into
+  a session whose own turns keep the prefix warm: sixteen call sites reach
+  ``list_conversations``, several of them after every single send.
+- The cooldown is claimed ATOMICALLY (``_claim_warm_slot``). The get-then-set it
+  replaces spanned a ``Jarvis Settings`` load, and two warms were observed
+  passing it 6ms apart on jarvis-pool-bf4097.
+
+NOT MEASURED
+------------
+No benefit number exists for any of this, and #548 is the right place to say so
+rather than to keep implying one. The turn telemetry line records no warm/cold
+state and no cached-token count, so ``first_delta_ms`` cannot be split by prefix
+state even in principle; the Stage-A/B harnesses run against a FakeGateway, so
+their ``warm_session`` measures nothing about provider caching; and "watch
+warm-turn token spend after this change" (commit 1017f3ab, which tripled the
+frequency) was never done - ``fire_ms`` below times the WS round trip, not the
+LLM work. Until a warm-vs-cold ``first_delta_ms`` split exists, treat the one
+remaining trigger as unproven and keep its volume proportional to real use.
+``jarvis_prefix_prewarm: 0`` in site_config turns it off entirely.
 """
 
+import pickle
 import time
 import uuid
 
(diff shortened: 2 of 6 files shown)


kavin-114:
## Verification

**Hosted CI on `9d58e37a`: all 6 checks green.** `lint`, `tests (1)`, `tests (2)`, `tests (3)`, `tests (4)`, `coverage`. Run: https://github.com/Aerele-RnD/jarvis/actions/runs/30545740941

The eleven new and modified tests all ran (shard 3) and all passed:

```
✔ test_two_concurrent_warms_fire_exactly_one
✔ test_recent_turn_skips_the_warm_and_arms_the_full_cooldown
✔ test_the_recent_turn_window_is_the_documented_one
✔ test_no_recent_turn_means_the_prefix_may_be_cold
✔ test_absent_flag_leaves_prewarm_on
✔ test_explicit_falsy_values_switch_prewarm_off
✔ test_disabled_site_neither_enqueues_nor_warms
✔ test_no_prewarm_on_login
✔ test_no_periodic_keep_warm_cron
✔ test_the_removed_entry_points_are_gone
✔ test_warm_prefix_failure_does_not_arm_full_cooldown
```

### The race test was proved load-bearing, not just asserted to be

A green suite does not show that a test would have caught the bug. So the pre-#548 get-then-set cooldown was restored on a throwaway branch with the tests unchanged, and CI dispatched against it. `tests (3)` went **red on exactly the two tests that pin this fix**, everything else stayed green:

```
✖ test_two_concurrent_warms_fire_exactly_one
✖ test_warm_prefix_failure_does_not_arm_full_cooldown

AssertionError: True is not False : the concurrent second warm must be refused
```

That assertion message is the defect itself: under get-then-set the second concurrent caller reached `fire_agent`, so the race billed a duplicate upstream run. The throwaway branch has been deleted.

### Review

Two independent review passes (correctness/concurrency, and test-validity/leftovers) found no HIGH or MEDIUM defects. Specifically confirmed:

- `RedisWrapper` inherits `set` and `ttl` raw from redis-py but overrides `exists` to apply `make_key` itself, so the hand-applied prefix in `_claim_warm_slot` and the raw key in `_warm_slot_held` both resolve to the same Redis key. No double-prefix, no missing prefix.
- `pickle.loads` is protocol-agnostic, so the claim's marker stays interoperable with `get_value` / `set_value` / `delete_value` on that key. Nothing reads the cooldown key through `get_value` any more, so bypassing `set_value` cannot leave `frappe.local.cache` stale.
- `Jarvis Chat Message` is a plain non-child DocType, so it carries frappe's default `index creation(creation)`. The new EXISTS is indexed, `add_to_date(..., seconds=-120)` returns a datetime, and `creation` shares `now_datetime()`'s tz convention.
- Every TTL exit path is finite, so no path can wedge warming permanently off, and the key is never left unset while claimed.
- #545 (`fired_at` + 5.0s grace) and #531/#544 (explicit model/provider, `jarvis-pinned-prewarm:` run id) are byte-identical to develop.
- The kill switch's off-value set and absent-means-on semantics match `_pump_flag_explicit_off` exactly.

One LOW finding was raised and fixed in `391cd133`: the comment above the claim said the full cooldown is armed whenever there was nothing to do, but the unconfigured-bench exit deliberately keeps the short 90s claim. The comment now states which two exits extend the TTL and why the skipping exits should not, since they spend nothing and a bench that becomes configured mid-onboarding should start warming in 90s rather than four minutes.

Note: this PR is still a draft and has not been run against a live tenant. Prewarm sits on the chat page-load path of a tenant whose quota is currently exhausted, and two other sessions are working the same bench, so live verification is left to the reviewer.

kavin-114:
## Adversarial review: MERGE-READY

Re-verified the load-bearing claims against the live bench, the frappe source, and the site logs rather than the diff alone. Verdict: merge-ready. Findings ranked below; nothing blocks.

### Verified (re-run, not taken from the PR body)

- **Volume claim is real.** `sites/jarvis.proxy/logs/jarvis.chat.latency.log*` counts **532 `warm_prefix fire_ms` lines against 86 turn lines** right now (528/86 at time of writing; it has grown since, warms were still firing today at 17:22, 18:16, 18:33, 19:10, 19:27). The 6:1 ratio is not rhetoric.
- **The atomic claim is sound against frappe's actual RedisWrapper.** `exists()` applies `make_key` itself (redis_wrapper.py:221), `set_value` pickles (matching `_CLAIM_VALUE = pickle.dumps("1")`), and the raw `cache.set(make_key(key), ..., nx=True)` is the one true SET NX EX. The docstring's claims about which wrappers prefix the key are all accurate against the frappe checkout on this bench.
- **The kill switch gates the spend.** `_prewarm_enabled()` is checked in `enqueue_warm_if_due` AND at the top of `warm_prefix`, and a repo-wide grep confirms `warm_prefix` is the only path to `fire_agent`: `list_conversations` and `warm_session` both go through `enqueue_warm_if_due`. No other trigger survives.
- **Fail-closed has no silent-permanent failure mode.** A Redis outage that breaks the claim also breaks the RQ enqueue that would run the warm, so warming could not have happened anyway; the exception path logs at WARNING (`prefix warm-up failed`, with traceback), and nothing persists an off state, so it self-recovers when Redis does. The only degraded mode I found: a Redis blip exactly at the post-warm cooldown extension (set_value suppresses ConnectionError) shortens one cooldown from 240s to 90s. One extra warm per blip, negligible.
- **`_prefix_recently_used` is indexed as claimed.** `SHOW INDEX` on `tabJarvis Chat Message` (jarvis.proxy) confirms a `creation` index.
- **Hosted CI 6/6 green on the exact head sha** `391cd133`, which per project rules is the authority for this repo.
- **The removed entry points are gone**, the trigger-surface tests pin hooks.py, and the scheduler cron block at head is structurally intact.

### Finding 1 (accuracy of the PR's own argument): the "INVERTED gate" claim is overstated

The deleted cron's documented purpose was: "so a returning user's first turn is warm after an idle gap." Provider retention is roughly 5 to 10 minutes; the gate covered 30. So in the band from cache-eviction to 30 minutes after the last message, the cron was doing exactly what it was designed to do: re-warming a cooled cache for a user who might return. Calling the gate "inverted with respect to its own purpose" is only true inside the retention window; outside it the gate was a sane proxy for imminent return. The PR body's "fired only when warming was worthless" is factually wrong for that middle band.

**The removal still stands**, on the honest grounds the PR also gives: measured cost 6 warms per real turn, benefit never measured, and the SPA refetches conversations on tab refocus (`ChatView.onVisibility` -> `onResync` -> `loadConversations`), so a returning user who switched tabs still hits the remaining trigger. The only genuinely regressed case is a tab that stayed visible and idle past the cooldown, then a send with no list reload: that first turn now eats a cold prefill.

And on "does that matter": I attempted the warm-vs-cold split the PR says the telemetry cannot support, using log adjacency as a crude proxy (tool-free turns only, jarvis.proxy). Turns preceded by a prewarm within 5 minutes: median `first_delta_ms` 12.6s (n=14). Turns with nothing in the prior 5 minutes: median 10.4s (n=6). Tiny, noisy, confounded samples, but the direction is the point: **no first-turn benefit from prewarm is even visible in the only production data that exists.** Removing two of three triggers regresses nothing measurable.

Suggested (non-blocking): soften the "inverted" wording in the module docstring and the hooks.py comment so the next reader is not taught a wrong account of what the cron was for. The economic argument carries the decision by itself.

### Finding 2 (low): stale Scheduled Job Type row until migrate

Deleting `keep_warm_if_active` leaves the existing `Scheduled Job Type` row pointing at a now-missing method until the next `bench migrate` runs `sync_jobs` / `clear_events` (verified frappe prunes it there). Deploys here migrate, so this is a note, not a defect.

### Finding 3 (nits)

- `warm_session` still answers `{"ok": true, "enqueued": true}` when it in fact skipped. Nothing consumes it (no SPA caller exists), but the field is now sometimes a lie; consider reporting the truth since the contract is "always ok" anyway.
- `_EXPLICIT_OFF_VALUES` misses float `0.0` (`str(0.0)` is `"0.0"`, so prewarm stays ON). Unreachable via JSON (`0` parses as int); mentioning it only because the test claims "every explicit falsy value."

### On the questions a reviewer should ask

- **120s recent-turn window:** right order of magnitude. Its real job is absorbing the post-send list refreshes, which land within seconds; anything between about 60s and the 240s cooldown would work. Not worth bikeshedding without the measurement the PR itself says is missing.
- **4-minute cooldown untouched:** correct call, and the Gemini-has-no-documented-TTL caveat is now stated in the comment instead of implied away. Verified the comment says exactly that.

### Taken on trust

The 6ms double-fire observation on jarvis-pool-bf4097 (it originates in issue #548). The race window it describes is real in the old code (get-then-set spanning the Settings DB round trip) and the new re-entrancy test reproduces it deterministically, so nothing hangs on the observation itself.

PR #558, opened 2026-07-30
feat(reconnect): tell a replaced site where its account went

Navin-S-R:
## Why

Revoking the old site's container access (jarvis-admin-v2 #153) closes the security hole, but on its own it leaves the customer with a site that **looks** connected and fails on first use. They deserve to be told.

## The constraint that shapes this

Redemption rotates the *customer's* credentials, and both benches share one customer login. The moment access is revoked, the replaced site can no longer authenticate at all — so it cannot be told over any normal call. There is no authenticated channel left.

## How it works

1. **At redemption**, the control plane leaves a receipt keyed by `sha256(old_site_url)` — no account, no email, 30-day TTL.
2. **The bench asks the one question it still can**: an unauthenticated lookup with its own URL, `{replaced, at, moved_to}`.
3. **The chat gate** — which previously failed **open** on any admin error, sending a dead site into a chat that cannot work — now returns `reason: site_replaced` with the destination.
4. **The banner names the site**: *"Your Jarvis account is now connected to jarvis-2.example.com. Chat here stopped working when it moved."* plus **"Reconnect this site instead"** as the way back.

Both sites belong to the same account holder, so naming the destination discloses nothing across a trust boundary — and it's what makes the message actionable.

## Deliberate choices

- **`site_replaced` is not in `NOT_ONBOARDED_REASONS`.** This site *is* onboarded; it just no longer holds the account. The full-screen setup poster would say the wrong thing.
- **Cached both ways (300s)** so a replaced site doesn't hammer the guest endpoint, and an ordinary outage doesn't either.
- **A failing probe means "not replaced"** — never invent a replacement we cannot prove, and never lock a working site out of chat over a blip.

## Tests

4 Python: a replaced site explains itself; an ordinary admin blip still fails open; the verdict is cached (one probe across two gate misses); a failing probe is treated as not-replaced.

3 JS: the banner names the destination (scheme and trailing slash stripped), still explains itself without one, and stays silent when nothing moved.

29 `test_account`, 59 `test_admin_client`, 48 `test_onboarding_sync`, 11 `readiness.spec`.

## Depends on

`jarvis-admin-v2` `fix/reconnect-revokes-old-bench` (#153) for the receipt. Without it this is inert — the lookup simply always answers "not replaced".

## Not wired yet

`replacedBanner()` and `replacedNoticeOf()` are exported and tested, but no view renders them — ChatView's banner slot is where it belongs, and that's a focused follow-up I did not want to bury in this PR.

### DIFF
--- a/frontend/src/onboarding/readiness.js
+++ b/frontend/src/onboarding/readiness.js
@@ -76,6 +76,31 @@ export async function needsOnboarding() {
 }
 
 // Billing banner payload from the same memoized verdict - no extra round-trip.
+// The account was reconnected on ANOTHER site, so this one lost the workspace.
+// `site_replaced` is deliberately absent from NOT_ONBOARDED_REASONS: this site IS
+// onboarded, it just no longer holds the account, and the full-screen setup poster
+// would say the wrong thing. Returns {} unless it applies.
+export async function replacedNoticeOf() {
+	const r = await checkReady();
+	if (!r || r.reason !== "site_replaced") return {};
+	return r.replaced_notice || { replaced: true };
+}
+
+// Copy for that banner. `moved_to` is the site now holding the account - both
+// belong to the same account holder, so naming it is what makes this actionable.
+export function replacedBanner(notice) {
+	if (!notice || !notice.replaced) return null;
+	const where = (notice.moved_to || "").replace(/^https?:\/\//, "").replace(/\/$/, "");
+	return {
+		tone: "warning",
+		title: "This site no longer has access to your workspace",
+		message: where
+			? `Your Jarvis account is now connected to ${where}. Chat here stopped working when it moved.`
+			: "Your Jarvis account was reconnected on another site. Chat here stopped working when it moved.",
+		action: "Reconnect this site instead",
+	};
+}
+
 export async function billingNoticeOf() {
 	const r = await checkReady();
 	return (r && r.billing_notice) || {};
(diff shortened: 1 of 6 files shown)


Navin-S-R:
## Test status

Run locally on `test_jarvis` (Python) and vitest (JS) against this branch.

| suite | result |
|---|---|
| `jarvis.tests.test_account` (the chat gate this PR changes) | **29 / 29 OK** |
| `jarvis.tests.test_admin_client` | 59 / 59 OK |
| `jarvis.tests.test_onboarding_sync` | 48 / 48 OK |
| `jarvis.tests.test_dev` | 7 / 7 OK |
| `frontend/src/onboarding/readiness.spec.js` (banner copy) | **11 / 11 passed** |
| `frontend/src/onboarding/steps.test.js` | 24 / 24 passed |

**Known pre-existing failure elsewhere in the app suite:** `jarvis/tests/test_api_errors.py` reports 26 errors (redaction/taxonomy). Verified unrelated — stashing this branch's changes and running it on clean `develop` produces the identical 26. Not introduced here, and not fixed here either.

**What the new tests pin:**

- A replaced site explains itself (`reason: site_replaced` carrying the destination) instead of the chat gate failing **open** into a chat that cannot work.
- An ordinary admin blip still fails open — an outage must not lock a working site out of chat.
- The verdict is cached, so a replaced site does not hammer the unauthenticated lookup (one probe across two gate misses).
- A probe that fails is treated as **not** replaced — never invent a replacement we cannot prove.
- The banner names the destination with scheme and trailing slash stripped, still explains itself without one, and stays silent when nothing moved.

**Depends on** `jarvis-admin-v2` #153 (merged) for the receipt written at redemption. Without it this is inert: the lookup simply always answers "not replaced".

Blocked on review approval — I cannot approve my own PR.

PR #563, opened 2026-07-30
White-label: strip "openclaw" -> "agent" across the app (+ retire self-host fields)

venkat102:
## What

White-labels the app: strips the vendored agent-runtime codename **"openclaw" → "agent"** everywhere a customer, operator, or casual reader could see it, and finishes retiring the (already-removed) self-hosted deployment path. Scoped to the **app repo only** — the cross-repo wire contract is deliberately preserved.

## Commits

| | |
|---|---|
| `759b95fe` | rename the 3 transport modules `openclaw_*` → `agent_*` (+ their tests), fix all imports |
| `716bd69f` | rename `Openclaw*` classes → `Agent*` (Session + 2 exceptions) |
| `caec299d` | `ping_openclaw` → `ping_agent` (whitelisted diagnostic + Desk JS caller) |
| `18faefa4` | strict prose strip across 110 files + 2 dir renames + a **guard test** + boundary comment |
| `aa43ec4e` | drop the 4 dead `selfhost_*` Settings fields + patch `v2_09` |
| `0510da70` | `openclaw_seq_watermark` → `agent_seq_watermark` + migration `v2_10` |
| `00f3392f`, `387ab9de` | review-loop fix wave (see Review below) |

## The wire-contract boundary

"openclaw" is retained **only** where it is the runtime's own protocol/config identity — renaming it in the app alone would break `jarvis-openclaw-plugin` / `jarvis-fleet-agent` / the container:

- **Files:** `chat/agent_client.py`, `device.py`, `events.py`, `agent_ws.py`, `relay_mux.py` (WS protocol + device-auth), `tools/tool-names.json` + `_tool_contract.py` (byte-identical plugin contract), `agent_templates/openclaw.json.j2`, historical applied patches.
- **Literals anywhere:** the WS method/param names, the `__openclaw` transcript key the gateway stamps, `openclaw.json`, `.openclaw/` container paths, `ghcr.io/openclaw/openclaw`, `__openclaw__/ws`, and `openclaw/{src,extensions,docs}` cross-repo source citations.

`agent_client.py` documents this boundary, and **`test_no_openclaw_leak`** walks the tree and fails if any un-sanctioned "openclaw" reappears.

## Migration (`v2_10`)

`frappe` model-sync adds the renamed column zero-filled and keeps the old one as an orphan, so the `post_model_sync` patch **copies** `openclaw_seq_watermark` → `agent_seq_watermark` with a clobber-safe, idempotent guard (`WHERE agent_seq_watermark = 0 AND openclaw_seq_watermark > 0`). The old column is **retained this release as the rollback net** and dropped in a later contract patch. `permlevel:1/hidden:1/read_only:1` preserved. `test_seq_watermark_migration` covers copy / idempotency / no-downgrade on populated rows.

## Review

Went through the full multi-dimension review loop. It earned its keep: it caught a **Critical** the mechanical strip introduced — the blanket rename had changed the wire-protocol transcript key `.get("__openclaw")` → `.get("__agent")`, which would have silently disabled snapshot-recovery watermark fencing in prod (stale-answer / double-finalize), with the test fixtures renamed in lockstep so it looked green. Fixed in `00f3392f` (revert the key + fixtures + extend the guard allowlist), plus corrupted cross-repo citations and a fresh-install-safe migration test. Cycle-2 confirmed resolved.

## Verification

Targeted suites green in isolation on `jarvis-test.localhost`: client 33/33, pool 14/14, diagnostics 12/12, security 19/19, native-rpcs 10/10, turn_recovery 35/35, pump_pipeline 39/39, migration 3/3, oauth 20/20+7/7, guard 1/1. `bench migrate` clean (both seq columns present post-migrate). Sanctioned wire literals byte-verified.

> ⚠️ **The full local suite is red from two pre-existing `develop` issues, not this PR** (proven by isolation + stash-compare): the branding validation `jarvis_settings.py:34` rejecting the test-site's seed `brand_logo`/`brand_favicon` placeholders (commit `3769819c`), and the `release_update_required` send gate (`chat/policy.py`, commit `215af7af`).

## Follow-ups (not this PR)

- A later **contract patch** to drop the retired `openclaw_seq_watermark` column once this is proven in prod.
- The **persona** white-label (scrub + confidentiality guardrail) — separate `jarvis-persona` PR.
- `jarvis_admin` operator docs still mention `ping_openclaw` — a fast-follow doc fix in that repo before rollout.
- Pre-existing (predates this branch): `hooks.py` cites `openai-codex-device-code.ts` where the real file is `openai-chatgpt-device-code.ts`.


Vigneshsekar:
Merge conflicts resolved — develop merged in at 0de6fa4d (merge commit, no history rewrite).

- **ChatView.vue:** took develop's side. Develop moved the model-picker block out of the conversation header into the new `ModelEffortPicker.vue`/`PersonaPill.vue` components; this branch's one comment rename inside the old block is superseded by the move. Nothing from the picker rework was dropped.
- **patches.txt:** union, develop's `v2_09_backfill_persona_enabled` first (it's already on deployed sites), then this branch's `v2_09_drop_selfhost_settings_fields` and `v2_10_rename_openclaw_seq_watermark`.
- **White-label semantic pass:** develop's newer commits had introduced three "openclaw" comment references (admin_client.py, GeneralPane.vue, llm/pool.js) — renamed to the agent vocabulary; `test_no_openclaw_leak` is green on the merged tree.

Local gates on the merged tree: patterntest migrate runs the full patch chain cleanly (both watermark columns + persona backfill verified); backend suites green (no_openclaw_leak, chat_worker, pump, turn_state, chat_admission, user_settings, unified_llm_config); frontend node sweep 432/432, vite build ✓; pre-commit clean on all hand-touched files. One vitest failure in `tests/support-extraction/support-thread.test.js` reproduces identically on plain develop — pre-existing there, not from this merge.

Not addressed here (per scope): the watermark rollback finding — post-upgrade writes still touch only `agent_seq_watermark`, so the retained old column is not a working rollback net. That still needs the dual-write (or a documented drain-before-rollback policy) before merge.

Vigneshsekar:
**Complete review of this PR (5-dimension panel + adversarial verification of every finding; head 0de6fa4d). Verdict: REQUEST CHANGES — three items, then this is a merge candidate.**

**What's verified sound (rely on it):** the wire contract holds everywhere — no renamed identifier crosses a repo or process boundary (protocol verbs, `__openclaw`/`__openclaw__`/`openclaw_state` keys, `openclaw.json`, the ghcr image, plugin entry points verified against jarvis-openclaw-plugin source, oauth on-wire values, redis locks, realtime events). No scheduled/enqueued function's module was renamed (no Scheduled Job Type orphans or stranded RQ jobs); all 171 mock-patch string targets and 410 dotted literals resolve; a rename-normalized whole-file diff across all 135 files shows zero semantic drift; the only renamed whitelisted endpoint (`ping_openclaw`→`ping_agent`) has no code callers anywhere in the sibling repos. Patch ordering is correct (post_model_sync), the drop patch is idempotent, and the self-host fields were already dead. UI text is fully renamed and the Settings pane stays coherent.

**Blocking:**

1. **Watermark transition correctness (both directions).** `v2_10` is a one-shot copy; old-code workers keep writing only `openclaw_seq_watermark` until the post-migrate restart while new code reads only `agent_seq_watermark` — those turns read watermark 0, which disables the min-seq recovery fence (the exact answer-stealing bug the watermark was added for, observed live 2026-07-03). Fix for this compat release: **dual-write both columns at both write sites (prepare.py, turn_handler.py) and read `GREATEST(agent, openclaw)` at the read sites (pump.py, turn_recovery.py)**; drop the old column in a later contract patch. One fix closes the deploy window AND the previously-raised rollback net.
2. **The migration test must actually run.** `test_seq_watermark_migration` self-skips whenever the legacy column is absent — i.e., on every fresh CI site — so the PR's only data-preserving migration has zero automated coverage. Have setUp create the legacy column when missing (simulated upgrade), then test copy + idempotency + clobber-guard.
3. **Sed-proof wire-contract pin.** The guard is one-directional: it catches leftover "openclaw", never a wrongly-renamed wire identifier — and this branch's own history shipped exactly that critical once (00f3392f: `__openclaw` read sites renamed to `__agent` with fixtures renamed in lockstep; tests stayed green against a fictional wire shape). A verifier proved by simulation that a blanket rename still passes the suite today. Add positive assertions that the frozen literals appear at the production read sites (`__openclaw` in prepare/turn_handler/turn_recovery; the `__openclaw__` marker in canvas.py + dashboardSrcdoc.js), assembled sed-proof (string parts or digest — precedent: test_tool_contract.py:56-65).

**Strongly recommended in the same push:** extend the guard's scan roots — `pwa/src` (a shipped customer SPA), the desk widget, `.html`/`.css`, and README.md are unscanned today (clean now, but history shows these surfaces carry openclaw text) — plus a scanned-file floor so a moved root can't pass vacuously.

**Non-blocking follow-ups (recorded):** README link text "agent" over an unchanged openclaw href + pyproject "powered by agent" prose; `agent doctor` in an admin_client docstring names a CLI command that doesn't exist; allowlist trims (relay_mux.py needs no exemption); admin-v2/persona doc rot naming renamed symbols (doc-only sibling PRs); `agent_workspace_seeds/` is orphaned dead code — delete or wire up.

Full review record: fix-phase-20260726/REVIEW-563-WHITELABEL.md (in the design/ops repo on the dev box).

Vigneshsekar:
All three review blockers + the guard extension are implemented at 5b2a0ae1:

- **B1 — watermark transition safety, both directions.** New `jarvis/chat/seq_watermark.py`: `stamp_watermark()` dual-writes the new and legacy columns while the legacy column exists (rollback-safe), `wm_expr()` reads `GREATEST` of the pair (the deploy window — an old-code worker stamping only the legacy column after the one-shot v2_10 copy — stays visible; `min_seq` can no longer silently read 0). Wired at both write sites (prepare, turn_handler) and every read site (pump reconcile bound, turn_recovery scans, `_next_turn_watermark`). When the later contract patch drops the legacy column, both helpers collapse to the new column with no code change.
- **B2 — the migration test now runs everywhere.** setUpClass creates the legacy column on fresh sites (simulated upgrade) instead of skipping; three new transition tests cover dual-write, a legacy-only write surfacing through `wm_expr`, and the real `_next_turn_watermark` path. Replayed red against a single-column compat layer (3 failures, exactly the new tests).
- **B3 — `test_wire_contract_pins`:** positive, sed-proof assertions (literals assembled from parts) that the gateway transcript-metadata key and the live-frame marker are read/written verbatim at the named production sites. Replayed red against a blanket-renamed prepare.py — the failure message names the file and the consequence.
- **S1 — guard reach:** now scans `pwa/`, the desk-widget source (only the built `/public/frontend` bundle stays skipped), `.html`/`.css`, with a per-root scanned-file floor so a moved root fails loudly instead of passing vacuously.

Local gates: wire pins, leak guard, migration (6), turn_recovery, chat_worker, pump, pump_pipeline all green on patterntest; pre-commit clean. With these in, my review verdict flips to **APPROVE** once CI confirms — remaining items from the review are the recorded non-blocking follow-ups (README/pyproject prose, allowlist trims, admin-v2/persona doc rot, orphaned agent_workspace_seeds/).

Deploy note for the eventual merge: `bench migrate` before restart (v2_09 backfill + v2_09 drop + v2_10 copy), and schedule the later contract patch that drops `openclaw_seq_watermark` once the release has soaked.

Vigneshsekar:
CI 6/6 green at bd305a7e — the fresh-site run now exercises the simulated-upgrade migration path end to end. As stated in the review: **verdict is APPROVE**. All three blockers (watermark transition compat layer, un-vacuumed migration test, sed-proof wire pins) plus the guard extension and the lockfile-churn revert are in. Ready for merge — deploy with `bench migrate` before restart, and schedule the later contract patch that drops `openclaw_seq_watermark` once the release has soaked.

kavin-114:
## Review at `2e82132f`

Reviewed the full diff (136 files, +1448/-1032), the two patches, the compat shim and the guard test. The engineering here is good: the wire-contract boundary is drawn in the right place, the guard test is built to resist going vacuous, and the `seq_watermark` dual-write shim is a genuinely careful answer to a rename that has live traffic on both sides of it.

**One blocker, and it is not this PR's fault.** `develop` moved underneath it today.

### Blocker: the guard test fails once `develop` is merged in

`test_no_openclaw_leak` passes on this branch in isolation (I ran its logic standalone: PASS, 0 offenders). After merging the current `develop` tip `de517103` it reports **9 offenders**, all introduced by `fa36601d` (PR #577, "record which model produced each reply", merged today):

```
jarvis/chat/finalize.py:315                                   an openclaw-direct pool that ...
jarvis/chat/usage.py:143                                      openclaw only keeps its model.fallbacks ...
jarvis/tests/test_chat_agent_client.py:1001                   An openclaw-direct pool ...
jarvis/tests/test_pump_pipeline.py:1858                       the REAL openclaw payload shape ...
jarvis/tests/test_pump_pipeline.py:1951                       An openclaw-direct pool ...
jarvis/.../jarvis_chat_message.json:198                       description: "... openclaw ..."
jarvis/.../jarvis_chat_message.json:207                       description: "the openclaw provider id ..."
frontend/src/utils/modelBadge.js:6                            keeps openclaw's failover chain live ...
frontend/src/views/ChatView.vue:4488                          keeps openclaw's failover chain ...
```

Merging as-is turns `develop` red. Two of those are DocType field descriptions, which are operator-visible in Desk, so they are exactly what the white-label is meant to catch. The guard did its job.

Fix is mechanical: rename those 9 to "agent". It can land either on this branch after a `develop` merge, or on `develop` first as a follow-up to #577.

### On the "conflicting" badge

GitHub shows `mergeable=CONFLICTING / dirty`, but that is a stale cache. A real trial merge of `refs/pull/563/head` into `de517103` is clean, zero conflicted paths (`git merge-tree` exit 0, and a full worktree merge auto-merged all 12 overlapping files). Do not go hunting for text conflicts. The semantic conflict above is the real one.

### Verified good

- **Merged DocType JSON is coherent.** #577 added `model` to the same `Jarvis Chat Message` JSON this PR renames a field in. Post-merge: 24 fields / 24 `field_order` entries, no duplicates, no orphans, `agent_seq_watermark` present, `openclaw_seq_watermark` gone, #577's `model` field intact.
- **Both patches sit under `[post_model_sync]`** (line 5 marker, entries at 40 to 42), so `v2_10` runs after the new column exists.
- **`v2_10` is fresh-install safe and clobber-safe.** The `get_table_columns` guard returns early when the old column never existed, and `WHERE agent_seq_watermark = 0 AND openclaw_seq_watermark > 0` makes a re-run unable to downgrade a live watermark. This matters given `install_app(set_as_patched=True)` marks patches done without running them.
- **`v2_09_drop_selfhost_settings_fields`** is parameterized and idempotent, and clears the Settings cache. No `selfhost` residue anywhere outside patch and test prose.
- **Wire contract intact.** `.get("__openclaw")` is preserved in `turn_recovery.py:96`, `turn_handler.py:1006` and `prepare.py:175`, and the `__openclaw__/ws` and canvas markers are untouched. This was the Critical the review loop caught and the fix held.
- **No cross-repo break from the `ping_openclaw` rename.** I grepped admin, admin-v2, fleet-agent, plugin and persona for executable callers of `jarvis.diagnostics.*`: zero. The only hits are markdown docs, which matches the follow-up already listed. In-repo, `jarvis_settings.js:40` calls `ping_agent` and the def matches.
- **No SQL injection in the compat shim.** `wm_expr(alias)` interpolates only literal call-site aliases (`""` and `"m."`), values stay parameterized, and every call site is inside a function body, so `has_legacy_column()` never runs at import time without a DB.
- **`GREATEST(...)` costs nothing at the access path.** In `_next_turn_watermark` the index work is done by `conversation` plus `seq`; the watermark predicate was always a residual filter and there was never an index on that column.
- **Static resolution of all 635 Python files in the merged tree: every intra-app import resolves.** No dangling `chat.openclaw_client`, `openclaw_session_pool`, `openclaw_ws` or `openclaw_templates` reference survives. All JSON parses.
- **Diagnostics stay gated** on `require_jarvis_admin()`, so JRV-SEC-011 is not reopened.
- Isolating the non-mechanical changes with a rename-normalised diff confirms the only real logic in this PR is the `seq_watermark` shim plus the boundary docs. Everything else is substitution.

### Minor, non-blocking

`jarvis/agent_templates/` and `jarvis/agent_workspace_seeds/` have no reader in this app and no consumer in fleet-agent, plugin or admin-v2. They look like dead assets carried along by the rename. Worth deleting in the same follow-up that drops the legacy column, rather than white-labelling files nothing loads.

### Verdict

Approve on content. Holding the merge only on the guard-test regression above. Once those 9 references are renamed and CI is green on a head that includes `de517103`, this is good to go.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

kavin-114:
Follow-up on the guard-test blocker above: I opened #592 against `develop` to fix it, since all nine offenders came from #577 rather than from this branch.

#592 renames only those nine, deliberately leaving every other `openclaw` string on `develop` alone because this PR renames them itself and touching them there would hand you textual conflicts.

Verified: trial-merging this PR onto #592 auto-merges with **0 conflicted paths**, and the guard's own scan against that merged tree reports **0 offenders**, down from 9. All six hosted checks pass on #592.

Merge order: #592 first, then merge `develop` into this branch and CI should go green.

kavin-114 (approved):

PR #573, opened 2026-07-31
fix(settings): decide endpoint locality by address, not provider id

### LINKED ISSUE #556: A model with a local or private base URL cannot be connected unless its provider is ollama or vllm
## Summary

A model whose **base URL is local or private** cannot be added through Settings,
AI models unless its provider happens to be `ollama` or `vllm`. Connect runs a
key probe from the **bench**, the bench cannot reach the endpoint, and Connect
then returns without saving anything.

The most realistic case is `OpenAI-Compatible` pointed at a self-hosted gateway,
which is exactly what that provider exists for.

## Repro

Settings, AI models, Add a model, API key tab:

- Provider: **OpenAI-Compatible**
- Model: `claude-sonnet-4-6`
- API key: (any valid key for that endpoint)
- Base URL: `http://host.docker.internal:9000/openai/v1`

Press **Connect**.

Observed: the panel stays open, nothing is written, and the pool is unchanged.

```
tabJarvis LLM Pool Model   ->  still only the 2 pre-existing rows
last_sync_status           ->  unchanged, "ok (pool via admin)"
```

Pressing **Test** first shows why:

> ✗ **Test failed.** Could not resolve host: host.docker.internal
> Tested from the bench's network. Live chat runs from inside your Jarvis
> container, for a public provider these agree.

The endpoint is genuinely fine. From inside the tenant container it is reachable:

```
$ docker exec <tenant> curl -s -o /dev/null -w "%{http_code}" \
    http://host.docker.internal:9000/openai/v1/models
401          # reached the service, rejected for missing auth
```

So the config is correct and the UI refuses to save it.

## Root cause

`connectApiKeyRow` in `frontend/src/components/LlmPoolEditor.vue` probes before
writing, and bails on a failed probe:

```js
if ((row.apiKey || "").trim() && !isLocalProviderRow(row)) {
    setBusy("Checking your key…");
    try { await testApiKeyRow(row); } finally { setBusy(""); }
    const probe = panel.value.testResult;
    // Nothing else to say: the red Test result block above the button is already
    // showing the provider's own words for why it refused.
    if (!probe || !probe.ok) return;
}
```

The escape hatch is `isLocalProviderRow`, which keys off the **provider id only**:

```js
function isLocalProviderRow(row) {
    return !!(row && LOCAL_PROVIDER_IDS.has(providerId(row.provider)));
}
```

and

```js
// frontend/src/llm/pool.js:219
export const LOCAL_PROVIDER_IDS = new Set(["ollama", "vllm"]);
// jarvis/llm_key_probe.py:71
LOCAL_PROVIDER_IDS = {"ollama", "vllm"}
```

The base URL is never consulted. The surrounding comments show the locality
problem was understood:

> "a local endpoint (ollama, vllm) is reachable from the CONTAINER rather than
> from this bench, so a failure here would say nothing about the key while
> blocking a perfectly good save."

That reasoning is right, it is just applied by provider name instead of by where
the endpoint actually points.

## Affected, beyond the case above

- `OpenAI-Compatible` at `localhost` / `127.0.0.1` / `host.docker.internal`
- Any provider on a customer LAN (`10.x`, `192.168.x`, `172.16-31.x`)
- Any provider behind internal-only DNS
- A self-hosted enterprise gateway on a private address

`openai_compat` is the provider a customer reaches for when self-hosting, so this
is a normal configuration rather than an exotic one.

## Suggested fix

Decide locality from the **resolved base URL**, not the provider id. Treat the
row as local when the base URL host is `localhost`, `127.0.0.0/8`, `::1`,
`host.docker.internal`, a private RFC1918 range, or `.local` / a bare hostname,
and skip the blocking probe for it, keeping the existing "can only be verified
from inside your container" disclaimer.

Keep `LOCAL_PROVIDER_IDS` for the key-optionality rule (ollama and vllm genuinely
take no key); it is only the probe gate that needs the URL-based test. Note the
constant is duplicated in three places (`pool.js`, `LlmPoolEditor.vue` via import,
and `llm_key_probe.py`) with comments requiring they stay in sync, so a
URL-based helper should live in one place too.

## Secondary: the failed Connect is easy to read as a no-op

The bail is deliberately silent because "the red Test result block above the
button is already showing the provider's own words". That holds when the block is
on screen, but the button gives no feedback of its own, so pressing Connect on a
tall panel (where the Test block may be scrolled out of view) looks like the
button simply did nothing. A short inline result next to the button, or scrolling
the Test block into view on a blocked Connect, would remove the ambiguity.

## Also spotted in the same form

1. **Stale validation message.** "Choose a provider first." stays visible after a
   provider is chosen, and remained visible with every field filled. It only
   cleared when Connect was pressed.
2. **Clicking "Add a model" immediately inserts a row into the live list**
   (`3  Subscription · OpenAI  gpt-5.5  Pending re-check`, with Edit / Reconnect /
   Remove) and flips the pool badge to `Proxy (failover)`, before anything is
   configured. It is correctly **not** persisted (verified in the DB), but it is
   visually indistinguishable from a saved row and defaults to a provider and model
   the customer never picked.
3. The 15-item provider dropdown has no typeahead filtering, so reaching
   `OpenAI-Compatible` at the bottom needs scrolling.

Found during the manual e2e checklist follow-up on 2026-07-30.

kavin-114:
Fixes #556

**Stacked on #569** (`fix/connect-ui-polish`), which also rewrites
`LlmPoolEditor.vue`. Merge #569 first and this diff shrinks to its own commit.

A model with a local or private base URL could not be connected unless its
provider happened to be `ollama` or `vllm`.

## Root cause

The pre-flight probe skipped only those two, **by provider id**:

```js
if ((row.apiKey || "").trim() && !isLocalProviderRow(row)) { ... }
// LOCAL_PROVIDER_IDS = {"ollama", "vllm"}
```

So a self-hosted endpoint reached through any other provider (`openai_compat`
pointed at a LAN address, a private gateway) was probed from the bench, could not
be reached, and the connect refused. The existing comment already had the right
reasoning ("reachable from the CONTAINER rather than from this bench"); it was
just testing the wrong thing.

## The fix

Locality follows the **address**. A private endpoint reached through
`openai_compat` is exactly as unreachable from this bench as one reached through
`ollama`, and the provider id was never evidence either way. Loopback, RFC1918,
CGNAT, link-local, `.local`-style suffixes and bare single-label container aliases
all skip the probe.

`ollama` and `vllm` keep their id-based carve-out for the case where no base URL
has been typed yet.

## SSRF direction

Worth being explicit, since this touches a known open concern. The change **only
ever removes** a bench-originated request to a private address; it never adds one.
`link_fetch.request_pinned` is untouched and remains the thing that actually
refuses a request. A malformed host is deliberately treated as public, not local,
so a typo still gets probed rather than being quietly waved through unchecked.

## The silent half

A probe that returned a **result** was already explained by the red Test block
above the button. A probe whose request failed outright set no result at all, and
the bare `return` left the customer pressing Connect against complete silence.
That case now says so.

## Tests

5 JS (`node --test`, 75 pass in that file) and 5 Python. Both planes assert the
same table of addresses, including the two cases my first draft got wrong and the
tests caught: `URL.hostname` keeps IPv6 brackets in JS but not in Python's
`urlsplit`, and `100.64/10` is CGNAT which `ipaddress.is_private` does not cover.
Both also assert that a malformed host is not treated as local, and that public
endpoints are still probed.

## Not verified

`npm run build` could not be run in this worktree: vite writes its temp config
into the worktree-root `node_modules` and cannot resolve itself from there, a
known local nested-worktree trap unrelated to this change. ESLint, which parses
Vue SFCs, passes via the repo's own pre-commit gate, and the unit tests pass. CI
is the real gate.

No live private endpoint was available, so the end-to-end connect of a real LAN
model was not exercised.


kavin-114:
## Correction: the backend half is reverted

CI caught a real regression in the first commit, and it is worth recording rather
than quietly amending.

`test_ssrf_blocked_endpoint_for_a_non_local_provider_still_fails_cleanly` probes
`169.254.169.254`, the cloud metadata service address and the canonical SSRF
target, and asserts `local_endpoint` is `False` with none of the reassuring
"container" wording. Widening `is_local` to cover any private address made that
probe report `local_endpoint: True`, which would have softened an SSRF rejection
into a benign "we cannot check this from here" disclaimer on exactly the address
an attacker would aim at.

The test was right and the change was wrong. It is reverted, and the test is
untouched.

## Why keeping the two planes in step was the wrong instinct

They are answering different questions:

| plane | question | correct key |
|---|---|---|
| frontend | should we **probe** this address at all? | the address |
| backend | how do we **present** a probe that already ran and was blocked? | the provider |

The frontend change alone fixes #556. A private endpoint is never probed, so it
never reaches the blocked-and-explained path at all. A probe that *is* run against
a private address is now only a deliberate Test-button press, and reporting that
as blocked is correct behaviour, not a bug.

The Python helper and its tests are removed with the revert rather than left
sitting unused.

The PR description above still says the two copies are kept in step. That sentence
no longer applies, and this comment supersedes it.

PR #581, opened 2026-08-01
fix(onboarding): make the failed-payment resume reachable again

kavin-114:
## The bug

A customer whose card is declined lands with a `Pending Payment` row. Retrying the wizard hits admin's duplicate guard, and `start_signup` is supposed to fall back to the authenticated `resume_pending_signup` so checkout reopens.

It never did. The fallback gated on the substring `"already registered or pending"`. Admin was later reworded to `"An account for this email and company already exists."`, the guard stopped matching, and every real failed-payment retry dead-ended on the duplicate error.

## Evidence

Reproduced live against `jarvis.admin.v2`:

```
POST .../billing.signup.signup   (fresh)  -> 200, razorpay order created, status Pending Payment
POST .../billing.signup.signup   (retry)  -> 409 DuplicateEntryError
                                             "An account for this email and company already exists."
```

Then, with credentials stored and the email matching (the best possible case for the fallback):

| message fed to the guard | resume attempted |
|---|---|
| what admin sends today | **NO** -> dead end |
| the wording the guard matched | YES |

## Why no test caught it

The only surviving copy of the old wording in the tree was this suite's own `_DUP` fixture. All four tests kept passing against a message nothing emits. `OnboardingView.vue` had already been updated to match both wordings; only the backend guard was missed.

## The fix

Key the decision off admin's exception **class**, not its prose:

- `AdminValidationError` carries `exc_type` through from the response envelope.
- `_is_duplicate_signup_error()` checks `exc_type == "DuplicateEntryError"` first, with a regex fallback covering **both** wordings for an admin old enough not to send `exc_type`.
- `start_signup` calls `admin_client.signup` directly instead of through `_surface`, which flattened the error to a bare `frappe.throw` and discarded `exc_type`; anything non-resumable still goes through the identical `_throw_admin_error` mapping, so no behaviour change there.

Tests now pin admin's real message, add the `exc_type` path (which cannot rot), and keep the legacy wording covered.

## Scope

This fixes recovery when the bench **still holds** its credentials. A customer who uninstalled and reinstalled the app has no credentials to authenticate the resume with, and is still blocked until the 30-day reaper. That case needs the `reconnect.py` code flow widened to `Pending Payment` and is deliberately not in this PR.

## Verification

Hosted CI is the gate for this repo. Locally, all five guard cases pass against the branch, including the two that were broken.

### DIFF
--- a/jarvis/admin_client.py
+++ b/jarvis/admin_client.py
@@ -1628,7 +1628,10 @@ def _do_post(url: str, body: dict, headers: dict, timeout_s: int, admin_url: str
 	# class isn't recognised.
 	if exc_type:
 		if exc_type in ("ValidationError", "DuplicateEntryError", "DoesNotExistError"):
-			raise AdminValidationError(clean)
+			# Carry admin's exception class forward. Callers branch on it
+			# (see onboarding._is_duplicate_signup_error) rather than on the
+			# message prose, which admin is free to reword at any time.
+			raise AdminValidationError(clean, exc_type=exc_type)
 		# AuthenticationError ~ a token/credential failure (retry-eligible, 401);
 		# PermissionError ~ an authorization denial (terminal, 403). Tag the
 		# status so _post re-mints on the former but surfaces the latter as-is.
--- a/jarvis/exceptions.py
+++ b/jarvis/exceptions.py
@@ -160,7 +160,18 @@ class AdminAuthError(JarvisError):
 class AdminValidationError(JarvisError):
 	"""jarvis_admin raised a Frappe ValidationError (or similar user-input
 	error) inside a whitelisted endpoint. Carries the clean operator-facing
-	message - never the traceback dump."""
+	message - never the traceback dump.
+
+	``exc_type`` carries admin's own exception class name (e.g.
+	"DuplicateEntryError") when the response declared one. Callers that need to
+	branch on WHICH validation failed must key off this, never off the message
+	prose: the failed-payment resume used to match the substring "already
+	registered or pending", admin later reworded that message, and the resume
+	became silently unreachable. None when admin sent no exc_type."""
+
+	def __init__(self, message: str, *, exc_type: str | None = None):
+		super().__init__(message)
+		self.exc_type = exc_type
 
 
 class AdminRateLimitedError(JarvisError):
--- a/jarvis/onboarding.py
+++ b/jarvis/onboarding.py
@@ -29,9 +29,7 @@ from jarvis.permissions import grant_onboarding_admin, require_jarvis_admin
 # error from the customer. The onboarding view keys its reconnect OFFER off the
 # looser pair, which is fine for a display hint but too loose for control flow.
 # exc_type is the primary signal; this only runs when admin sent none.
-_DUPLICATE_SIGNUP_RE = re.compile(
-	r"account\b.{0,60}?already\s+(?:exists|registered or pending)", re.I | re.S
-)
+_DUPLICATE_SIGNUP_RE = re.compile(r"account\b.{0,60}?already\s+(?:exists|registered or pending)", re.I | re.S)
 
 
 def _require_admin_url() -> None:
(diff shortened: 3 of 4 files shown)


Vigneshsekar:
Review finding from the onboarding recovery workstream (verified against develop `de517103` + admin-v2 `1437298`):

**The exc_type direction here is right, but the resume has a second, independent kill this PR doesn't reach — the identity check.**

- Admin's signup response returns `"customer": customer.user`, which is the **synthetic OAuth login** `cust-<hash>@jarvis.invalid` (`billing/signup.py` `_synthetic_login`, and the response-build comment says exactly why: it's the password-grant USERNAME).
- The bench stores that verbatim: `onboarding.py:108` `s.db_set("jarvis_admin_customer_email", data["customer"])`.
- The gate this PR keeps compares it against the user's typed real email: `onboarding.py:580-583` — **structurally never equal on any real deployment**.

So with this PR merged, `_is_duplicate_signup_error()` now correctly fires, and then `_try_resume_pending_signup` still returns `None` at the email comparison — the wizard dead-ends on the same duplicate error. The live repro in the PR body ("with credentials stored and the email matching — the best possible case") hand-builds the one state production never produces.

Suggested fix inside this PR's scope: drop the email comparison entirely and let possession of the stored credentials be the ownership proof — admin's `resume_pending_signup` already resolves the customer from authentication, so the bench-side check adds nothing but this false negative. (A cheap belt-and-braces alternative: also store the *real* email at signup time in a separate settings field and compare against that — but deleting the dead check is cleaner.)

Two adjacent facts worth knowing for this path (not asks on this PR): admin's duplicate rejection currently surfaces as **HTTP 500** (DuplicateEntryError sets no `http_status_code`), which also writes an Error Log snapshot on the control plane per retry; and for autopay/mandate plans the Pending Payment row is **never reaped** (`_holds_money_handle` skips it), so the dead end is permanent, not 30 days. Both are being picked up in the broader onboarding contract work (plans in `/home/vignesh/jarvis-ux-fixes`, judgment in `FABLE-JUDGMENT.md` there).

Vigneshsekar:
Heads-up, Kavin — #586 has landed covering this territory, so this draft is likely superseded. Sharing the reasoning in case it's useful either way.

While building it we found the dead-end has **two independent kills**, and this draft closes the first one:

1. **Prose matching** (what you fixed) — the bench matched admin's message text, which admin no longer sends.
2. **The synthetic-login identity check** — admin returns `customer.user` as `cust-<hash>@jarvis.invalid`, the bench stores that as `jarvis_admin_customer_email`, and then compares it against the user's *real* email. That comparison is structurally never equal, so even with the prose fix in place the resume path still can't fire.

Verified first-hand. #586 removes both, and moves the detection onto stable machine codes + `exc_type` rather than any message text, so an admin-side copy change can't re-break it.

No action needed from you — flagging it so you're not debugging a path that's already covered. If there's anything in this draft #586 misses, say so and it goes in.

kavin-114:
Closing as superseded by #586, and @Vigneshsekar's finding is correct. I verified it against develop rather than take it on trust:

- `billing/signup.py:435` builds the response with `"customer": customer.user`, and its own comment says why it must be the synthetic login rather than the customer hash.
- `onboarding.py:108` stores that verbatim: `s.db_set("jarvis_admin_customer_email", data["customer"])`, so the field holds `cust-<hash>@jarvis.invalid`.
- `onboarding.py:580-583` then compares it against the user's typed real address. Structurally never equal.

So this PR fixes kill 1 and the resume still returns `None` at kill 2. My claim that it made the path reachable was wrong.

**Where my verification failed, since it is the more useful lesson:** my repro script stubbed `Jarvis Settings` with a fake returning the real email for `jarvis_admin_customer_email`. I mocked the exact field that was broken, so the harness manufactured the one state production never produces and reported a pass. Same trap as the `test_link_fetch` urllib3 case: mock the boundary hard enough and the test asserts the bug as correct.

#586 removes both lines, moves detection onto `exc_type` + machine codes, and carries fixtures for the legacy wording, so nothing here is lost. Nothing in this draft that #586 misses.

One thing to carry across: **the activation argument moves from this PR to #586.** admin-v2 #163 (resume nulls `razorpay_subscription_id`, the only key both webhooks resolve by, orphaning a live mandate) is latent purely because resume has never run. #586 is what makes it run. The fix is admin-v2 **#200**, green, and it should land with or before #586 or you ship the activation without the guard.

PR #585, opened 2026-08-02
feat(list-filters): the shared FilterGroup UI on the Skills and Macros pilots (plan 08, Phase 2)

Vigneshsekar:
## What this is

Phase 2 of the universal list-filter workstream — **stacked on #583** (the server core); lands on develop when it merges. The Frappe-style Field/Operator/Value filter panel, generated entirely from `get_list_filter_schema`, live on the two pilot surfaces (Skills, Macros). Every other surface keeps the legacy FilterButton untouched — source-pinned, and cross-pinned to the server registry (shipping FilterGroup on a PENDING view fails a test).

## What ships

- **`FilterGroup.vue` + `FilterValueControl.vue` + `filterModel.js`**: searchable field picker with child-DocType grouping, per-field operator menus with schema defaults, the full value-control matrix (text with like-hints, Select with honest "Not set", Check, both-bounds-required date/datetime Between, numeric, Link autocomplete via the existing permission-checked `search_link`, in/not-in chips), repeatable fields, a 20-clause cap note, and full keyboard/screen-reader support (positional labels when a field repeats, deliberate focus routing on add/remove/clear-all).
- **URL as state of record** (`fv2`, versioned): reload/share/back-forward reconstruct the list; clauses the recipient can't use are dropped **with a visible notice** — never a silent unfiltered list (unreadable/oversize links get their own distinct notices; the panel refuses to write a URL too large to survive the round trip). Tab switches carry the query; a navigation that drops it is re-asserted, while an authored Clear stays cleared (structurally — the re-assert can only fire when a written param exists).
- **One reconcile path** for mount, back/forward, and navigation: schema fetch → drop-with-notice → quick-filter sync. Quick filters canonicalize into clauses (declared `rootDoctype`, schema not required) so the toolbar strip and the panel can never disagree, and legacy+v2 coexist for the compatibility window.
- **Error ergonomics**: every wire code maps to an actionable state (self-healing stale-schema loop, inline row attribution with word-boundary matching, transient-retry banners) — no dead-end toasts. A failed Link search degrades to a typeable input and **recovers** the moment a query matches again.

## Review trail (doctrine)

Sonnet UX (REQUEST-CHANGES: 2 a11y) + Opus adversarial (REQUEST-CHANGES: 2 P1 — tab-switch state loss, silent unreadable links — + 5 P2; XSS/prototype-pollution/fencing/registry-pin all CLEARED by executed probes) → consolidated fix round → re-probe (**APPROVE**: every finding re-attacked by running the original failing scenarios; the authored-clear-vs-history interleaving proven clean) → final micro-round. `tests/list-filters/` 172 · full vitest 522 (known pre-existing support-thread failure only) · build green.

Ruling recorded in the checklist: filters are **sticky** across bare navigation (Frappe-consistent); Clear All is the unfilter affordance.

## Deploy

Frontend-only on top of #583; no schema, no migrate. Wave 1–3 surface migrations follow the MIGRATION-CHECKLIST (now including the `rootDoctype` prerequisite and the sticky ruling).

https://claude.ai/code/session_01Nxn1qvS9817vFv6gmLYEoN

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -210,8 +210,16 @@ export const getActiveTurn = (conversation) =>
 	call("jarvis.chat.admission.active_turn_for_conversation", { conversation });
 
 // Mentions: reuse Frappe's built-in Link-field search (no custom backend).
-export const searchLink = (doctype, txt) =>
-	call("frappe.desk.search.search_link", { doctype, txt: txt || "", page_length: 8 });
+// `pageLength` defaults to the mention picker's 8; the list FilterGroup asks for
+// plan 08 §9's Link-suggestion cap of 20. Frappe's own search_link enforces the
+// target DocType's read permission + user permissions, which is why plan 08 adds
+// no second search endpoint.
+export const searchLink = (doctype, txt, pageLength) =>
+	call("frappe.desk.search.search_link", {
+		doctype,
+		txt: txt || "",
+		page_length: pageLength || 8,
+	});
 
 // Field metadata for the record-edit action card: powers Link/Select/Date
 // controls (returns {ok, doctype, fields:[{fieldname,label,fieldtype,options}]}).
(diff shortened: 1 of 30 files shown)


Vigneshsekar:
## Fix: the field picker was permanently empty (3c89ce0)

### The defect

`FilterGroup` requested its filter schema on the Popover's `open` event. That event never arrives, so the schema was never fetched and the field picker rendered frappe-ui Autocomplete's empty state — "No results found" — on **every migrated list surface** (skills, macros, triggers, saved dashboards, wiki).

Why `open` never arrives, in frappe-ui 0.1.278:

- `Popover.vue` emits `open` from exactly one place, `onUpdateOpen`, which is bound to reka-ui `PopoverRoot`'s `update:open`.
- frappe-ui binds that root with `v-model:open="isOpen"` using a real boolean, so the root is **fully controlled** and only reports changes *it* initiates (outside click, Escape).
- Our panel opens from a `Button` inside the `#target` slot, calling `togglePopover()`. That runs frappe-ui's own `isOpen` setter, which emits **`update:show` and nothing else**.

So the only path that ever fired `open` was one the panel can't take. The eager `fv2=` URL path populated the picker correctly, which is why it looked fine whenever a filter was already in the URL.

### The fix

`frontend/src/components/list/FilterGroup.vue` listens for `update:show` instead, which fires on both open paths:

```
-<Popover placement="bottom-end" @open="onOpen">
+<Popover placement="bottom-end" @update:show="onShowChange">
```

`update:show` also fires on close, and reka relays a single dismissal as **two identical emissions** (verified against the real component), so the handler acts only on a genuine change of state and still asks exactly once per open.

### The test was hollow, and was rewritten

The spec that "covered" this passed against a component that never asked for its schema in a browser. The shared double in `tests/list-filters/stubs.js` handed `#target` a `togglePopover` that emitted `open` — a contract the real component does not have — and `FilterGroup.spec.js` additionally hand-emitted `open` on the stub.

- `stubs.js` — the Popover double now reproduces frappe-ui's real open/close contract, including that the slot path emits `update:show` alone.
- `FilterGroup.spec.js` — the schema specs now press the Filter button, plus a new open → close → open case that pins no double-fetch.
- `tests/list-filters/FilterGroupPopover.spec.js` (new) — mounts the **real** frappe-ui Popover and pins both what it emits on each path and that pressing Filter requests the catalog, so the double cannot drift back into fiction.

Against the unfixed component the rewritten specs fail 5/5; with the fix the list-filters suite is 179/179.

### Checks

- `vitest run` — 529 passed, 1 failed (`tests/support-extraction/support-thread.test.js`, the known pre-existing failure on this branch line, untouched here).
- `npm run test:node` — 451/451 pass.

Commit: `3c89ce0cfc7991acedaea23721de9988d5786438`

PR #606, opened 2026-08-03
fix(onboarding): don't prefill Work email with a reserved-domain address

Navin-S-R:
Third attempt at this, and the narrow one. Prefilling the caller's email is the right behaviour and **stays** — the problem was only ever *which* address.

## The bug

`get_account_defaults` resolves the email from `frappe.session.user`:

```python
email = (frappe.db.get_value("User", user, "email") or user) if user and user != "Guest" else ""
```

On a fresh site that user is Administrator — `admin@example.com`. The SPA writes it into **Work email**, on the step that reads *"We'll set Jarvis up for this workspace and send receipts here."* So the field opens with an undeliverable address in it, rendered exactly like something the customer typed, ready to submit.

## The fix

Drop the email when its domain is reserved (RFC 2606 / 6761), rather than blacklisting one address:

```python
_UNDELIVERABLE_DOMAINS = ("example.com", "example.net", "example.org", "localhost")
_UNDELIVERABLE_TLDS = (".test", ".example", ".invalid", ".localhost")
```

That's the right rule because **every placeholder Frappe ships lives in one**: Administrator is `admin@example.com`, Guest is `guest@example.com` — verified on a live site:

```json
[{"name": "Guest", "email": "guest@example.com"},
 {"name": "Administrator", "email": "admin@example.com"}]
```

and `admin_v2` mints synthetic customer logins at `@jarvis.invalid`. Subdomains count too — `mail.example.com` is as undeliverable as `example.com`.

## No frontend change

Blank is already the "no prefill" signal, so `OnboardingView.vue` is untouched: its `if (d.email && !state.email.trim())` no-ops and the field's own `placeholder="you@company.com"` shows. A real address prefills exactly as before.

## Verification

Live endpoint, session = Administrator:

```json
{"email": "", "company": "Aerele", "companies": ["Aerele"]}
```

Reserved address dropped, company prefill intact. The helper was also exercised across edge cases — `X@EXAMPLE.COM` (case), `x@mail.example.com` (subdomain), `x@examplex.com` and `x@example.co.in` (must NOT match despite the prefix), blank, and a string with no `@`. 15/15 as expected.

## History

Supersedes #602 (removed the prefill outright — merged, then reverted by #605) and #604 (dropped the field from the payload — closed). Both were blunter than the bug warranted; this keeps the convenience and removes only the address that was never a real answer.


Vigneshsekar:
Two commits pushed on top of this branch (production change untouched — `onboarding.py` is byte-identical to the PR's version; verified against develop the diff is exactly the four additions of this PR).

**Why:** running this module **standalone on a cold cache** failed 3–4 of 12 — on the PR head and merged with develop — while the whole-suite CI stayed green. Root cause: the new `_caller_email` helper patched `frappe.db.get_value` broadly, so on a cold cache it intercepted Frappe's internal first load of the `User` meta; the corrupted meta was then cached and the module's *pre-existing* broad `frappe.get_all` patches blew up with `DoesNotExistError: DocType None not found`. CI never saw it because earlier suites had already warmed the metadata cache — an order-dependent green.

**`f66074b5` — test hygiene, three fixes in the one file:**
1. `_caller_email` now delegates to the real `get_value` for everything except the exact `("User", …, "email")` lookup — internal metadata reads are never intercepted.
2. The pre-existing broad `frappe.get_all` patches were narrowed the same way (`_company_rows` helper intercepts only `doctype == "Company"`), retiring the module's cache-warmth dependence entirely.
3. `test_user_default_company_wins` no longer requires the ERPNext `Company` DocType (narrow mock, not a skip) — the module now passes on frappe-only sites too.

**`bd77627a`** — merge of current `develop` (the branch was ~140 commits behind, so the green checks predated the state that would actually land).

**Verification:** standalone cold-cache runs now **12/12 OK on both an ERPNext site and a frappe-only site, twice consecutively** (no self-poisoning); a neighbouring module run immediately after (no cache clear) stays green (26/26) — no bleed. Vacuousness check: with the reserved-email guard neutralized, `test_reserved_caller_email_is_dropped` fails (`'admin@example.com' != ''`) — the tests are load-bearing. No assertion was changed or weakened anywhere.

The production behaviour of this PR is verified correct and unchanged — nice fix, and the narrow suffix rule is the right shape.

PR #608, opened 2026-08-03
fix(chat): confirmation cards break on Frappe v15 / Redis < 6.2 (RedisWrapper.expire_key + GETDEL)

venkat102:
## Symptom

On a customer bench running **Frappe v15** with **Redis 6.0.16**, every gated write failed before its confirmation card could render:

```text
AttributeError: 'RedisWrapper' object has no attribute 'expire_key'
  File "apps/jarvis/jarvis/chat/pending_confirm.py", line 176, in mint
```

The same stack would also reject Redis `GETDEL`, which was introduced in Redis 6.2.

## Root causes

1. `RedisWrapper.expire_key` is absent on Frappe v15. Because the call was inside the fatal park block, its `AttributeError` rolled back an otherwise valid confirmation.
2. `GETDEL` is unavailable on Redis 6.0.
3. The first compatibility patch treated every Redis read/consume error as an absent or expired token. It also assumed that an `EXEC` timeout meant the transaction did not commit; Redis can commit the transaction and lose only the reply.
4. Pending-card resync returned an authoritative empty/partial list during Redis read failures, which could clear live cards from the UI.

## Fix

- Refresh the owner-index TTL through the inherited, version-portable `expire(make_key(...))` path. This hygiene operation cannot roll back a valid park.
- Read token records directly through the inherited Redis `GET`, bypassing Frappe's worker-local cache and its suppression of `ConnectionError`.
- Retry one failed strict mint-verification read before rollback, so a one-off lost GET reply does not destroy an otherwise valid park. A sustained failure still fails closed.
- Consume with Redis 6.0-compatible `WATCH` + `MULTI/EXEC`. The transaction writes a short-lived request-specific claim and deletes the token atomically.
  - Exactly one concurrent request can claim the token.
  - If `EXEC` commits but its reply is lost, that request recovers only its own durable claim.
  - If recovery is also unavailable, the endpoint returns `ConfirmationOutcomeUnknownError` before dispatching any business write.
- Preserve storage failures as typed errors instead of presenting them as expiry:
  - confirm/discard execute no business action and keep the card available;
  - resync does not replace live cards with an empty result;
  - the single-flight gate fails closed rather than staging a second card;
  - Redis failures are logged without logging bearer tokens.
- Keep confirmation cards visible in the main chat, dashboard chat, and trigger chat on storage/transport failures.

## Compatibility

- Customer floor remains **Frappe 15 / Redis 6.0**.
- No `expire_key`, `get_value(use_local_cache=...)`, `GETDEL`, or other Redis 6.2+ command is used.
- The code relies only on inherited Redis commands and transactions available on Redis 6.0 and redis-py used by Frappe 15.
- No migration.

## Validation

- `jarvis.tests.test_pending_confirm`: **49/49 passed**
- `jarvis.tests.test_confirm_gate`: **48/48 passed**
- receipt-chip suite: **13/13 passed**
- realtime pending-event suite: **4/4 passed**
- affected action endpoint cases: **12/12 passed**
- fault injection covers transient mint verification, pre-commit failure, post-`EXEC` reply loss, failed recovery, concurrent confirms, strict resync, and fail-closed single-flight behavior
- all selected pre-commit lint/format hooks passed
- all three changed Vue SFCs compile successfully
- merged cleanly and repeated the core suites against current `develop` at `26c4c0e2`

The local runtime validation bench is Frappe 16 / Redis 7; Frappe 15 / Redis 6 compatibility is covered by source/API verification and explicit old-stack regression tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and hardened after review.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -157,15 +157,16 @@ export const applyAction = (action) =>
 // Write-safety gate (issue #186): confirm a parked ERP write by its one-time
 // token. Pass the conversation the click came from so the server enforces the
 // real conversation guard (#11). Returns the tool result envelope {ok, ...} on
-// success, or {ok:False, error:{type:"InvalidConfirmation", ...}} if the token
-// is gone.
+// success. A gone token returns InvalidConfirmation; a Redis outage returns a
+// ConfirmationUnavailableError/ConfirmationOutcomeUnknownError that the UI must
+// keep on screen because the business action did not run in that request.
 export const confirmTool = (token, conversation) =>
 	call(AC + "confirm_tool", { token, conversation: conversation || "" });
 // Discard a parked ERP write by its one-time token: consumes the token (so it
 // can't replay or re-surface on reload), leaves a durable "discarded" receipt
 // chip in the transcript, and queues a note so the agent's next turn learns it
 // was vetoed. Returns {ok, data:{status:"discarded"|"already_handled"}}; the SPA
-// drops the card either way.
+// drops the card for those success states; storage-error envelopes retain it.
 export const dismissTool = (token, conversation) =>
 	call(AC + "dismiss_tool", { token, conversation: conversation || "" });
 // Resync (issue #186, R3 fix for #3): re-surface the caller's own currently
--- a/frontend/src/pages/dashboards/DashboardChatPane.vue
+++ b/frontend/src/pages/dashboards/DashboardChatPane.vue
@@ -365,6 +365,7 @@ async function refreshPending() {
 	}
 	try {
 		const env = await listPendingConfirmations(conversation.value);
+		if (env && env.ok === false) return;
 		const items = (env && env.data && env.data.pending) || [];
 		// keep in-flight busy flags across refreshes
 		const busy = new Set(pendingCards.value.filter((c) => c.busy).map((c) => c.token));
@@ -376,6 +377,11 @@ async function refreshPending() {
 
 function removeCard(token) {
 	pendingCards.value = pendingCards.value.filter((c) => c.token !== token);
+}
+
+function confirmationStorageUnavailable(response) {
+	const type = response && response.error && response.error.type;
+	return type === "ConfirmationUnavailableError" || type === "ConfirmationOutcomeUnknownError";
 }
 
 // ── pending-card copy ─────────────────────────────────────────────────────────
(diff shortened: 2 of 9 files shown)


Navin-S-R:
## Review findings — `jarvis/chat/pending_confirm.py`

The core change looks sound: replacing `GETDEL` with a `MULTI/EXEC` GET+DEL transaction
preserves the single-winner guarantee on Redis < 6.2, and the reasoning in the docstring
holds up. All four findings below are in the surrounding error-handling paths, ordered
most severe first.

---

### 1. `list_for_owner`: blanket `except` empties the entire resync response — ~L384

```python
try:
    members = {... frappe.cache().smembers(_owner_key(owner)) ...}
except Exception:
    return []
```

A transient `TimeoutError`/`ConnectionError` on the index read makes **every** one of an
owner's live confirmation cards disappear from that resync, even though the records are
still in Redis with full TTL remaining.

What makes this worth fixing rather than tolerating: it contradicts the per-token loop
15 lines below, which deliberately refuses to prune on exactly this class of error
("Do NOT prune - the record may still be live"). The two paths in the same function
disagree about how to treat a transient read failure.

Suggested: narrow to `redis.exceptions.RedisError` and let it propagate (or return a
distinguishable error) so the SPA can retry, rather than rendering "you have no pending
confirmations".

---

### 2. `mint`: a blip on the verification read rolls back a *successful* park — ~L183

```python
if peek(token) is None:
    raise RuntimeError("pending-confirm record did not persist")
```

`peek()` → `_read_record()` swallows `RedisError` to `None`, so this check cannot
distinguish "the record never persisted" from "the verification GET itself failed".

If `set_value()` and `sadd()` both succeed and the record really is in Redis, a one-off
timeout on the read-back sends us into the `except` branch, which deletes the
just-written record and `srem`s the index entry. A park that actually worked is
discarded.

This self-heals — the caller surfaces a retryable error and the model re-runs the tool —
so it is not data loss. But the rollback destroys valid state in response to a read-path
failure.

Suggested: call `_read_record(token, swallow=False)` here and treat a raised
`RedisError` differently from a `None` return; only `None` means "did not persist".

---

### 3. `consume`: a Redis read failure is indistinguishable from expiry, and logs nothing — ~L307

```python
record = _read_record(token)   # default swallow=True
if not record:
    return None
```

Any `RedisError` becomes a plain `None`, which is the same signal as "token expired",
and nothing is logged.

During a sustained Redis degradation every Confirm click returns `None`; users get
generic "invalid/expired" toasts on cards that are actually live, and on-call has no way
to tell a tenant-wide incident from ordinary expiry.

Note this is inconsistent within the same function — the get-and-delete burn failure a
few lines below *does* `frappe.logger(...).error(...)` for the same error class.

Suggested: mirror that handler — log at `.error()` before returning `None`.

---

### 4. `list_for_owner`: the per-token transient-blip skip logs nothing — ~L398

```python
except redis.exceptions.RedisError:
    continue
```

The decision not to prune here is right and well commented, but it leaves no trace when
it fires. A user whose card fails to re-surface after a reload sees it simply missing,
and there is no signal separating "transient blip masking live cards" from normal
operation.

Suggested: one `frappe.logger(...).warning(...)` with the token id, same as (3).

---

<sub>Automated review pass; each finding was manually re-checked against the PR head
before posting. Two additional candidates were discarded as false positives, and one
(the extra Redis round trip added to `mint`'s hot path) was left out deliberately — it
is a stated trade-off, since `set_value` suppresses transient errors and the read-back
is the only way to confirm the record landed.</sub>

Vigneshsekar:
Review follow-up is pushed in commits 7b500aa9 and 53ad2f54. The owner-index and per-record read paths now log and return a typed storage error to strict resync callers instead of an empty/partial success. Mint uses strict raw Redis verification and retries one transient read once before rollback. Confirm and discard no longer report storage failures as expiry. Consume now uses a request-specific durable claim inside WATCH/MULTI/EXEC, so a lost EXEC reply can be recovered without allowing a second winner; an unrecoverable outcome stops before business dispatch. Logs intentionally omit the raw confirmation token because it is a bearer secret. The PR description now contains the full validation record.

PR #611, opened 2026-08-04
fix(wiki): voice-note ingest appends instead of replacing, so long pages survive

### LINKED ISSUE #488: wiki: voice-note ingest shows the model a truncated body then replaces the whole page, destroying everything past 3,000 chars
## Summary

The voice-note wiki ingest shows the model a **truncated** copy of a page body, then instructs it to return the **full** body, then **replaces** the stored field with the reply. Any content past 3,000 characters is destroyed on every ingest that touches the page. The model cannot preserve text it was never given.

Severity: high. Deterministic, silent, on a shipped mainline path triggered by an ordinary user action.

## Mechanism

Three individually reasonable steps compose into data loss.

1. `jarvis/chat/wiki.py:98` `_MAX_EXISTING_BODY_PROMPT_CHARS = 3000`
2. `jarvis/chat/wiki.py:775` in `_pages_for_prompt`:
   ```python
   r["body_md"] = (r.get("body_md") or "")[:_MAX_EXISTING_BODY_PROMPT_CHARS]
   ```
   A plain slice with no truncation marker, so the model has no signal that the body continues.
3. `jarvis/chat/wiki.py:119-121` `_INGEST_SYSTEM`:
   > `"body_md"` must be the FULL updated body: the existing body with the new durable knowledge merged in, never drop existing content.
4. `jarvis/chat/wiki.py:1112` `_merge_update_into_page`, non-contradiction branch:
   ```python
   doc.body_md = _clip_body(incoming)
   ```
   A wholesale replace. There is no length comparison, no diff, no minimum-retention check. `_clip_body` (`wiki.py:231-242`) enforces only the 20,000 upper bound, and the controller (`jarvis_wiki_page.py:213-218`) validates only maxima.

The safe branch exists but is unreachable. `append_md` at `wiki.py:1103-1104` appends without loss, but `_INGEST_SYSTEM` says "exactly these keys" and never lists `append_md`, so the ingest model is never told it exists.

## Reachability

Live mainline, not dead code.

- `frontend/src/views/ChatView.vue:8089-8102` `saveNudgeNote` posts a Conversation-context note from the shipped chat wiki-nudge card.
- `jarvis/chat/voice_notes_api.py:111-174` `save_voice_note` inserts it and enqueues `wiki.enqueue_ingest_note(doc.name)` at `:164-170` for every Conversation-context note. `_SOURCES` at `:38` also includes "Mobile".
- `jarvis/hooks.py:353` schedules `voice_facts.process_daily`, whose `_sweep_conversation_notes` (`voice_facts.py:793-823`) re-enqueues up to 50 New notes per run as a backstop.
- `wiki_enabled` defaults ON via the NULL=ON idiom (`wiki.py:134-146`).

## Failing scenario

1. An SM writes a 12,000 character `customer--acme-corp` body through `save_wiki_page` (payment terms, escalation contacts, SLA history). `save_wiki_page` sets `doc.body_md` unclipped and is valid to 20,000 (`wiki.py:1652-1653`, `jarvis_wiki_page.py:36`).
2. Weeks later a user chatting about Acme taps the nudge card: "Acme wants consolidated monthly invoices now."
3. `_pages_for_prompt` hands the model the first 3,000 characters, unmarked.
4. The model returns its "full merged body": roughly those 3,000 characters, paraphrased, plus one new bullet, with `contradiction: false`.
5. `wiki.py:1112` replaces the field. About 9,000 characters are gone.

Blast radius per note is `MAX_PAGES_PER_NOTE = 5` (`wiki.py:95, 940`). The note is then marked Processed with "N page update(s) applied" (`wiki.py:718-731`) and `last_confirmed_at` is refreshed, so the page reads as freshly reviewed and nothing alerts anyone.

Realistic paths to a body over 3,000 characters: human edits through `save_wiki_page`, monotonic contradiction appends (`wiki.py:1109`), promotion appends (`wiki.py:1347, 1380`), repeated ingest merges, and shared doctype-level pages like `doctype--sales-invoice` that every conversation touching that doctype writes into.

Note for scoping: `org-notes--<domain>` pages are **not** exposed to this prompt, because `_pages_for_prompt` derives slugs only from the note's ERP entities via `page_ref_for` (`jarvis/chat/entities.py:112-140`), which emits only `<pagetype>--<name>` and `doctype--<doctype>` forms. The deterministic targets are entity pages and doctype pages.

## Recovery

Partial. `jarvis_wiki_page.json:173` sets `track_changes: 1`, the merge uses `doc.save()` with no `ignore_version`, and Frappe writes the full old field value to a Version row. So the old body survives in `tabVersion`, but recovery is manual, admin-only, and unprompted.

## Why no test caught it

`jarvis/tests/test_wiki.py:209-219` asserts the destructive replace as correct behaviour:

```python
# Non-contradicting update: body_md is the merged replacement.
self.assertIn("45-day terms", doc.body_md)
self.assertNotIn("60-day terms", doc.body_md)
```

That assertion is right about the unit: the documented merge contract is "replace". The defect is in the composition. No test anywhere references `_pages_for_prompt` or `_MAX_EXISTING_BODY_PROMPT_CHARS`, and no test exercises a body over 3,000 characters.

## Fix directions

Any one of these closes it; they are not mutually exclusive.

- Have the ingest emit `append_md` for existing pages and add the key to `_INGEST_SYSTEM`, making the lossless branch the default.
- Refuse a non-contradiction `body_md` replace when the incoming body is materially shorter than the stored one, and fall back to append.
- Pass the full body when it fits the model budget, and when it does not, mark the truncation explicitly and forbid replace for that page.

Filed from a review of the Wiki feature. Related: the human-edit fence gap filed separately.

kavin-114:
## Summary

The voice-note wiki ingest no longer deletes the part of a page it never showed the model: a page over 3,000 characters now survives an ingest intact. Anyone who records a chat nudge note against an entity that has a long wiki page notices, because that page used to lose everything past 3,000 characters, silently, on every ingest.

I chose to make the ingest APPEND to an existing page rather than to keep the replace behind a "materially shorter" guard. The deciding fact is that the replace contract is not merely truncation-buggy, it is arithmetically impossible: the reply is capped at `max_tokens=4000` (roughly 16,000 characters, shared across up to 5 pages) while `MAX_BODY_LEN` is 20,000, so the model can never hand a long body back intact even if it were shown the whole thing. A shrink guard would then fire on essentially every long page and still have to fall back to append, so it buys a heuristic and no extra safety.

Appending does not break corrections. On this path a correction is already an append: `contradiction: true` appends a `## Contradiction flagged` section and sets `contradiction_flag`, and that branch is untouched. The non-contradiction branch had no correction remit at all, the prompt told it "never drop existing content", so replace only ever served tidiness.

Server side is the enforcement, the prompt is only the request: `allow_body_replace=False` makes a stray `body_md` against an existing page append instead of replace, so a model that ignores the prompt still cannot delete anything.

<details>
<summary>Root cause, the three steps that composed into loss</summary>

1. `wiki.py:98` `_MAX_EXISTING_BODY_PROMPT_CHARS = 3000`, and `_pages_for_prompt` sliced the body with no marker, so the model read a slice as the whole page.
2. `_INGEST_SYSTEM` listed exactly 8 keys with no `append_md` among them, and demanded `body_md` be "the FULL updated body ... never drop existing content". The lossless `append_md` branch at `wiki.py:1099` existed but was unreachable because the model was never told the key existed.
3. `_merge_update_into_page` did `doc.body_md = _clip_body(incoming)`. No length comparison, no minimum-retention check. `_clip_body` and the controller's `_validate_lengths` enforce only the 20,000 upper bound.

Entry path: ChatView nudge card, `voice_notes_api.save_voice_note`, `wiki.enqueue_ingest_note`, `_ingest_note`. `wiki_enabled` defaults ON via the NULL=ON idiom, and `voice_facts.process_daily` re-enqueues New notes as a backstop.

Changes:

- `_INGEST_SYSTEM`: one body key chosen by case. New page: `body_md`. Existing page: `append_md`, only the new knowledge, never the page body back. Contradiction: unchanged.
- `_body_for_prompt`: an over-budget body is cut on a line boundary and carries an explicit `[EXCERPT ONLY: N further characters ...]` marker.
- `apply_extracted_page_updates(allow_body_replace=True)`: new keyword, default True. Every existing caller is byte for byte unchanged, which matters for the app-learning scribe (`record_app_wiki`, `app_analysis`): it composes a page from a source it read in full, so its re-run must still replace its own page rather than double it. Only the ingest passes False.

Verified the regression test fails on unpatched `develop` with the real defect, a 3,979 character body reduced to 63 characters:

```
AssertionError: 'TAIL-SENTINEL-KEEP-ME' not found in
'## Payment\n\nAlways 60-day terms.\n\n- Consolidated monthly invoices.'
```

Local run, `bench --site site.jarvis run-tests --app jarvis --module jarvis.tests.test_wiki`: 25 + 29 tests, OK. Related modules also green: `test_app_learning` (44), `test_app_learning_agent` (59), `test_voice_facts` (21), `test_personalise_pipeline` (38), `test_wiki_scopes_api` (38), `test_confirm_card` (68). `test_wiki_graph.test_add_link_concurrency_no_lost_update` errors locally with MariaDB 1020 on a threaded `for_update` read, and it errors identically on unpatched `develop`, so it is pre-existing and unrelated.

Not fixed here, worth a follow-up: appends now accumulate, and `_clip_body` drops the OLDEST content once a body passes the hard 20,000 cap. That is the pre-existing policy on every append writer (`voice_facts`, the `update_wiki` tool, the contradiction branch) and 20,000 is enforced on human saves too, so it is a separate discussion, not part of this defect.

</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on :x:)
- [x] Branch is up to date with `develop` (branched from `upstream/develop` at 19471502)

Closes #488

### DIFF
--- a/jarvis/chat/wiki.py
+++ b/jarvis/chat/wiki.py
@@ -149,6 +149,8 @@ _BODY_EXCERPT_MARKER = (
 	"\n\n[EXCERPT ONLY: {n} further characters of this page are not shown. "
 	'Reply with "append_md" for this page, never a full body.]'
 )
+# How far back an excerpt may snap to end on a whole line.
+_EXCERPT_LINE_SNAP_CHARS = 200
 
 
 # --------------------------------------------------------------------------- #
@@ -811,8 +813,14 @@ def _body_for_prompt(body) -> str:
 	if len(body) <= _MAX_EXISTING_BODY_PROMPT_CHARS:
 		return body
 	head = body[:_MAX_EXISTING_BODY_PROMPT_CHARS]
+	# Snap back to a line boundary, but only a NEARBY one (the _clip_body idiom):
+	# rfind scans the WHOLE window, so a body whose only early newline is followed
+	# by one long unbroken run would otherwise be trimmed to almost nothing,
+	# spending the budget on the marker instead of on context.
 	nl = head.rfind("\n")
-	head = (head[:nl] if nl > 0 else head).rstrip()
+	if nl > len(head) - _EXCERPT_LINE_SNAP_CHARS:
+		head = head[:nl]
+	head = head.rstrip()
 	return head + _BODY_EXCERPT_MARKER.format(n=len(body) - len(head))
 
 
@@ -1160,24 +1168,33 @@ def _merge_update_into_page(
 	if not (doc.ref_name or "").strip() and update.get("ref_name"):
 		doc.ref_name = str(update["ref_name"]).strip()
 
+	# ``append_md`` still wins over ``body_md``, but which key carried the content
+	# no longer decides how a CONTRADICTION lands: a contradicting append used to
+	# slip past the flagged-section path and store contested knowledge as ordinary
+	# prose, leaving neither ``contradiction_flag`` nor the marker text that
+	# jarvis.learning.wiki_lint sweeps for. Only ``body_md`` may replace, and only
+	# when the caller is allowed to.
 	existing = (doc.body_md or "").strip()
+	incoming = ""
+	replaces = False
 	if isinstance(append_md, str) and append_md.strip():
-		doc.body_md = _clip_body(f"{existing}\n\n{append_md.strip()}".strip())
+		incoming = append_md.strip()
 	elif isinstance(body_md, str) and body_md.strip():
 		incoming = body_md.strip()
+		replaces = allow_body_replace
+	if incoming:
 		if contradiction and existing:
 			stamp = now_datetime().strftime("%Y-%m-%d")
 			doc.body_md = _clip_body(f"{existing}\n\n## Contradiction flagged ({stamp})\n\n{incoming}")
 			doc.contradiction_flag = 1
-		elif existing and not allow_body_replace:
-			# An append-only caller sent a body_md anyway (the ingest prompt asks
-			# for append_md, but the model is not bound by it). It only ever saw
-			# an EXCERPT of this page, so swapping the field would delete the
-			# rest. Append instead: a duplicated section is recoverable by a
-			# human editor, a deleted one is not (issue #488).
-			doc.body_md = _clip_body(f"{existing}\n\n{incoming}")
-		else:
+		elif replaces or not existing:
 			doc.body_md = _clip_body(incoming)
+		else:
+			# Either an append, or an append-only caller that sent a body_md
+			# anyway. The ingest only ever saw an EXCERPT of this page, so
+			# swapping the field would delete the rest; a duplicated section is
+			# recoverable by a human editor, a deleted one is not (issue #488).
+			doc.body_md = _clip_body(f"{existing}\n\n{incoming}")
 	append_source(doc, source, ref, user)
 	doc.last_confirmed_at = now_datetime()
 	doc.save(ignore_permissions=True)
--- a/jarvis/chat/wiki.py
+++ b/jarvis/chat/wiki.py
@@ -30,9 +30,10 @@ writes behind explicit channel checks + the controller sanitizer).
 Merge discipline (``apply_extracted_page_updates``): ``append_md`` appends,
 ``body_md`` replaces only when the update carries no contradiction; a flagged
 contradiction APPENDS a ``## Contradiction flagged (<date>)`` section and sets
-``contradiction_flag`` — extracted content never silently overwrites contested
-knowledge. Every applied update appends a ``{date, kind, ref, user}`` sources
-entry and refreshes ``last_confirmed_at``.
+``contradiction_flag`` WHICHEVER key carried it — extracted content never
+silently overwrites contested knowledge, and never buries it as ordinary prose
+where ``jarvis.learning.wiki_lint`` cannot find it. Every applied update appends
+a ``{date, kind, ref, user}`` sources entry and refreshes ``last_confirmed_at``.
 
 The voice-note ingest is APPEND-ONLY against pages that already exist
 (``allow_body_replace=False``). It shows the model at most
--- a/jarvis/chat/wiki.py
+++ b/jarvis/chat/wiki.py
@@ -33,6 +33,14 @@ contradiction APPENDS a ``## Contradiction flagged (<date>)`` section and sets
 ``contradiction_flag`` — extracted content never silently overwrites contested
 knowledge. Every applied update appends a ``{date, kind, ref, user}`` sources
 entry and refreshes ``last_confirmed_at``.
+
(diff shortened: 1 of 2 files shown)


kavin-114:
### Review round, two extra defects fixed

A review pass over the first commit found two real defects, both now fixed with tests in `bdcd4867`.

1. **A contradicting append was stored unflagged.** `_merge_update_into_page` only inspected `contradiction` inside the `body_md` branch. Asking the ingest for `append_md` made the `append_md` + `contradiction: true` shape reachable for the first time, and it stored contested knowledge as ordinary prose: no `contradiction_flag`, no `## Contradiction flagged` marker. `jarvis/learning/wiki_lint.py:92` sweeps on exactly those two signals, so the backstop could not see it either. The merge now honours `contradiction` whichever key carried the content. No existing caller sends `append_md` with `contradiction` set, so nothing else changes. Verified the new test fails on the previous commit: `AssertionError: '## Contradiction flagged (' not found in '## Payment\n\nAlways 60-day terms.\n\nThey now prepay everything.'`

2. **The excerpt could collapse to almost nothing.** `_body_for_prompt` snapped to a line boundary with `rfind` over the whole window, so a body whose only early newline was followed by one long unbroken run got trimmed to a few characters, spending the budget on the marker instead of on context (measured: 3000 chars of context down to 5). It now snaps back at most `_EXCERPT_LINE_SNAP_CHARS`, the idiom `_clip_body` already uses.

Local: `jarvis.tests.test_wiki` 26 + 30 tests OK. Also green: `test_app_learning` (44), `test_app_learning_agent` (59), `test_voice_facts` (21), `test_personalise_pipeline` (38), `test_wiki_scopes_api` (38).

Also verified live against the running `jarvis.proxy` site, driving the real write path with only the LLM reply stubbed: a 3,644 char page came out at 3,712 chars with its tail intact, and the excerpt shown to the model carried `[EXCERPT ONLY: 661 further characters ...]`.

Two pre-existing sharp edges found while reviewing, NOT fixed here (they predate this PR and deserve their own issues):

- An update whose page does not resolve and which lacks `title`/`page_type` is dropped by `_apply_one_update` returning `False`, which counts as "refused", not "failed". `_ingest_note` only retries on `failed`, so the note is marked Processed and the knowledge is gone with no log line.
- Appends now accumulate, and `_clip_body` drops the OLDEST content once a body passes the hard 20,000 cap. That is the existing policy for every append writer, but this change makes pages reach the cap sooner.

Issue #454, opened 2026-07-26
agents: finding dedupe and auto-resolve query the wrong owner, so both are dead when reviewer != owner


kavin-114:
## Summary

`record_delegate_run` writes each Finding under one owner and reads it back under a different one. When an installation's `reviewer` differs from its `owner`, finding dedupe and A16 auto-resolve both stop working silently.

## Mechanism

The write re-homes the row to the visibility owner:

```python
# jarvis/chat/agent_runs.py:749
_reassign_owner(FINDING, fd.name, visibility_owner)
```

`_visibility_owner(inst)` (agent_runs.py:304-314) returns `inst.reviewer or inst.owner` while the install is shadow.

Both reads filter on `inst.owner` instead:

```python
# jarvis/chat/agent_runs.py:691-695  (dedupe)
existing = frappe.db.get_value(FINDING, {
    "owner": owner,        # inst.owner, never the reviewer
    "agent": agent, "fingerprint": fp, "state": "open",
}, "name")

# jarvis/chat/agent_runs.py:790-799  (auto-resolve candidates)
candidates = frappe.get_all(FINDING, filters={
    "owner": owner,        # same problem
    "agent": agent, "state": "open", "rule_id": ["in", list(evaluated_tokens)],
})
```

At insert the row owner is the session user, which is the run-as user (the `record_agent_run` tool runs impersonated). `_reassign_owner` no-ops only when that already equals the visibility owner. Either way the final row owner is the visibility owner, so with a distinct reviewer neither read can ever match its own rows.

## Impact

With reviewer R different from owner X on a shadow install:

1. Run 1 records finding F, row lands owned by R.
2. Run 2 re-detects F, dedupe searches for owner X, finds nothing, inserts a duplicate.
3. Repeat every run. A daily auditor accumulates roughly 30 duplicate rows per finding per month. `findings_count` inflates with them.
4. Auto-resolve never finds candidates, so a finding that stops being detected stays open forever.
5. On promotion, every duplicate re-homes to the owner's surface as an open row.

## Fix

Use the same helper the write already uses, in both queries:

```python
"owner": _visibility_owner(inst),
```

Or drop `owner` from the dedupe key entirely, since the installation is already unique per (owner, agent).

## Test gap

[...]

PR #612, opened 2026-08-04
fix(agents): scope the uninstall finding cascade to its own installation

### LINKED ISSUE #455: agents: uninstall_agent force-deletes findings belonging to other owners
## Summary

`uninstall_agent` selects findings to delete by a broad `(agent, owner)` match rather than by installation membership, then hard-deletes them with `force=True`. It can destroy findings belonging to a different owner.

## Mechanism

```python
# jarvis/chat/agents_api.py:785-792
finding_names = set(frappe.get_all(FINDING, filters={
    "agent": doc.agent,
    "owner": ["in", [doc.owner, doc.reviewer]],
}, pluck="name", ignore_permissions=True))
...
# jarvis/chat/agents_api.py:800-801
frappe.delete_doc(FINDING, name, ignore_permissions=True, force=True)
```

The run-pointer clause at :793-799 only does `finding_names.update(...)`, so it widens the set, never narrows it.

Nothing constrains `reviewer` across installations. The `(owner, agent)` uniqueness rule is per owner only, and PP-6 explicitly institutionalises one reviewer covering multiple installations (`agents_api.py:997-1026`).

The code directly above knows about this exact hazard. `_rehome_installation_outputs` (`agents_api.py:1050-1052`) locates rows "precisely by installation membership (never a broad owner+agent match that could sweep a different install of the same agent)". `uninstall_agent` does the thing that comment warns against.

## Impact

Simplest reachable case, no second shadow install required:

- Agent G. User X installs it and names R as reviewer.
- R also has their own install of G, live, so R's findings are owned by R.
- X clicks Uninstall. The filter matches `(agent=G, owner in [X, R])` and force-deletes R's own findings.

Second case: Y also has an install of G with reviewer R, in shadow, so Y's findings are owned by R. X's uninstall destroys Y's audit history for G.

Both are irreversible. `ignore_permissions=True` plus `force=True` means no permission check and no link check can intervene, and `jarvis_agent_finding.py` has no `on_trash` guard.

## Fix

Scope the cascade to installation membership, the way `_rehome_installation_outputs` already does: collect this installation's runs first, then select findings whose `run` / `first_seen_run` / `last_seen_run` land in that set. Drop the `(agent, owner)` clause entirely.

## Verification

Independently verified. Confirmed there is no other scoping, no `on_trash` guard, and no uniqueness rule that would prevent the shared-reviewer configuration.

kavin-114:
## Summary

Uninstalling an agent now deletes only the findings that installation's own runs produced, instead of every finding matching `(agent, owner in [owner, reviewer])`. Any customer who shares a reviewer with another customer notices: their audit history used to be force-deleted when the other customer clicked Uninstall.

`uninstall_agent` picked findings by a broad owner match and destroyed them with `force=True, ignore_permissions=True`. A reviewer can legitimately be the reviewer-of-record for many installations (PP-6 institutionalises exactly that), and PP-4 re-homes a shadow installation's findings to that reviewer, so the filter reached across customers. The delete was irreversible and no permission check or link check could intervene.

Membership now comes from the finding's creation stamps, `run` and `first_seen_run`, which are written once at insert and frozen by the controller. That is the rule `_rehome_installation_outputs` already applies to the same rows.

`last_seen_run` is deliberately not a membership signal, and findings with no surviving run pointer are left alone rather than swept. Reasoning for both is in the details below. A `on_trash` guard on the finding doctype re-derives membership independently, so a caller cannot destroy a foreign finding even with `force` and `ignore_permissions`.

Fixes #455.

<details>
<summary>Root cause, scoping rule, and the two judgement calls</summary>

### Root cause

`jarvis/chat/agents_api.py` (before):

```python
finding_names = set(frappe.get_all(FINDING, filters={
    "agent": doc.agent,
    "owner": ["in", [doc.owner, doc.reviewer]],
}, pluck="name", ignore_permissions=True))
...
frappe.delete_doc(FINDING, name, ignore_permissions=True, force=True)
```

The run-pointer clause below it only ever called `.update()`, so it widened the set and never narrowed it. Nothing constrains a reviewer to a single installation: the uniqueness rule is per `(owner, agent)`, and `_verify_reviewer_two_pack_capacity` requires a reviewer to span two packs before a second live module is allowed. Reachable case with no exotic setup: user X installs agent G naming R as reviewer, R also runs their own install of G, X uninstalls, R's own findings are destroyed.

### Scoping rule

A finding belongs to this installation iff `run` or `first_seen_run` is one of this installation's runs. Both are stamped at insert by `agent_runs.record_delegate_run` and listed in the controller's `_FROZEN_FIELDS`, so they are the one durable statement of which installation produced the row.

### Why `last_seen_run` is excluded

It is the only run pointer the engine ever re-points. The recurrence bump that re-points it matches on `(owner, agent, fingerprint)`, and under PP-4 shadow re-homing that owner can be a shared reviewer, so it can attach a different owner's finding to our run. That is the #454 defect. A row reached only through `last_seen_run` was, by construction, created by somebody else's installation, so including it could only ever re-open the sweep this PR closes.

That leaves a surviving foreign row pointing at a run we are about to force-delete. `get_findings` INNER JOINs `last_seen_run` for the run drill-down, and Frappe link validation would reject that owner's next acknowledge or resolve save, so `_detach_last_seen_run` falls the pointer back to the row's own `first_seen_run` before the runs go.

### Why orphans are left behind

A finding whose run pointers are empty, or whose runs were already deleted, cannot be attributed. There is no evidence tying it to this installation rather than to any other install of the same agent, and guessing is what caused this bug. An orphan is visible and removable from Desk. Another customer's destroyed audit history is not. Rows already lost to the old code are not recoverable by this PR.

### Verification

The four new tests fail on unpatched `develop` and pass with the fix:

```
 ✖  test_cross_install_recurrence_bump_is_detached_not_deleted   AssertionError: None is not true
 ✖  test_finding_with_no_run_pointer_is_left_alone               AssertionError: None is not true
 ✖  test_on_trash_refuses_a_foreign_finding_inside_a_cascade     AssertionError: PermissionError not raised
 ✖  test_uninstall_spares_a_co_reviewed_owners_findings          AssertionError: None is not true
Ran 13 tests in 0.837s
FAILED (failures=4)
```

With the fix, plus every other local module that touches findings or runs:

```
jarvis.tests.test_platform_agents_api_hardening   Ran 13 tests   OK
jarvis.tests.test_platform_min_apps               Ran 14 tests   OK
jarvis.tests.test_agents_marketplace              Ran 36 tests   OK
jarvis.tests.test_agent_dashboard_wire            Ran  7 tests   OK
jarvis.tests.test_agent_identity                  Ran  7 tests   OK
jarvis.tests.test_app_learning_agent              Ran 59 tests   OK
jarvis.tests.test_list_registry                   Ran 16 tests   OK
jarvis.tests.test_part3_security                  Ran 19 tests   OK
jarvis.tests.test_platform_activation             Ran 13 tests   OK
jarvis.tests.test_platform_capability_contract    Ran 38 tests   OK
jarvis.tests.test_platform_false_clean            Ran  8 tests   OK
jarvis.tests.test_platform_launch_provenance      Ran 11 tests   OK
jarvis.tests.test_platform_render_gate            Ran  9 tests   OK
jarvis.tests.test_platform_schema                 Ran 17 tests   OK
jarvis.tests.test_platform_shadow_dashboard       Ran  2 tests   OK
jarvis.tests.test_platform_write_caps             Ran 12 tests   OK
jarvis.tests.test_platform_writeback              Ran 30 tests   OK
```

### Relationship to #454

Complementary, not conflicting. #454 fixes the dedupe and auto-resolve queries in `agent_runs.py`, which is where the cross-owner `last_seen_run` bump originates. This PR does not touch `agent_runs.py`. Once #454 lands, `_detach_last_seen_run` becomes a no-op for new data and stays useful for rows already bumped. `test_cross_install_recurrence_bump_is_detached_not_deleted` constructs the bump directly with `db.set_value` rather than by driving a run, so it does not depend on the buggy behaviour and will keep passing afterwards.

</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on x)
- [x] Branch is **up to date with `develop`** (branched from `upstream/develop` at 19471502)
- [x] New/changed behavior has tests (5 new tests, 4 of which fail on unpatched develop)

### DIFF
--- a/jarvis/chat/agents_api.py
+++ b/jarvis/chat/agents_api.py
@@ -751,20 +751,89 @@ def set_run_as_user(installation: str, user: str) -> dict:
 	}
 
 
+def _installation_finding_names(run_names: list) -> set:
+	"""#455 — the findings an uninstall cascade may destroy: ONLY those this
+	installation's own runs CREATED.
+
+	Membership is read off the finding's creation stamps, ``run`` and
+	``first_seen_run``. Both are written once at insert
+	(``agent_runs.record_delegate_run``) and are frozen by the finding controller
+	thereafter, so they are the one durable statement of which installation
+	produced the row.
+
+	``last_seen_run`` is deliberately NOT a membership signal even though it too
+	points at a run. It is the only run pointer the engine ever re-points, and the
+	recurrence bump that re-points it matches on ``(owner, agent, fingerprint)``.
+	Under PP-4 a shadow installation's findings are re-homed to the REVIEWER, so a
+	reviewer who also runs their own installation of the same agent can have a
+	DIFFERENT owner's finding bumped onto one of their runs (#454). Treating that
+	as membership would re-open exactly the cross-owner sweep this function
+	exists to close: a bumped row was, by construction, created by somebody else's
+	installation.
+
+	The old ``{"agent": ..., "owner": ["in", [owner, reviewer]]}`` filter is gone
+	entirely. Nothing constrains a reviewer to a single installation (PP-6
+	institutionalises the opposite, see ``_verify_reviewer_two_pack_capacity``), so
+	that clause matched — and ``force=True`` destroyed — live findings belonging to
+	other customers.
+
+	Findings with NO surviving run pointer (a pre-run-tracking row, or one whose
+	runs were already deleted) are left ALONE. There is no evidence tying such a row
+	to this installation rather than to any other install of the same agent, and
+	guessing is what caused #455. An orphan is visible and removable from Desk;
+	another customer's destroyed audit history is neither.
+	"""
+	if not run_names:
+		return set()
+	names = set()
+	for field in ("run", "first_seen_run"):
+		names.update(
+			frappe.get_all(FINDING, filters={field: ["in", run_names]}, pluck="name", ignore_permissions=True)
+		)
+	return names
+
+
+def _detach_last_seen_run(run_names: list) -> None:
+	"""Re-point any SURVIVING finding whose ``last_seen_run`` lands in the runs this
+	cascade is about to force-delete.
+
+	A survivor here is by definition another installation's row that the #454
+	recurrence bump attached to one of our runs. Leaving the pointer would dangle
+	it: ``get_findings``' run drill-down INNER JOINs ``last_seen_run``, so the row
+	would vanish from its real owner's history, and Frappe's link validation would
+	reject that owner's next acknowledge/resolve save outright. Falling back to the
+	row's own ``first_seen_run`` (which belongs to ITS installation and survives, or
+	is empty) keeps the recurrence pointer truthful and the row usable.
+
+	Raw ``db.set_value`` — ``last_seen_run`` is a frozen audit field, so this must
+	bypass the finding controller exactly as the recurrence bump does.
+	"""
+	if not run_names:
+		return
+	for row in frappe.get_all(
+		FINDING,
+		filters={"last_seen_run": ["in", run_names]},
+		fields=["name", "first_seen_run"],
+		ignore_permissions=True,
+	):
+		frappe.db.set_value(
+			FINDING, row.name, "last_seen_run", row.first_seen_run or None, update_modified=False
+		)
+
+
 @frappe.whitelist()
 def uninstall_agent(installation: str) -> dict:
 	"""Delete an installation (owner-gated) plus its run + finding history —
 	bottom-up, mirroring ``macros_api.delete_macro``: findings link runs (via
 	``run`` / ``first_seen_run`` / ``last_seen_run``) and runs link the
-	installation, so leaving either behind would block the delete with
-	LinkExistsError. One install is one (owner, agent) pair, so the owner's
-	findings for the agent go, PLUS (belt-and-braces) any finding whose run
-	pointers land in this install's runs. The ``uninstalled`` activity row is
-	written FIRST — it is Link-free by design, so the history survives the
-	cascade. The bundle leaves the container on the next Apply (the fleet
-	endpoint does a full reconcile); the dirty flag records that an Apply is
-	now pending — but only when the install was ENABLED (a disabled install
-	was never in the pushed set, so removing it changes nothing)."""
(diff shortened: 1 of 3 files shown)


kavin-114:
## Live verification on a running bench

Not a unit test. This seeds the issue's shared-reviewer topology on a real site and calls the same whitelisted endpoint the SPA's Uninstall button calls (`agents_api.uninstall_agent`), first against the unpatched `develop` checkout, then against this branch. Same script, same site, same fixture. Synthetic rows only, cleaned up after (verified 0 leftover rows in all five doctypes).

Fixture: Alice and Bob each install the same agent and both name Rita as reviewer. Rita also runs her own install of it. All three are shadow, so PP-4 re-homes every finding to Rita, which is what made the old `owner in [owner, reviewer]` filter reach across customers.

**A. Unpatched `develop`**

```
--- BEFORE -------------------------------------------------
  alice  owner=e2e455-rita@example.com      Alice: 2B mismatch 1,20,000
  bob    owner=e2e455-rita@example.com      Bob: 2B mismatch 4,80,000
  rita   owner=e2e455-rita@example.com      Rita: 2B mismatch 90,000

--- Alice clicks Uninstall (real endpoint) ------------------
  agents_api.uninstall_agent -> {'ok': True}

--- AFTER --------------------------------------------------
  alice  finding  DESTROYED
  bob    finding  DESTROYED
  rita   finding  DESTROYED
  bob install    ALIVE
  rita install   ALIVE
  alice install  DESTROYED

  VERDICT: FAIL - another owner's findings were destroyed
```

Bob's and Rita's installations survive with their finding history silently gone.

**B. This branch**

```
--- AFTER --------------------------------------------------
  alice  finding  DESTROYED
  bob    finding  ALIVE
  rita   finding  ALIVE
  bob install    ALIVE
  rita install   ALIVE
  alice install  DESTROYED

  VERDICT: PASS - only Alice's own history was removed
```

Alice's own cascade still works end to end. Nobody else is touched.

Issue #460, opened 2026-07-26
agents: install uniqueness is check-then-act with no DB constraint


kavin-114:
## Summary

Uniqueness of (owner, agent) on Jarvis Agent Installation is enforced only by check-then-act, with no DB constraint behind it. Two concurrent requests can both pass and both insert.

## Mechanism

```python
# jarvis/jarvis/doctype/jarvis_agent_installation/jarvis_agent_installation.py:190-200
clash = frappe.db.exists("Jarvis Agent Installation",
    {"owner": owner, "agent": self.agent, "name": ["!=", self.name or ""]})
if clash:
    frappe.throw(_("You have already installed this agent."))
```

`agents_api.install_agent:567` repeats the same `frappe.db.exists` check.

The doctype JSON uses `naming_rule: Random / autoname: hash` and declares zero `unique` fields, so no naming collision and no DB constraint backstops the check. `install_agent` holds no lock and the DB runs REPEATABLE-READ, so two concurrent inserts both see no clash and both commit under distinct hash names.

## Impact

Low. Same user double-submits (double click, or two gunicorn workers). Both `exists()` calls return None before either commits, and two rows exist for one (owner, agent).

Consequence today is cosmetic: inflated `install_count`, and uninstalling once leaves the phantom row so the agent still renders as installed.

## Note on prior incident

This defect is **not** what caused the #434 outage, contrary to an earlier read. #434 was `build_agent_push_payload` being bench-global and keyed by slug while installs are per-(owner, agent): two *different* users each legitimately installing and enabling the same agent emitted the same slug twice, and admin rejected the duplicate-slug payload, killing every push. No race and no duplicate (owner, agent) row was involved. #434's slug de-dupe with union semantics also means a race-produced duplicate row can no longer brick the push.

## Fix

Outsource the integrity to the database. Add a composite unique index on `(owner, agent)` via `on_doctype_update` (not a patch, per #370) and let the insert fail on the constraint, keeping the friendly `frappe.throw` as the fast path.

kavin-114:
Fixed by #647, which is merged and whose merge commit is an ancestor of `develop` (verified with `git merge-base --is-ancestor`, not just from the PR's merged state).

Closing by hand because `develop` is not this repo's default branch, so a `Fixes #` reference in a PR merged into it never auto-closes the issue. That is why this sat open after the fix shipped.

PR #623, opened 2026-08-04
perf(agents): add missing indexes on agent run/finding tables

### LINKED ISSUE #459: agents: no indexes on any agent table and no on_doctype_update to add them
## Summary

No agent table carries an index on any column its hot queries filter or sort on, and no `on_doctype_update` exists in any agent doctype to add one. Not a present hazard, but it will not fix itself and the indexes will never appear on CI or a fresh customer bench without code.

## Measured

`SHOW INDEX` on jarvis.proxy:

| Table | Indexes |
|---|---|
| `tabJarvis Agent Finding` | PRIMARY, creation, modified |
| `tabJarvis Agent Run` | PRIMARY, creation, session_key_index |
| `tabJarvis Agent Installation` | PRIMARY, creation, modified |

Frappe 16.24.1 does not auto-index Link fields or `owner`. `frappe/database/schema.py:105` sets `set_index=field.get("search_index")`, so only `search_index` and `unique` flags create indexes. The agent doctype JSONs carry `search_index` on exactly one field, `session_key` on Run, which matches what is there.

There is no `on_doctype_update` in any agent doctype (the app's only one is `jarvis_user_model_usage.py:19`). Existing index patches (`v1_7_chat_message_page_index`, `v2_04_chat_turn_admission_indexes`) touch only Jarvis Chat Message and Jarvis Chat Turn.

## Hot queries that scan

- `agents_api.py:1452-1454` runs four `frappe.db.count` calls per Findings page load (total plus blocker, warning, note) filtering on `owner` and `severity`. Four full scans per page view.
- `agent_runs.py:691-695` does a `get_value` on `{owner, agent, fingerprint, state}` inside the per-finding loop. One scan per finding per run.
- `agents_api.py:1349` filters `owner` and orders by `started_at desc, creation desc` with `started_at` unindexed, forcing a filesort.
- The run drill-down at `agents_api.py:1432-1440` joins on unindexed `first_seen_run` and `last_seen_run`.

## Why this is low, and the one thing that changes that

Tables are empty today and growth is capped: 62 runs/month per install by default, an activation ceiling of 2 live modules, 20 installs per owner, and dedupe suppressing repeat rows. The ~100k-row threshold where page loads cross roughly 0.3 to 0.5 seconds of pure scan is years away on a per-customer site.

**Caveat: one of those caps is dedupe, and #454 breaks dedupe.** If #454 is not fixed, duplicate accumulation removes the cap this severity rating depends on.

## Fix

Add `on_doctype_update` to the three doctypes, not a patch. Per prior experience on #370, `install_app(set_as_patched=True)` marks patches done without running them, so a patch-only index is missing on CI and on every new customer bench.

Suggested minimum: `(owner, agent, state)` and `fingerprint` on Finding, `(owner, started_at)` and `installation` on Run, `(enabled, schedule_enabled, next_run_at)` on Installation.

kavin-114:
## Summary

Adds three composite indexes on the agent run/finding tables so owner-scoped reads stop doing table scans; customers with many agent runs or findings see faster Runs and Findings tab loads. Fixes #459.

Indexes go in `on_doctype_update()`, not a patch, matching the existing correct example (`jarvis_user_model_usage.py`): `install_app(set_as_patched=True)` marks patches done without running them, so a patch-only index would silently never exist on a fresh install.

**Jarvis Agent Run**
- `(owner, started_at, creation)`: serves `list_runs_page` (`agents_api.py:1349`), the Runs tab's only paging query, which filters `{"owner": me}` and always sorts `started_at desc, creation desc`.

**Jarvis Agent Finding**
- `(owner, agent, fingerprint)`: serves the per-finding dedupe lookup inside the per-run findings loop in `record_delegate_run` (`agent_runs.py:691`, N lookups per recorded run) and the same `(owner, agent)` prefix used by the "which runs observed this finding" drilldown join in `list_findings` (`agents_api.py:1432-1440`).
- `(owner, severity, state)`: serves the four `frappe.db.count` calls the Findings page load fires per request (`agents_api.py:1452, 1454`): one total count and three per-severity counts.

Out of scope: no unique index on `(owner, agent)` for Jarvis Agent Installation. That is #460 and needs a duplicate-row cleanup first, or the migration hard-fails on any bench already carrying the duplicates it's meant to prevent.

## Pre-merge checklist

- [x] New/changed behavior has tests (verified via a forced doctype resync against a real bench, see below; relevant test modules pass)

## Verification

Ran the real `on_doctype_update()` code path against a live bench (`site.jarvis`) by forcing a doctype resync, since the local site already has these doctypes installed and Frappe skips reimporting an unchanged JSON on a routine `bench migrate`, forcing is what a genuine fresh install does automatically. Confirmed with `SHOW INDEX` before and after, ran the resync a second time to prove idempotency (no errors, no duplicate indexes), then ran the agent-related test modules (`test_agents_marketplace`, `test_agent_dashboard_wire`, `test_agent_identity`, `test_platform_agents_api_hardening`, `test_app_learning_agent`), all green.


kavin-114:
Superseded, not abandoned: this branch's content is already on `develop`.

It was merged into `integration/backlog-2026-08-04` (along with the reload patch that makes the index actually get created on an existing site), and that branch merged to develop via #641. Verified on current develop: all four agent/macro doctypes define `on_doctype_update`, and both `v2_12_reload_*_doctypes_for_indexes` patches are registered in `patches.txt`.

The branch tip later absorbed an unrelated develop merge, so merging it now would drag in noise rather than add anything. Closing to keep the queue honest.

Issue #468, opened 2026-07-27
macros: scheduled steps bypass every usage cap, quota and subscription gate (~19k turns/month/user uncapped)


kavin-114:
## Summary

Scheduled macro steps never pass through the entitlement gate. Every per-user token cap, per-model cap and subscription-suspended check is bypassed, while the spend is still recorded against the owner. The owner's own chat then gets locked out by usage the macro consumed.

## Mechanism

`validate_can_send` has exactly three call sites, all human paths:
- `jarvis/chat/api.py:907`
- `jarvis/chat/api.py:1001` (`send_message`)
- `jarvis/chat/api.py:1710` (`retry_message`)

`_enqueue_turn` (`api.py:2119-2222`), which is the sole dispatch path for macro steps (`macros.py:342`, `macros.py:395`), has none. There is no equivalent gate in `dispatch.py`, `admission.py`, `turn_handler.py` or `worker.py`.

So a scheduled macro step bypasses:
- the per-user monthly token cap (`policy.py:140-173`)
- the per-model cap (`policy.py:76`)
- `subscription_suspended` (`policy.py:60`)

Usage IS still recorded at `turn_handler.py:1075-1079`. The inversion is complete: the macro drains the owner's quota, and the quota then rejects the human's interactive chat while the macro keeps running unimpeded.

## Scale

There is no run budget of any kind in `macro_scheduler.run_due_macros` (`macro_scheduler.py:20-76`).

Effective ceiling per owner:
- `MAX_MACROS_PER_OWNER = 25` and `MAX_STEPS = 25` (`jarvis_macro.py:19-21`)
- daily is the shortest frequency (`macro_scheduler.py:81-93`, no hourly option)
- 25 x 25 = **625 agent turns per day per user**, roughly **19,000 per month**, multiplied by the number of Jarvis users on the bench

Nothing reduces that. The only bound in the path is `admission.MAX_QUEUE_DEPTH = 25` (`admission.py:61`), which is per-shard backpressure, not a spend cap, and overflow parks the run in `waiting_capacity` and RETRIES up to 20 times rather than dropping it.

For comparison, the sibling agent scheduler caps at 62 runs per month per installation (`agent_scheduler._over_run_budget:885`, `_agent_run_budget_monthly:854-863`) with the stated rationale that scans must not drain the customer's own subscription.

## Fix

Two parts, both needed:

[...]

PR #624, opened 2026-08-04
perf(macros): add owner indexes on macro tables

### LINKED ISSUE #475: macros: no owner index on either macro table and no on_doctype_update to add one
## Summary

Neither macro table carries an index on `owner`, which every owner-scoped read filters on, and neither doctype has an `on_doctype_update`. Not a present hazard, but it will never appear on CI or a fresh customer bench without code.

## Measured

Live `information_schema` on jarvis.proxy:

| Table | Indexes | Row count today |
|---|---|---|
| `tabJarvis Macro` | PRIMARY, creation, modified | 0 |
| `tabJarvis Macro Run` | PRIMARY, creation, modified, macro, status | 0 |

`macro` and `status` are indexed because the doctype JSON sets `search_index` on them. `owner` is not indexed on either table.

## Hot queries

- `macros_api.list_macro_runs` (:490-517) filters `r.owner = %(owner)s` in raw SQL with `ORDER BY r.creation DESC`, plus a `COUNT(*)` under the same filter on every page load.
- `macros_api.macro_run_stats` (:532-540) does two owner-filtered scans.
- `macro_scheduler.run_due_macros` (:26) scans `{schedule_enabled: 1, next_run_at <= now}` hourly.

## Why this is low

Growth is well bounded. `MAX_MACROS_PER_OWNER = 25` and `MAX_STEPS = 25` (`jarvis_macro.py:19-21`) keep the Macro table trivially small forever, so the hourly scheduler scan is a non-issue. Frequencies are daily/weekly/monthly only (no hourly), so worst-case run growth is roughly 25 rows per user per day. Each site is single-tenant with few users, so `owner` is weakly selective anyway and the existing `creation` index already serves the `ORDER BY ... LIMIT` path.

The only unbounded cost is the per-pageload `COUNT(*)` and the stats scans once runs reach six figures, which is years out. Note there is no retention sweep for Jarvis Macro Run, so rows accumulate indefinitely.

This is equivalent to, arguably slightly better than, the agent-tables index finding rated low for the same reason.

## Fix

Add an `owner` index (or a composite `(owner, creation)`) to both doctypes via `on_doctype_update()`, NOT a patch. Neither controller has one today. Per the trap hit on #370, `install_app(set_as_patched=True)` marks patches done without running them, so a patch-only index is missing on CI and on every new customer bench.

Consider a retention policy for Jarvis Macro Run at the same time.

kavin-114:
## Summary

Adds owner-scoped composite indexes on both macro tables; customers with many macro runs see faster loads of the Macro runs dashboard (history list and summary tiles). Fixes #475.

Indexes go in `on_doctype_update()`, not a patch, matching the existing correct example (`jarvis_user_model_usage.py`): `install_app(set_as_patched=True)` marks patches done without running them, so a patch-only index would silently never exist on a fresh install.

**Jarvis Macro Run** (the table that actually grows, no per-owner cap)
- `(owner, creation)`: serves `list_macro_runs` (`macros_api.py:456`), which filters `r.owner = %(owner)s` and sorts `ORDER BY r.creation DESC`, and `macro_run_stats` (`macros_api.py:488`), whose `GROUP BY status WHERE owner = ...` and `MAX(creation) WHERE owner = ...` both scan owner-filtered rows.

**Jarvis Macro** (low priority, see below)
- `(owner, macro_name)`: serves the doctype's own `_validate_unique_per_owner` exists() check and `_validate_owner_cap` count, which fire on every macro save.

Honestly: `MAX_MACROS_PER_OWNER=25` caps any single owner's slice of `Jarvis Macro` permanently, so this index buys little for one tenant; its value is skipping every *other* tenant's rows as the tenant count grows, not speeding up a tiny per-owner scan.

Touches only the two doctype controllers. Left `macro_scheduler.py`, `macros.py`, and `hooks.py` alone (concurrent work on #468/#469/#471 lives there).

## Pre-merge checklist

- [x] New/changed behavior has tests (verified via a forced doctype resync against a real bench, see below; relevant test modules pass)

## Verification

Ran the real `on_doctype_update()` code path against a live bench (`site.jarvis`) by forcing a doctype resync, since the local site already has these doctypes installed and Frappe skips reimporting an unchanged JSON on a routine `bench migrate`, forcing is what a genuine fresh install does automatically. Confirmed with `SHOW INDEX` before and after, ran the resync a second time to prove idempotency (no errors, no duplicate indexes), then ran the macro-related test modules (`test_macro_merge`, `test_v3_list_extensions` including its direct `list_macro_runs` tests, `test_part3_security`, `test_chat_mining`), all green.


kavin-114:
Superseded, not abandoned: this branch's content is already on `develop`.

It was merged into `integration/backlog-2026-08-04` (along with the reload patch that makes the index actually get created on an existing site), and that branch merged to develop via #641. Verified on current develop: all four agent/macro doctypes define `on_doctype_update`, and both `v2_12_reload_*_doctypes_for_indexes` patches are registered in `patches.txt`.

The branch tip later absorbed an unrelated develop merge, so merging it now would drag in noise rather than add anything. Closing to keep the queue honest.

PR #649, opened 2026-08-04
onboarding: route a recovered checkout to Pay instead of a container that does not exist

kavin-114:
## Summary

Customer-plane half of Aerele-RnD/jarvis-admin-v2#162 and #166. A customer whose card was declined and who then rebuilt their bench can now be reconnected by admin, and this bench has to put them down in the right place: a recovered checkout has no container, so the old "connected" answer sent the wizard down `sync_connection` and it waited on a workspace that never appears.

`check_account_reconnect` reads admin's new `subscription_status` and answers `resume_payment` for a `Pending Payment` account. The wizard hands that to the same mount reconciliation a reload runs and lets server truth place the step, which is Pay: the credentials the code just restored are exactly what `resume_pending_signup` was missing.

`reconnect_available` also forwards `company_account_exists` (#166), so the wizard can say "somebody else at this company already has an account". That is not a reconnect offer and must not be rendered as one: this caller cannot recover a colleague's account, and admin will not say whose it is.

**Merge order: this PR merges BEFORE admin-v2 #225.** Both new keys default to the behaviour that shipped before them, so this bench is safe against an older admin. The reverse order is not safe: it lets an older bench redeem a `Pending Payment` account and strand it.

<details>
<summary>Root cause</summary>

`check_account_reconnect` had exactly one success answer, `connected`, because admin's `_ELIGIBLE_STATUSES` was `("Active", "Suspended")` and every reconnectable account therefore had a live tenant. admin-v2 #225 widens that to `Pending Payment` for the failed-payment population, whose whole problem is that `resume_pending_signup` authenticates with the api_key/api_secret the first signup response wrote into a bench that no longer exists. Those accounts have no container by definition, so the single answer stopped being sufficient the moment admin could hand one back.

The SPA branch is placed after the `connected` branch rather than instead of it, and falls through to Details if reconciliation could not establish any server state. Credentials are persisted before either branch runs, so no path here can leave the site unauthenticated, and none of them is a spinner.

The `company_account_exists` forward is deliberately a plain boolean with no identity attached. Admin gates the disclosure on a shared non-consumer mail domain and never names the other account; this wrapper adds nothing and defaults it to false on an older admin or any control-plane failure.
</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on a failure)
- [x] Branch is up to date with `develop` (based on `e887aeb9`, current head)
- [x] New/changed behavior has tests (`TestAccountReconnect`, 11 passing, revert-proofed)

Blocks Aerele-RnD/jarvis-admin-v2#225, which must merge after this.

### DIFF
--- a/frontend/src/views/OnboardingView.vue
+++ b/frontend/src/views/OnboardingView.vue
@@ -2109,6 +2109,21 @@ async function submitReconnectCode() {
 			state.step = "connect";
 			return;
 		}
+		if (d && d.status === "resume_payment") {
+			// Recovered an UNFINISHED checkout, not a live workspace (admin-v2 #162):
+			// there is no container, so the Connect shortcut above would strand this
+			// customer on a workspace that never appears. What the code bought them is
+			// the ability to AUTHENTICATE, which is the one thing the resume path was
+			// missing - so hand off to the same mount reconciliation a reload runs and
+			// let server truth place the step.
+			state.payBusy = false;
+			await reconcileMidFlightSignup();
+			// Reconciliation lands a mid-signup customer on Pay from real state. If it
+			// established none (a control-plane blip), Details is still forward progress
+			// and never a spinner - the credentials are persisted either way.
+			if (state.step === "reconnect") state.step = "details";
+			return;
+		}
 		state.payErr =
 			d && d.status === "expired"
 				? "The reconnect request expired. Start it again."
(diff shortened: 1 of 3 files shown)


kavin-114:
Merge-order note, because the order flipped: **admin-v2 #225 was merged first** (`8ab7d17`, 2026-08-04 18:45 UTC), so `develop` on the control plane now hands a `Pending Payment` account back through reconnect.

That does not break anything on its own, but it does change what this PR is. It is no longer "ship the bench half first so the control plane can follow"; it is now the thing that has to land before admin-v2's `develop` is deployed. Until it does, a customer who recovers a declined-card account on a bench without this change gets `{"status": "connected"}`, the wizard takes the Connect shortcut, and `sync_connection` polls for a container that was never provisioned. The customer sits on "setting up your workspace" with nothing failing, which is the exact symptom admin-v2 #162 set out to remove.

All checks green here (`lint`, `frontend-tests`, `tests` 1 through 4, `coverage`).

PR #653, opened 2026-08-05
feat(onboarding): reset depth ladder L1–L4 + bench disconnect

Navin-S-R:
## ⚠️ Draft — NOT ready to merge

Code review **round 6 is RED** (1 BLOCKER / 7 MAJOR / 16 MINOR) and **no flow review has ever run**. This is opened as a draft to bank the work and make it reviewable, not to merge. The project's review gate was bypassed deliberately, at the repo owner's direction, to get it pushed.

Pairs with `Aerele-RnD/jarvis-admin-v2` branch `feat/reset-depth-ladder` — **neither half works without the other.**

---

## What this adds

A customer-facing **reset depth ladder** in Settings → Reset Workspace, plus a separate **"Disconnect this bench"** action. Levels 1–3 already existed and are re-presented as one coherent choice; L4 and the disconnect are new, and both end with the bench holding no admin credentials.

| Level | Adds | Ends |
|---|---|---|
| L1 | rebuild the container | connected, chat works |
| L2 | + wipe workspace content | connected |
| L3 | + revoke LLM connections | connected, wizard owns LLM setup |
| L4 | + clear the admin connection | **disconnected** → emailed-code reconnect |

L4's ordering is the whole point: the `CONNECTION` clear is deferred until the poll has seen the new container Ready and persisted the fresh connection. There is no window where a reset is in flight and unpollable.

Plan: `.claude/workflow/plans/2026-08-04-reset-workspace-cli-option.md` (+ Amendments 1–4).

## The parts that took six review rounds to get right

**A claim must be atomic, durable, and distinguishable from a rebuild.** `request_workspace_reset` now claims `last_sync_status` *before* its long admin calls, under `jarvis/_redis_lock`, and **promotes** that claim to a convergeable marker only once admin has accepted the request.

That two-state split exists because of a specific defect: `redis_lock`'s release is a `finally`, which **does** run on an exception. So a single transient failure of the one `reset_workspace` call left a convergeable marker *and* a released lock — and the `*/5` reconcile then converged against a container that was never rebuilt, tearing it down and clearing the credentials on an L4.

**Guards must not read stale snapshots.** Frappe holds one MariaDB transaction per request at `REPEATABLE-READ` with `autocommit=False`, so `settings.reload()` **cannot** see another worker's commit — verified with a two-process probe. All three entry points now commit before the guarded read.

**A refusal is not unreachability.** `_clear_admin_connection` makes one authenticated admin call and **aborts** unless the paired chat devices were provably unpaired. Only genuine unreachability proceeds. A dead *container* still disconnects (the unpair is a file op that survives it); a dead *admin* still disconnects (otherwise a lost control plane traps the bench forever); an admin that answers and refuses does not.

**No marker may outlive what can clear it.** `reconcile_pending_workspace_reset` resolves or expires every claim — promote on evidence, leave a young one alone, give up honestly past the deadline.

Also: reset depth is now cumulative **server-side** (only the SPA's radio group enforced it, so `disconnect_after=1, revoke_llm=0` produced a rung the ladder has no name for), and two new Jarvis Settings fields — `workspace_reset_claimed_at`, `reconnect_needs_company` — deliberately in **no** `ResetSpec`, because they describe the aftermath of an operation rather than the tenancy it acted on.

## Frontend

The four depths as one explicit radio choice with per-level consequences, the disconnect as a separate danger-zone action, `OnboardingGate` showing the emailed-code recovery route on the screen a disconnected bench actually lands on, and an L4 that degraded to L3 surfaced to the customer rather than only logged.

## Tests

`jarvis.tests.test_onboarding_sync` **121 → 165** · `GeneralPane.spec.js` **21 → 27** · `test_dev` 10 · `test_account` 74 · `test_settings_on_update` 39 · `test_part4_security` 19 · frontend 737/738 (the one failure is `tests/support-extraction/support-thread.test.js`, pre-existing and outside this change set) · `ruff check` + `format --check` clean at the repo-pinned v0.14.10.

## Known open — read before reviewing

Full detail in `.claude/workflow/reviews/code-feat-reset-depth-ladder.md` and `.claude/workflow/HANDOVER-reset-ladder-2026-08-05.md`.

- **BLOCKER** — `bench_connection_state` returns `disconnected: true` for a bench that has **never onboarded** (both credential fields are blank pre-signup), so every first-time customer sees "Reconnect Jarvis" instead of "Finish setting up".
- **MAJOR** — the `on_update` guard covers 1 of **27** `last_sync_status` writers; `_enqueued_sync_via_admin` writes past it from the worker.
- **MAJOR** — the promote comparison is algebraically unchanged from the form it replaced, and `_LIVE` admits `Applied`, which admin's own `OPEN_STATES` does not.
- **MAJOR** — `_RECONCILE_GIVE_UP_MINUTES = 60` abandons a `Pending Capacity` reset admin will still fulfil.
- **MAJOR** — the two `frappe.db.rollback()` calls and the give-up timer have **zero coverage**.
- **MAJOR** — the new test classes mutate the site they run against; they cascade blanks through Jarvis Settings and must be run against a scratch site. This destroyed a working local bench during development.
- 16 MINORs.

## Review history

Six adversarial rounds. From round 3 onward **every round found defects in work the previous round had signed off**, and three times the defect was introduced *by a fix for another finding in the same session*. Round 6 verified three findings by execution rather than by reading. Weight your review accordingly — reading the code was repeatedly not enough here.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -62,13 +62,31 @@ export const getLlmApplyOperation = (operationId) =>
 	call("jarvis.account.get_llm_apply_operation", { operation_id: operationId });
 
 // ---- workspace reset (customer self-serve, admin-gated) --------------------
+// Four cumulative depths (L1-L4, jarvis/onboarding.py request_workspace_reset):
+// disconnectAfter (L4) additionally clears the bench's OWN admin connection,
+// deferred until the poll below observes the rebuilt container Ready - so it
+// is refused up front (frappe.ValidationError) when the emailed-code reconnect
+// would not be able to restore the site, rather than left half-done.
 export const requestWorkspaceReset = (reason, opts = {}) =>
 	call("jarvis.onboarding.request_workspace_reset", {
 		reason: reason || "",
 		wipe_data: opts.wipeData ? 1 : 0,
 		revoke_llm: opts.revokeLlm ? 1 : 0,
+		disconnect_after: opts.disconnectAfter ? 1 : 0,
 	});
 export const workspaceResetState = () => call("jarvis.onboarding.workspace_reset_state");
+// Terminal, separate from the reset ladder: no rebuild, no poll, just leaves
+// the account. Idempotent - a second call on an already-disconnected bench
+// returns {disconnected:true, already_disconnected:true} rather than erroring.
+// Refused (frappe.ValidationError) up front, same as L4, when the emailed-code
+// reconnect would not be able to restore the site.
+export const disconnectBench = () => call("jarvis.onboarding.disconnect_bench");
+// Poll durable disconnected state: read from local settings with no admin call,
+// so a disconnected bench (which has no credentials) can still answer it. Returns
+// {disconnected: bool, needs_company: bool}. Used on reload to restore the
+// terminal disconnected state with a definitive needs_company answer, so the
+// recovery text never falls back to hedging.
+export const benchConnectionState = () => call("jarvis.onboarding.bench_connection_state");
 
 // --- Per-user chat settings + real (measured) usage tracking, incl. the
 // tenant-admin usage table (design doc §4/§6). All return the house
--- a/frontend/src/components/settings/GeneralPane.spec.js
+++ b/frontend/src/components/settings/GeneralPane.spec.js
@@ -1,0 +1,423 @@
+// Reset Workspace / Disconnect this bench (Settings > General): the depth
+// ladder composes the right flags, the two lockout paths (L4 and the separate
+// Disconnect action) show their cost and their way back before they can be
+// submitted, and the terminal `disconnected` state is never mistaken for a
+// plain unreachable-admin blip or a normal "ready, reloading" reset.
+//
+// Mounted via @vue/test-utils and driven through the exposed <script setup>
+// bindings (w.vm.*) rather than DOM clicks - the same convention
+// LlmPoolEditor.spec.js in this directory uses for its disconnect coverage.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { mount, flushPromises } from "@vue/test-utils";
+
+// GeneralPane.vue -> stores/shell.js reads window.matchMedia at MODULE
+// EVALUATION time (its narrow-sidebar breakpoint watcher, not inside a
+// function), so this has to run before the `import GeneralPane` below even
+// begins to resolve, not just before the first test. vi.hoisted is what makes
+// that ordering guarantee - a plain top-of-file statement would still run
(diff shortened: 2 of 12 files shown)


kavin-114:
Heads up: `develop` has moved under this branch. #697 and #698 both landed today, so a rebase plus a re-run is worth doing before this goes ready.

I trial-merged `feat/reset-depth-ladder` onto current `develop` locally and there is **no text conflict**, so the rebase itself should be uneventful. Two things to look at anyway.

**1. Eight files you touch were also changed on develop since your branch point.** Git merged them cleanly, which is exactly the case where a green PR plus a green base still produces a broken merge. Worth re-running rather than trusting the existing green:

```
frontend/src/api.js
frontend/src/components/settings/GeneralPane.vue
jarvis/admin_client.py
jarvis/dev.py
jarvis/jarvis/doctype/jarvis_settings/jarvis_settings.json
jarvis/jarvis/doctype/jarvis_settings/jarvis_settings.py
jarvis/onboarding.py
jarvis/tests/test_onboarding_sync.py
```

**2. One specific interaction with #697, which changed the onboarding gate's timing.** `AppShell` now re-derives the gate on every route change while gated (`regateOnRouteChange`), and holds a neutral surface via `shellReady` while that check is in flight. Your `OnboardingGate.vue` sits behind that same `v-if`, so it can now **unmount and remount** during a re-check where before it mounted once and stayed. If anything in its `onMounted` is meant to run a single time, or if the `benchConnectionState` read is expensive, that path now fires more often than it used to. It looked fine to me on reading, but you know what that component is meant to do better than I do.

**3. Unrelated, but currently blocking you:** `lint` is failing on the latest run while the other six jobs pass. That is usually the prettier trap on this repo. The gate runs prettier **2.7.1 via pre-commit**, which is not the version `npx prettier` picks up, so local formatting can look clean and still fail CI. `pre-commit run --all-files` reproduces it.

No rush from my side, nothing of mine is waiting on this.

Navin-S-R:
Closing as superseded / over-engineered for the actual requirement.

The requirement this program grew from was simply a 'Reset onboarding' button in Jarvis Settings — now shipped small in jarvis#847 (bench-only, ~216 lines, reuses the existing tested `dev.reset_onboarding`). The L1–L4 depth ladder + the hand-rolled bench-side claim/promote/release coordination layer here are more machinery than that need warrants.

Branch `feat/reset-depth-ladder` is left in place (not deleted) so the salvageable security core — the identity-gated single-call disconnect and the two-legged profile/device teardown — can be lifted later if wanted. Paired with jarvis-admin-v2#231, closed alongside.

PR #654, opened 2026-08-05
fix(settings): clear the row on disconnect, drop the blue focus ring

kavin-114:
## Summary

Disconnecting your last AI model now clears the list instead of leaving the deleted model on screen, and focused inputs no longer draw a blue box inside their own border. Both were reported from a live workspace, where a successful Disconnect left the old provider and model showing with a "Pending re-check" pill, under a banner saying the keys were deleted.

## What changed

- A successful Disconnect empties the list, so the "No models yet" empty state shows. A failed one still leaves the row there to retry.
- An empty pool no longer renders a blank placeholder row numbered "1" with no provider. Onboarding keeps its row.
- Focused inputs stop painting the browser's blue focus ring. Keyboard users get a ring in the app accent on `:focus-visible` only.
- In the provider and model combobox, that ring sits outside the field border rather than inside it.

## Risk and verification

- Both `LlmPoolEditor` suites pass, 31 of 31, including two new tests covering the row list after a successful disconnect and after the reload that follows it fails.
- Full frontend suite against this branch's base: 719 passed before, 721 after, same 31 pre-existing failures both times. Those are the known `localStorage` failures from Node 26 locally versus the Node 24 CI pins, not this change.
- The focus rule is global, so it touches every input. That is intended: the blue ring appeared app wide.

<details><summary>Root cause</summary>

**The surviving row.** `disconnect()` never mutated the row list. It called `api.disconnectLlm()`, set the success banner, then called `load()`, delegating the actual removal to a fresh `getLlmConfig()` refetch. `load()` wraps that refetch in a try block whose catch only sets an error string, so when the fetch throws, `rows` is never reassigned and the deleted model survives on screen. The pill read "Pending re-check" because `accountHealth()` degrades every row when the editor reads dirty, and the snapshot was never reconciled to the now empty pool.

The server half was verified correct and final before this line runs: `disconnect_llm` calls admin first, aborts the whole operation if admin fails, and only then clears the bench. On the live site the settings row already showed `models: []` and `last_sync_status: "disconnected"`, confirming the defect was entirely client side.

The fix asserts the empty state locally on the success path, so a failed refetch can no longer resurrect the row. The failure path is untouched, which the existing "reports a failed disconnect instead of claiming the keys are gone" test still pins.

**The placeholder row.** `load()` had two separate `if (!rows.value.length) rows.value = [newRow()]` guards. For an empty pool with no preset, the first one runs, so removing only the later one had no effect. Onboarding gets its row from the `singleMode` branch, which is the only place that needs one.

**The blue ring.** `index.css` used `outline: revert` in a rule whose own comment says it exists to suppress the forms plugin's blue focus ring. `revert` rolls the property back to the user-agent stylesheet, and the UA rule for a focused input is the browser's blue focus ring, so the declaration reinstated exactly what the rule was written to remove. The neighbouring `box-shadow: none` and `border-color: revert` were correct.

</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on ❌)
- [x] Branch is **up to date with `develop`** (branched from `dabeba01`, the current head)

https://claude.ai/code/session_01KDsRq5555BfFv15X6gBQbc

### DIFF
--- a/frontend/src/components/JvCombo.vue
+++ b/frontend/src/components/JvCombo.vue
@@ -213,6 +213,23 @@ onBeforeUnmount(() => document.removeEventListener("mousedown", onDocClick));
 	font: inherit;
 	padding: 0;
 }
+/* The inner input is not the control the customer sees - .jvc-field is, and it
+   carries the border. Any ring on the input would therefore be drawn INSIDE that
+   border, which is the "box inside a box" look. Focus is signalled on the wrapper
+   instead: a tinted border while anything inside has focus, and for keyboard users
+   a real ring OUTSIDE the border (positive outline-offset), matching the
+   outline: 2px solid var(--cta) / offset 2px idiom used elsewhere in the app. */
+.jvc-input:focus,
+.jvc-input:focus-visible {
+	outline: none;
+}
+.jvc-field:focus-within {
+	border-color: var(--cta-bd);
+}
+.jvc-field:focus-visible {
+	outline: 2px solid var(--cta);
+	outline-offset: 2px;
+}
 .jvc-val {
 	flex: 1;
 	min-width: 0;
--- a/frontend/src/components/JvCombo.vue
+++ b/frontend/src/components/JvCombo.vue
@@ -9,7 +9,7 @@
 			class="jvc-field"
 			:class="{ 'jvc-open': open, 'jvc-dis': !editable }"
 			role="button"
-			tabindex="0"
+			:tabindex="allowCustom ? -1 : 0"
 			:aria-expanded="open"
 			@click="onFieldClick"
 			@keydown.down.prevent="openAnd(0)"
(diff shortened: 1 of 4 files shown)


kavin-114:
### Code review

Found 1 issue, now fixed in [`9ed8067`](https://github.com/Aerele-RnD/jarvis/commit/9ed8067e1003ebd9e33aa3befc87b4c3948517f5).

1. The new keyboard focus ring never fired on the editable combos it was written for. `.jvc-field:focus-visible` only matches when the wrapper div is the focused element, but an `allowCustom` JvCombo contains a real `<input>`, and that is the element the user tabs to and types into. The same hunk suppressed the inner input's own ring, so the model id and base URL fields ended up with no keyboard focus indicator at all. The wrapper also stayed in the tab order, giving one control two consecutive tab stops.

https://github.com/Aerele-RnD/jarvis/blob/bcfc3967b69319d2d4080ca6837d84d75521ca21/frontend/src/components/JvCombo.vue#L225-L232

Fix adds `:has(.jvc-input:focus-visible)` so the ring stays on the wrapper and outside its border, and drops the wrapper from the tab order when it holds an input.

Also checked and found clean: no historical regression (`git log -S` confirms the `outline: revert` line was original and never a deliberate hardening, and the placeholder row has contradicted the "No models yet" empty state since the component was extracted); every `rows[0]` reader is null guarded or singleMode gated; the empty list is not a dead end since "Add a model" is never gated on `rows.length`.

PR #658, opened 2026-08-05
fix(agents): derive rehome membership from origin, not the bumped run pointer

### LINKED ISSUE #615: agents: _rehome_installation_outputs uses last_seen_run as a membership signal, so promote/demote can reassign a foreign owner's finding
`_rehome_installation_outputs` (`jarvis/chat/agents_api.py:1105-1141`) builds its membership set using `last_seen_run`, so a promote or demote can silently reassign the visibility `owner` of a finding that belongs to a different customer's installation.

This is the same unsafe signal that PR #612 (issue #455) removed from the uninstall cascade, on a different code path.

Why `last_seen_run` is not a membership signal: it is the only run pointer the engine re-points, and the recurrence bump that re-points it matches on `(owner, agent, fingerprint)`. Under PP-4 shadow re-homing that bump can attach another owner's finding to our run. So a row reachable only through `last_seen_run` was, by construction, created by somebody else's installation.

`run` and `first_seen_run` are both stamped once at insert by `agent_runs.record_delegate_run` and are listed in the controller's `_FROZEN_FIELDS`, which is what makes them durable statements of origin. Membership should be derived from those two alone, matching what #612 now does in the cascade.

Different fix from #455 though, and it needs its own reasoning and tests: this one changes a finding's visibility rather than deleting it, so the failure is a foreign customer's finding appearing under the wrong owner rather than disappearing.

Found while fixing #455.

kavin-114:
Closes #615

Promoting or demoting an installation could silently rewrite the visibility `owner` of a **different customer's** finding, making it appear under the wrong owner.

`_rehome_installation_outputs` built its membership set from `run`, `first_seen_run` **and** `last_seen_run`. That third pointer is not a membership signal. It is the only run pointer the engine ever re-points, and the recurrence bump that re-points it matches on `(owner, agent, fingerprint)`. Under PP-4, a reviewer holds findings from several installations at once, so that bump can attach another owner's finding to one of our runs. A row reachable *only* through `last_seen_run` was, by construction, created by somebody else's installation.

`run` and `first_seen_run` are both stamped once at insert by `agent_runs.record_delegate_run` and frozen by the controller thereafter, which is what makes them a durable statement of origin.

Membership now comes from `_installation_finding_names`, the same helper the uninstall cascade has used since #455/#612, rather than a local variant of the rule. One definition, so the two paths cannot disagree again, which is how this survived #612 in the first place.

Same unsafe signal as #455, different consequence, which is why it needed its own fix and tests: **the cascade deleted such a row, this path silently reassigns it.** A deletion is at least visible; a reassignment shows one customer's finding to another.

## Tests

`TestRehomeMembershipScope` sets up two installations of the same agent sharing one reviewer, so PP-4 has already pulled both owners' findings under that reviewer, which is the shape that makes this reachable:

- a foreign finding bumped onto our run is **not** re-homed (the regression)
- our own findings still are, and so do our runs, so the fix is not just a break
- a finding we created whose bump has since moved onto the other installation's run is still ours, since origin is what counts and not where the pointer currently sits
- the co-reviewed installation's finding and run are untouched throughout

### DIFF
--- a/jarvis/chat/agents_api.py
+++ b/jarvis/chat/agents_api.py
@@ -1176,23 +1176,31 @@ def _rehome_installation_outputs(inst, to_owner: str) -> None:
 	re-homing is how promotion OPENS the owner surface (and demotion re-closes it).
 	Raw ``db.set_value`` bypasses the finding/run immutability controllers. Rows are
 	located precisely by installation membership (never a broad owner+agent match
-	that could sweep a different install of the same agent)."""
+	that could sweep a different install of the same agent).
+
+	#615: findings come from :func:`_installation_finding_names`, the SAME membership
+	rule the uninstall cascade uses since #455/#612, rather than a local variant that
+	also accepted ``last_seen_run``. That extra pointer is not a membership signal: it
+	is the only run pointer the engine re-points, and the recurrence bump that
+	re-points it matches on ``(owner, agent, fingerprint)``, so under PP-4 shadow
+	re-homing it can attach ANOTHER owner's finding to one of our runs. A row reachable
+	only through it was, by construction, created by somebody else's installation.
+
+	The consequence here differs from #455 and is why this needed its own fix: the
+	cascade DELETED such a row, while this path rewrites its visibility ``owner``. So
+	the failure was a foreign customer's finding silently appearing under the wrong
+	owner rather than disappearing."""
 	run_names = frappe.get_all(
 		RUN, filters={"installation": inst.name}, pluck="name", ignore_permissions=True
 	)
-	dash_names, finding_names = set(), set()
+	dash_names = set()
+	finding_names = _installation_finding_names(run_names)
 	if run_names:
 		for r in frappe.get_all(
 			RUN, filters={"name": ["in", run_names]}, fields=["dashboard"], ignore_permissions=True
 		):
 			if r.dashboard:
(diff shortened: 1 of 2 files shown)


kavin-114:
Note on a deliberate trade in this change, so it reads as considered rather than overlooked.

`_installation_finding_names` leaves alone any finding with no surviving `run` / `first_seen_run` pointer (a pre-run-tracking row, or one whose runs were already deleted). For the uninstall cascade that means "do not delete it", which is unambiguously the safe direction.

Here it means "do not re-home it", and the safe direction is less obvious. A finding that genuinely belongs to this installation but has lost both origin pointers will now stay under the reviewer after a promotion instead of moving to the installer, so the owner will not see it on their surface.

I kept the fail-closed behaviour anyway. Such a row is indistinguishable from another owner's finding bumped onto one of our runs, which is the whole defect this fixes, and the helper's own reasoning applies unchanged: there is no evidence tying it to this installation rather than to any other install of the same agent, and guessing is what caused #455. A legitimate finding staying visible to the reviewer is recoverable; a foreign customer's finding appearing under the wrong owner is not.

Worth knowing when reading the diff: the narrow case that changes behaviour is a finding whose `run` and `first_seen_run` are both gone while its `last_seen_run` still points at a surviving run of this installation.

PR #662, opened 2026-08-05
fix(agents): never re-dispatch a slot whose launch already succeeded

### LINKED ISSUE #652: agents: a failure in _advance after a successful _launch_audit is misreported as a launch failure and re-dispatches the slot
## Summary

In `run_due_agent_audits`, one `try` covers both `_launch_audit` and the `_advance` that follows it. `_launch_audit` commits before it returns, so if `_advance` raises the dispatch has already happened and cannot be undone. The handler treats it as a launch failure: it records a `failed` run, tells the owner nothing ran, and deliberately does not advance `next_run_at`. The row stays due, and the next hourly tick launches the SAME slot a second time.

The customer is told their audit did not run, is billed for two runs against the A14 monthly budget, and gets two audit conversations for one schedule slot.

## Mechanism

`jarvis/chat/agent_scheduler.py`, inside `_sweep_one`:

```python
try:
    frappe.set_user(run_as)
    inst = frappe.get_doc(INSTALLATION, row.name)
    _launch_audit(inst, trigger="scheduled", source_apps=source_apps, initiating_human=None)
    frappe.set_user(original_user)
    _advance(row, now)  # O4: advance ONLY after a successful enqueue
except Exception:
    frappe.set_user(original_user)
    frappe.log_error(title=f"jarvis scheduled audit failed: {row.name}", ...)
    _record_failed(row, "scheduled audit enqueue failed; see Error Log")
    _notify_owner(row.owner, row)
    # Do NOT advance -> retry next hour.
```

`_launch_audit` calls `frappe.db.commit()` before returning, so by the time `_advance` runs the conversation and the `running` Jarvis Agent Run are durable. Any failure in `_advance` (a transient `set_value` or commit error, a lock timeout, a deadlock) therefore lands in a handler whose every action is wrong for that state:

* `_record_failed` writes a `failed` run for an audit that is actually running
* `_notify_owner` tells the customer it could not be started, which is false
* not advancing leaves the row due, so the next tick re-dispatches

The `# O4: advance ONLY after a successful enqueue` comment shows the intent was to treat post-enqueue failures differently. The single `except` does not distinguish them.

## Not introduced by #648

`fix/648-agent-sweep-fault-isolation` moved this block into `_sweep_one` and re-indented it, so it shows as changed in that diff, but the logic is untouched. Filing separately rather than widening that PR.

## Suggested fix

Split the `try` at the enqueue boundary. Once `_launch_audit` returns, the run exists and the slot is spent, so a failure in `_advance` should be logged and the slot consumed rather than re-dispatched. Something like:

```python
try:
    frappe.set_user(run_as)
    inst = frappe.get_doc(INSTALLATION, row.name)
    _launch_audit(...)
except Exception:
    # genuine enqueue failure: nothing durable, safe to retry
    ...
    return
finally:
    frappe.set_user(original_user)

try:
    _advance(row, now)
except Exception:
    # the audit IS running; do not record a failure or re-dispatch
    frappe.log_error(title=f"jarvis advance failed after a successful launch: {row.name}", ...)
```

The second handler should NOT `_record_failed` or `_notify_owner`. Leaving `next_run_at` stale is the lesser evil, and the reaper plus the next successful advance recover it; a duplicate dispatch is unrecoverable and costs the customer budget.

## Verification

* patch `_advance` to raise once, with `_launch_audit` stubbed to succeed, and assert exactly one dispatch across two consecutive `run_due_agent_audits()` ticks
* assert no `failed` Jarvis Agent Run is written for the slot
* assert the owner is not notified that the run could not be started

## Provenance

Found by an adversarial review pass on #651 (CONFIRMED by an independent verifier).

kavin-114:
Closes #652

A scheduled audit could run twice for one slot, be billed twice, and the customer would be told it never started.

`_launch_audit` commits its `running` Jarvis Agent Run before it returns. Anything that goes wrong between that commit and `_advance` therefore leaves the row still **due** with a live run already going, and the next hourly tick dispatches the same slot again.

The old handler was wrong three times over on that path: it wrote a `failed` run for an audit that was actually running, notified the owner that it could not be started, and deliberately did not advance, which is what set up the duplicate.

## Two changes, and the second is the one that closes it

**1. The `try` is split at the enqueue boundary.** A failure after a successful launch no longer records a failure or notifies, because both statements are false. It logs under a distinct title.

**2. Dispatch is now idempotent on the run itself.** An installation that already has a `running` row is skipped.

That second part is load-bearing and I want to flag it, because the issue (which I filed) proposed only the first. **Splitting the `try` alone does not fix the duplicate.** The row stays due either way, so the next tick re-dispatches regardless. I only caught this while writing the tests, after checking whether any dispatch guard already existed. None did.

Guarding on the run also covers every cause rather than one: an RQ worker killed mid-sweep, or an OOM between the two statements, produce exactly the same stale-due-with-live-run state as a failed `_advance`.

<details>
<summary>Why the skipped slot is advanced rather than left alone</summary>

The slot really was consumed, by the launch that is still running. Advancing repairs the `next_run_at` that never got written, instead of leaving the row re-entering this path every hour for the life of the run.

A legitimately long audit that spans the next tick is skipped too. That is correct: two concurrent audits of one installation is not something anyone wants. And `reap_stale_agent_runs` terminalizes a genuinely stuck run at 3h, so this can never wedge a schedule permanently.
</details>

## Tests

`TestDispatchIdempotency`, six cases:

- an installation with a live run is not dispatched again (the regression)
- the skipped slot is advanced, so a stale `next_run_at` self-repairs
- an `_advance` failure after a launch is not recorded as a launch failure
- a genuine launch failure still records, still names the reason, and still leaves the slot due for a retry
- a clean installation dispatches normally

### DIFF
--- a/jarvis/chat/agent_scheduler.py
+++ b/jarvis/chat/agent_scheduler.py
@@ -234,6 +234,26 @@ def _sweep_one(row, now, original_user: str, seen: set) -> None:
 		_advance(row, now)
 		return
 
+	# #652: idempotency at DISPATCH, which is the only place it actually holds.
+	# ``_launch_audit`` commits its ``running`` Jarvis Agent Run before returning, so
+	# anything that goes wrong between that commit and ``_advance`` below - a failed
+	# set_value, an RQ worker killed mid-sweep, an OOM - leaves the row still DUE with a
+	# live run already going. The next tick then dispatches the SAME slot a second time:
+	# a duplicate audit, billed twice against the A14 budget, for a customer who was
+	# told nothing ran.
+	#
+	# Guarding on the run itself covers every one of those causes, where guarding only
+	# the advance covers one. A legitimately long audit spanning the next tick is
+	# skipped too, which is correct - two concurrent audits of one installation is not a
+	# thing anyone wants - and ``reap_stale_agent_runs`` terminalizes a genuinely stuck
+	# run at 3h, so this can never wedge a schedule permanently.
+	#
+	# The slot IS advanced here: it was really consumed by the launch that is still
+	# running, so this also repairs the ``next_run_at`` that never got written.
+	if frappe.db.exists(RUN, {"installation": row.name, "status": "running"}):
+		_advance(row, now)
+		return
+
 	# S1 hinge: mint the run session + create conv/run INSIDE set_user(run_as).
 	# Row ownership is reassigned to the human owner inside _launch_audit; only
 	# the ERP-read identity is the run-as user.
@@ -243,10 +263,8 @@ def _sweep_one(row, now, original_user: str, seen: set) -> None:
 		# initiating_human=None is EXPLICIT (JF-021): a cron run is unattended, so
 		# it has no initiating human — the scheduler user (Administrator) is not one.
 		_launch_audit(inst, trigger="scheduled", source_apps=source_apps, initiating_human=None)
-		frappe.set_user(original_user)
-		_advance(row, now)  # O4: advance ONLY after a successful enqueue
 	except Exception:
-		frappe.set_user(original_user)
+		# Nothing durable was created, so this slot really can be retried.
 		frappe.log_error(
 			title=f"jarvis scheduled audit failed: {row.name}",
 			message=frappe.get_traceback(),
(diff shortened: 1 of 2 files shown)


kavin-114:
Closing unmerged. The high-effort review on this PR found three independent holes, and they show the premise was wrong rather than the implementation being incomplete.

Splitting the try stops the false failure record and the false notification, which is worth having, but the row is left due either way so the next tick still re-dispatches. And the running-run guard I added does not close it either:

1. it stops protecting the moment the run finishes, which is the common case (audits take minutes, ticks are hourly)
2. it is an unlocked check-then-act, so a manual Run Now racing the sweep can still double-launch
3. an orphaned `running` row from a mid-launch exception makes it silently skip and advance real slots for up to 3h, masking a genuine failure

I would rather not land a fix that reads as closing #652 while leaving the duplicate reachable by three routes. Re-scoped with the constraints a correct fix has to satisfy, and the orphaned-row question that needs settling first.

PR #673, opened 2026-08-05
fix(wiki): a refused ingest update is no longer silent, and is retried

### LINKED ISSUE #613: wiki: a refused ingest update is dropped with no log and no retry, so the note is marked Processed and the knowledge is lost
A voice-note update that names no existing page and carries neither `title` nor `page_type` is dropped silently, and the note is marked Processed, so the knowledge is gone with no log line and no retry.

`_apply_one_update` returns `False` for this case, which is counted as "refused" rather than "failed". `_ingest_note` only retries on `failed`, so a refused item is indistinguishable from a deliberate no-op and never comes back.

This is pre-existing, but PR #611 raises its likelihood: now that the ingest is told to emit `append_md`, items that carry only an append and no page-identifying fields become the normal shape rather than a rare one.

Two things worth deciding together:

- Whether a refusal should be logged at all. Today there is no trace, so a tenant losing knowledge this way has nothing to look at.
- Whether "refused" and "failed" should stay separate. If a refusal is always a caller or model error rather than a legitimate skip, it probably belongs on the retry path.

Found while fixing #488.

kavin-114:
Closes #613

A voice note's knowledge could be lost with no log line, no retry, and no trace anywhere.

A refused update counted as **neither** applied nor failed. `_ingest_note` retries only on `failed`, so an all-refused batch returned `(0, 0)`, the note was marked Processed with "nothing durable found", and the knowledge was gone for good.

That message was itself misleading: the model *did* produce updates. They were refused.

The refusing shape is an update naming no existing page while carrying neither `title` nor `page_type`, so nothing can be minted from it. PR #611 made that shape ordinary rather than rare by telling the ingest to emit `append_md`.

## Answering both questions the issue raises

**Should a refusal be logged?** Yes, and it now is, with the shape that caused it: the slug, whether a title was present, the `page_type`, the scope, and which content keys the update carried.

**Field names and presence only, never the note's text.** A voice note is the customer's own words and must not be copied into an Error Log. There is a test asserting that.

**Should refused and failed stay separate?** No. A refusal is now counted as failed, which is what puts it on the retry path. It also makes the tuple path agree with the `return_outcomes` path, which already treats a refusal as failed. **The two disagreeing is what kept this invisible**, and that pattern has come up repeatedly in this backlog.

The note therefore stays New for the daily `voice_facts` sweep instead of being marked Processed with its knowledge dropped.

<details>
<summary>One residual I deliberately did not fix here</summary>

`Jarvis Voice Note` carries no attempt counter (only `status`, `processed_at`, `processed_note`), so a note that refuses deterministically will now be retried daily indefinitely.

I judged that acceptable and strictly better than the current behaviour: the cost is bounded and visible (one Error Log line and one extraction per day, per stuck note), where today the failure is silent and permanent. Bounding it properly needs a schema change to carry an attempt count, which is a different piece of work and wants its own issue rather than being smuggled in here.

Worth knowing when reading this: it converts a silent permanent loss into a noisy recoverable one, which is the trade I would want, but it is a trade.
</details>

## Tests

`TestRefusedUpdateIsNotSilent`, five cases:

- an update that cannot mint a page counts as **failed**, not `(0, 0)`
- the refusal is logged with the shape that caused it
- **the log does not leak the note's content**
- a valid update still reports `(1, 0)` and logs nothing
- an update carrying an identity still creates the page rather than refusing

### DIFF
--- a/jarvis/chat/wiki.py
+++ b/jarvis/chat/wiki.py
@@ -1111,11 +1111,11 @@ def apply_extracted_page_updates(
 	``_apply_one_update`` returning ``False``) as failed — the aggregate tuple
 	cannot distinguish "created page B" from "refused colliding page A".
 
-	#613: the tuple path now counts a refusal as FAILED too, and logs it. It used to
-	count it as neither and leave no trace, so an all-refused batch returned (0, 0);
-	``_ingest_note`` retries only on ``failed``, so the voice note was marked Processed
-	with "nothing durable found" and its knowledge was lost for good. Both paths now
-	agree on what a refusal means, which is what stopped this being visible.
+	#613: the tuple path now LOGS a refusal (it left no trace at all before), but still
+	counts it as neither applied nor failed. Whether a refusal belongs on the retry path
+	is a separate decision that this change deliberately does not take: ``test_wiki``
+	pins the current contract with an explicit comment, and flipping it would retry
+	unsalvageable input forever with no attempt counter to bound it.
 	"""
 	if not isinstance(updates, list):
 		return [] if return_outcomes else (0, 0)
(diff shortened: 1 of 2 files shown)


kavin-114:
Scope reduced after CI, and the reason is worth recording.

Counting a refusal as `failed` broke four existing tests, and reading them showed the first cut was wrong rather than the tests being stale. `test_wiki.py:334` pins the contract with an explicit comment:

```python
# A skipped (identity-less) update is not a FAILURE - the note may
# still be marked Processed.
self.assertEqual((applied, failed), (0, 0))
```

That is a deliberate documented decision and I overrode it silently. Updating those tests to match my change would have been rewriting a contract to make my own patch pass.

There is also a concrete harm I had not accounted for. One refusal pinning that contract is slug `"!!!"`. No amount of retrying repairs it, and `Jarvis Voice Note` carries no attempt counter, so the daily sweep would have retried unsalvageable junk forever.

So this PR now takes only the unambiguous half: a refusal is logged with the shape that caused it, so knowledge lost this way is diagnosable. The retry question stays open on #613, where it needs a way to distinguish retryable model error from junk, and probably an attempt counter.

The tests now assert the documented tuple is UNCHANGED, so nobody flips it by accident later.

Issue #702, opened 2026-08-07
Chat turn errors show upstream gateway text, not a cause the customer can act on


kavin-114:
A failed chat turn shows the gateway's own wording instead of a cause the customer can act on, and that wording can be flatly wrong.

**Steps:**
1. Onboard a workspace and send any chat message.
2. Have the turn fail for a reason that is not network related. Observed here: openclaw could not parse `/home/node/.openclaw/devices/paired.json` while the file was mid-rewrite.

**Expected:** a headline naming what happened, a "what you can do" hint, and the raw reason behind Show details.

**Actual:** headline "Something went wrong", detail "LLM request failed: network connection error.", no hint. Nothing was wrong with the network. The container log showed the real cause, `JsonFileReadError ... SyntaxError: Expected double-quoted property name at position 892`, and a retry 30 seconds later succeeded.

That string exists in no Jarvis repo, so it is openclaw's text passed through verbatim.

**Suspected location:** `ActionError.vue` already documents the shape it wants, `{ok:false, error:{code, message, detail, hint}}`, but turn failures reach it as a plain string and take the `typeof e === "string"` fallback, dropping code and hint. `lib/errors.js` normalizes Frappe API errors and has no counterpart for gateway or turn errors.

Proposal, not a decision: audit request to response error handling end to end, decide which upstream conditions map to which customer-facing cause, and give turn errors the same normalizing layer Frappe API errors already have. Two failures that need different customer action should never render as the same sentence.

kavin-114:
Fixed by #704: chat turn failures now carry a classified cause and a retry hint instead of relaying the gateway's wording verbatim. The observed string `LLM request failed: network connection error.` now classifies as a retryable gateway fault, with a test asserting it does NOT classify as `unreachable`, while genuine transport failures still do. Closing by hand because `Closes #N` only fires on the default branch (`beta`).

PR #704, opened 2026-08-07
Give chat turn errors a customer-facing cause and a retry hint

kavin-114:
## Summary

A failed chat turn now shows a headline naming what happened plus a retry hint, instead of openclaw's raw wire text ("LLM request failed: network connection error.") shown as an unexplained "Something went wrong", in both the main chat and the Desk mini-chat widget.

## What changed

- `turn_handler.py`: `_classify_error` gets a `gateway` code for a mid-run failure matching no unreachable/timeout/provider marker, plus an `internal` marker so a reload matches the live event.
- `frontend/src/lib/errors.js`: `turnErrorInfo()`, the counterpart to `errMessage()` for a turn's raw error text. Total, mirrors the Python taxonomy, adds a hint per code.
- `ChatView.vue`: renders the hint; classifier moves into `errors.js`, one source instead of two.
- `pump.py` / `prepare.py`: a pre-ack rejection and a local prompt-assembly bug keep an honest code, not `gateway`.
- Desk mini-chat widget (`Panel.vue`): its own separate copy of this taxonomy, synced.

## Risk and verification

- New tests pin the #702 string to `gateway`, the backstop text to `internal`, and the pre-ack/local-bug cases above.
- Frontend suite green under Node 24, pre-commit clean, hosted CI green.
- `_STALLED_ERROR` / `_ORPHAN_ERR` (pre-existing, unmatched text) also classify `gateway` now, deliberately: a lost run is best answered with a fresh message.

<details>
<summary>Error-path map and root cause</summary>

```
ChatView.vue / Panel.vue (composer send)
  -> jarvis/chat/api.py chat_send -> frappe.enqueue(jarvis.chat.worker.run_agent_turn)
  -> jarvis.chat.worker.run_agent_turn (thin shim)
  -> jarvis/chat/turn_handler.py handle_chat_send
       - AgentSession.chat_send (agent_client.py) -> WS RPC to openclaw
       - AgentSession.relay_turn_events consumes broadcast frames until a
         terminal `chat` event (state final|error|aborted)
       - a state=="error"/"aborted" chat event yields
         {"kind": "relay:error", "error": payload.get("errorMessage") or state}
         - openclaw's own errorMessage text, passed through VERBATIM. This is
           where "LLM request failed: network connection error." came from:
           openclaw's own generic wording for a mid-run failure, observed live
           to be a JSON parse race on paired.json, nothing to do with the
           network.
  -> turn_handler._publish_run_error(err_text) -> _classify_error(err_text)
       - before this PR: no keyword matched "network connection error", so it
         fell through to "internal", giving the SPA headline "Something went
         wrong" with no hint.
  -> realtime `run:error` event {error, code} -> ChatView.vue / Panel.vue
     socket handler -> headline + (new) hint; raw text stays behind
     "Show details"
```

The newer Relay-Pump path (`jarvis/chat/pump.py` -> `settlement.py`) and the
prepare-stage path (`jarvis/chat/prepare.py`) are additional callers into the
same turn, all delegating to `turn_handler._classify_error` too
(`settlement._classify`, `pump._classify_error`, `prepare._prepare_error`).
Two of them are pre-ack (the run never started: a definite gateway rejection,
or a local bug in our own prompt assembly) and needed an explicit code rather
than the new mid-run `gateway` default, since neither is "a run that started
and then hit a brief hiccup".

`ActionError.vue` (the `{ok:false, error:{code, message, detail, hint}}`
envelope named in the issue) renders a different failure surface: gated
write/tool confirmations (`jarvis/api.py _translate_write_error`). Its own
code comment says that headline vocabulary is kept separate from turn or
transport errors on purpose. Turn failures never reach that component; they
render via a hand-rolled card in `ChatView.vue` (around line 825, and its own
copy in the Desk widget's `Panel.vue`) that already mirrors its layout
(headline, details expander, retry button) but was missing the hint. This PR
gives that card the same enriched treatment `ActionError.vue` documents,
headline plus hint plus expandable raw detail, without merging the two
components, since the retry button and "no changes made" line have no
`ActionError.vue` equivalent yet.
</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on ❌)
- [ ] Branch is up to date with `develop`

### DIFF
--- a/frontend/src/components/ActionError.vue
+++ b/frontend/src/components/ActionError.vue
@@ -29,8 +29,9 @@ const props = defineProps({
 });
 
 // Plain-language headline keyed by the wire `code`. This is the ACTION-failure
-// axis - distinct from ChatView's run/transport ERROR_HEADLINES, kept separate
-// on purpose.
+// axis - distinct from ChatView's run/transport taxonomy (lib/errors.js
+// turnErrorInfo, formerly ChatView's own ERROR_HEADLINES before #702), kept
+// separate on purpose.
 const HEADLINES = {
 	PermissionDeniedError: "You don't have permission to do this",
 	InvalidArgumentError: "Some values need attention",
--- a/frontend/src/lib/errors.js
+++ b/frontend/src/lib/errors.js
@@ -93,8 +93,8 @@ export function errHtml(e, fallback) {
 // ---------------------------------------------------------------------------
 // Chat TURN failures (#702): the counterpart to errMessage() above, for a
 // chat turn's `error` column instead of a Frappe API exception. That text is
-// raw prose relayed from openclaw over the realtime channel, sometimes
-// verbatim - #702 was openclaw reporting "LLM request failed: network
+// raw prose relayed from the agent gateway over the realtime channel, sometimes
+// verbatim - #702 was the agent reporting "LLM request failed: network
 // connection error." for a turn that failed for an entirely different reason
 // (a device-pairing file caught mid-rewrite), which the SPA had no way to
 // tell apart from a real network failure. Mirrors the operator-facing
@@ -165,6 +165,7 @@ function classifyTurnErrorCode(raw) {
 		[
 			"quota",
 			"rate limit",
+			"rate-limit",
 			"cooldown",
 			"overloaded",
 			"insufficient",
@@ -173,7 +174,7 @@ function classifyTurnErrorCode(raw) {
 		].some((k) => low.includes(k))
 	)
 		return "provider";
-	// #702: openclaw's own mid-run failure text (e.g. "LLM request failed:
+	// #702: the agent's own mid-run failure text (e.g. "LLM request failed:
 	// network connection error.", relayed verbatim) is not a reliable signal
 	// that the network was actually the problem - see the module comment
 	// above. A run that got far enough to be accepted and start is
--- a/frontend/src/lib/errors.js
+++ b/frontend/src/lib/errors.js
@@ -89,3 +89,115 @@ export function escapeHtml(v) {
 export function errHtml(e, fallback) {
 	return escapeHtml(errMessage(e, fallback));
 }
+
+// ---------------------------------------------------------------------------
+// Chat TURN failures (#702): the counterpart to errMessage() above, for a
+// chat turn's `error` column instead of a Frappe API exception. That text is
+// raw prose relayed from openclaw over the realtime channel, sometimes
+// verbatim - #702 was openclaw reporting "LLM request failed: network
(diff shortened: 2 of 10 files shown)


kavin-114:
### Code review

A 5-angle review (CLAUDE.md compliance, shallow bug scan, git history, prior PR comments, code comment compliance) found 5 real issues. All are fixed on this branch before merge; listed for the record.

1. Classifier keyword sync gap: JS was missing the hyphenated `"rate-limit"` provider keyword Python had, and Python's `unreachable` check was missing `"connection timed out"`, which JS already treated as a transport failure. Same text classified differently live vs. on a reload. Fixed in 9d64c354.

https://github.com/Aerele-RnD/jarvis/blob/dcedb31aaf328b7fa630aa269842fe591a36803d/jarvis/chat/turn_handler.py#L1955-L1967

2. `pump.py`'s `_handle_ack_failure` held an `AgentUnreachableError` for a definite pre-ack rejection (e.g. a policy denial) but did not pass it to `_classify_error`, so text-only matching landed on the new `gateway` default ("brief hiccup, retry") instead of `unreachable`, for a request that never even started. Fixed in d52d14a6.

https://github.com/Aerele-RnD/jarvis/blob/dcedb31aaf328b7fa630aa269842fe591a36803d/jarvis/chat/pump.py#L2059-L2063

3. `prepare.py`'s `assemble_prompt` failure is a bug in our own code, not a gateway signal, but had no way to avoid the same guess and would also default to `gateway`. Added an explicit `code` override. Fixed in d52d14a6.

https://github.com/Aerele-RnD/jarvis/blob/dcedb31aaf328b7fa630aa269842fe591a36803d/jarvis/chat/prepare.py#L129-L133

4. The Desk mini-chat widget (`Panel.vue`) carries its own independent copy of this taxonomy (a separate Desk bundle, no shared import path into `frontend/src/lib`). It never got the `gateway` code or any hint, so the exact #702 bug still reproduced there. Synced. Fixed in dcedb31a.

https://github.com/Aerele-RnD/jarvis/blob/dcedb31aaf328b7fa630aa269842fe591a36803d/jarvis/public/js/jarvis_chat/widget/Panel.vue#L719-L727

5. Two comments this PR's own edits made stale: `ActionError.vue` still named the deleted `ChatView` `ERROR_HEADLINES` symbol, and a `ChatView.vue` comment overclaimed a `detail` key `errorInfo()` does not return. Fixed in dcedb31a.

https://github.com/Aerele-RnD/jarvis/blob/dcedb31aaf328b7fa630aa269842fe591a36803d/frontend/src/components/ActionError.vue#L31-L34

PR #706, opened 2026-08-07
fix(onboarding): stop offering a second payment while one is still confirming

### LINKED ISSUE #705: Wizard offers Start a new payment while a live mandate is still confirming
The onboarding wizard offers "Start a new payment" while a mandate the customer already authorized is still resolving, which is how a customer ends up authorizing a second one.

`PAYMENT_CONFIRMATION_PENDING` in `frontend/src/onboarding/paymentCodes.js:177` carries `actions: [ACTIONS.CHECK, ACTIONS.INITIATE]`, so the screen renders a "Start a new payment" button next to "Check payment status".

The admin plane deliberately assigns that same code no recovery action at all. From `jarvis_admin_v2/billing/signup.py:2657`:

> PAYMENT_CONFIRMATION_PENDING deliberately has none: there is nothing for the caller to do but wait, and inventing an action for it is how a waiting customer gets talked into paying twice.

So the two planes disagree, and the wizard is on the wrong side of a rule the backend states in words.

**Steps:** 1. Complete a real checkout. 2. Return to the wizard before the mandate confirms.
**Expected:** check and wait, no invitation to pay again.
**Actual:** "Start a new payment" is offered.

Observed during an onboarding e2e today. A second checkout was opened 84 seconds after the first, and the original subscription is now referenced nowhere on the admin plane.

Note the sibling state in the same file, `PAYMENT_AUTHORIZED_PENDING_CONFIRM`, already gets this right: `actions: [CHECK]` only, with body text saying "please do not pay again, as that would authorize a second mandate". The fix should make the two consistent rather than invent a third treatment.

Proposal, not a decision: drop INITIATE from `PAYMENT_CONFIRMATION_PENDING`. Before doing so, confirm no legitimate path reaches that code with no money outstanding, since removing the only forward action there would strand such a customer. If such a path exists, the real defect is that one code covers two different situations and it needs splitting.

kavin-114:
Fixes #705

The onboarding wizard no longer offers "Start a new payment" while a payment on the
same signup is still unconfirmed, so a customer can no longer authorize a second
mandate by clicking it.

## What changed
- `PAYMENT_CONFIRMATION_PENDING` now carries `actions: [CHECK]`, matching its sibling
  wait state `PAYMENT_AUTHORIZED_PENDING_CONFIRM` and admin's own contract, which
  gives this code no recovery action on purpose.
- Its body copy now says to check status before doing anything else and warns that
  paying again would authorize a second mandate.
- Extended `paymentCodes.test.js` with the same "check only, never a second payment"
  assertion the sibling state already had, and fixed the one existing test that
  asserted the old (buggy) behavior.

## Risk and verification
- A customer with a genuinely abandoned checkout is not stranded: RESUME (added
  dynamically, unaffected by this change) reopens the same live token, and CHECK
  itself resolves an untouched checkout to `NO_CURRENT_INTENT`, whose row still
  offers Initiate first. Traced in `jarvis_admin_v2/billing/signup.py`.
- `node --test` on the onboarding suite (597 tests) and `vitest run` (848 tests) pass
  under Node 24. `pre-commit` passes on the changed files.

<details><summary>Investigation: does a no-money path reach this code, and does dropping INITIATE strand it?</summary>

Yes, a live pay-page token with no money moved does reach plain
`PAYMENT_CONFIRMATION_PENDING` (`signup.py` `_signup_payment_state`, ~2863-2880: any
Pending Payment subscription with a live token, regardless of whether the customer
ever opened the gateway sheet). It is not stranded by dropping INITIATE, for two
independent reasons:

1. While the token lives, RESUME is injected dynamically by `recoveryActions` in
   `OnboardingView.vue` and reopens the same checkout, no new intent.
2. CHECK converges it: `_RECONCILE_CODES` maps a dead or never-opened intent to
   `NO_CURRENT_INTENT` (`signup.py` 2966-2967), whose row leads with Initiate. A
   Pending Payment subscription whose token has simply expired gets `NO_CURRENT_INTENT`
   directly (`signup.py` 2913-2919), same result.

So the backend already performs the split this code seemed to need, dynamically,
through CHECK and RESUME. No second wizard code was needed.

The incident mechanism: admin's `CommittedMoneyError` guard (`signup.py` 2049) that
would refuse a second intent only fires once the ledger has recorded money as
committed, which needs a webhook or an explicit reconcile. In the reported 84 second
window neither had run, so `can_initiate_payment` was still true and the button was
live and clickable. That is why the frontend must not render Initiate on this
ambiguous code at all, rather than relying on the backend flag alone.

Repo-wide grep confirmed `paymentCodes.js` is the only wizard copy table; no
duplicate exists under `pwa/`.

</details>

## Pre-merge checklist

- [x] **CI is green** - the `tests` check on this PR passes (never merge on X)
- [x] Branch is **up to date with `develop`** (so it is tested against the latest code)

### DIFF
--- a/frontend/src/onboarding/paymentCodes.js
+++ b/frontend/src/onboarding/paymentCodes.js
@@ -175,10 +175,20 @@ const TABLE = {
 		actions: [ACTIONS.VERIFY],
 	},
 	[CODES.PAYMENT_CONFIRMATION_PENDING]: {
+		// A WAIT state, same treatment as PAYMENT_AUTHORIZED_PENDING_CONFIRM below
+		// (jarvis#705). This row cannot tell apart a checkout the customer never
+		// touched from a mandate they already authorized whose webhook has not
+		// landed yet, so it must never offer INITIATE for either: a live e2e run
+		// authorized a second mandate by clicking it 84 seconds after the first,
+		// before admin's own committed-money guard had anything to refuse. Nobody
+		// is stranded by CHECK alone - while the token lives, RESUME (added
+		// dynamically, see OnboardingView.vue) reopens the SAME checkout, and CHECK
+		// itself resolves a genuinely untouched one to NO_CURRENT_INTENT, which
+		// offers INITIATE.
 		headline: "We have not confirmed this payment.",
-		body: "If money was deducted, check the status before starting another payment.",
+		body: "Check the status before doing anything else. If money already moved, please do not pay again, as that would authorize a second one.",
 		tone: TONE.STATUS,
-		actions: [ACTIONS.CHECK, ACTIONS.INITIATE],
+		actions: [ACTIONS.CHECK],
 	},
 	[CODES.PAYMENT_DECLINED]: {
 		headline: "The payment was declined. No payment was confirmed.",
--- a/frontend/src/onboarding/paymentCodes.js
+++ b/frontend/src/onboarding/paymentCodes.js
@@ -381,8 +381,10 @@ const TABLE = {
  *
  * Deliberately NOT a code of its own: admin answers the ordinary
  * PAYMENT_CONFIRMATION_PENDING when money is parked, because a decline there
- * would invite a second payment. The flag is the only thing that distinguishes
- * the two, and it is the flag - never the code - that suppresses Pay.
+ * would invite a second payment. The flag only changes the SENTENCE now
+ * (jarvis#705 made the plain row check-only too, same as this one); it still
+ * vetoes the backend can_initiate_payment flag one layer down, in
+ * paymentMachine.js.
  */
 const PENDING_AWAITING_RECONCILIATION = Object.freeze({
 	headline: "We are still confirming your payment.",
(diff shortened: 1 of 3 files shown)


kavin-114:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

One review angle flagged a stale docstring that a prior commit in this PR made inaccurate ([`PENDING_AWAITING_RECONCILIATION`'s doc comment](https://github.com/Aerele-RnD/jarvis/blob/2f366b29eade1ff4d4cc70cf5a4e3e58225a59d2/frontend/src/onboarding/paymentCodes.js#L379-L386) claimed the flag alone suppressed Pay, which stopped being true once the plain row went check-only too). Fixed in [7f7402c6](https://github.com/Aerele-RnD/jarvis/commit/7f7402c6).

Four independent review angles (CLAUDE.md compliance, a shallow bug scan of the diff, git-history/precedent check, and prior-PR-comment precedent) returned clean.

Issue #681, opened 2026-08-05
A completed turn stays stuck on "Finishing..." after a follow-up lane 401s


kavin-114:
A successful chat turn rendered its answer, "2 tools" and "15.9s", then kept showing "Finishing..." indefinitely. Several minutes later the label had still not cleared.

Container logs show the tool call succeeded (`tool=get_list result=ok`) and a later lane ended with `isError=true rawError=401 invalid or missing credentials`, roughly a minute after credentials had been updated and the main lane had already picked them up. So a background lane was still holding stale auth.

**Steps:** 1. Update the API key on a live tenant. 2. Send a message that succeeds.
**Expected:** The status clears, or the follow-up failure surfaces its own error.
**Actual:** A permanent "Finishing...", which implies the answer is incomplete when it is not.

Two things worth separating here: the stuck label, and why a lane kept stale credentials after an apply that the main lane honoured.

Found during the full onboarding e2e on 2026-08-06 (finding F26).

kavin-114:
Fixed and merged. Closing.

**Fix:** PR #736, merged into `develop`. **Requires an SPA rebuild** (`cd frontend && npm run build`). No `bench migrate`.

## The stuck label

The label was driven by a Set that exactly one thing could ever clear: the `message:enriched` realtime push (`ChatView.vue:7860-7863` on develop). Not `loadConversation`, not a reconnect resync, not any timeout. One missed best-effort event meant a permanent label.

Three independent paths reached that state:

1. **A failing effect holds the turn open.** `run_finalize` publishes `message:enriched` only once EVERY required effect is done (`finalize.py:125`). The usage effect retries by design when the gateway session row is missing or stale (`finalize.py:315`), which is exactly what a live credential change produces. The only retry driver is the `*/5` watchdog (`hooks.py:275` into `pump.py:3240`) with a budget of 3 (`turn_state.py:68`), so the force-done floor is about 15 minutes, and never where that cron is stopped.
2. **The single publish is lost** to a socket blip, a hidden tab, or a client on another route. There is no replay on the socket.
3. **Nothing could heal case 2.** `run_finalize` returned early on an already-`done` turn (`finalize.py:70`) without republishing. The `run:end` terminal has a CDX-12 re-publish backstop (`finalize.py:364`); the enrichment terminal had no counterpart.

[...]

PR #736, opened 2026-08-08
fix(chat): bound the "Finishing..." affordance so it can never stick (#681)

kavin-114:
## Summary

A finished chat turn could keep showing "Finishing..." forever; now that label always clears, so a complete answer never looks unfinished. Anyone whose post-turn enrichment stalls notices, and the reporter's live case (a background lane dying on credentials that had just been rotated) is one of several ways to trigger it.

The label was driven by a single best-effort `message:enriched` realtime push and had nothing else that could ever clear it. Two changes make it self-limiting:

- **Client (the real fix).** The bookkeeping moves out of `ChatView` into `frontend/src/lib/enrichmentPending.js`, where every entry carries a 120s deadline. On expiry the entry is dropped and the conversation is refetched once, so any enrichment that did land still renders. This holds whatever the server is doing.
- **Server.** A finalize that arrives after the turn is already `done` now re-delivers `message:enriched` instead of returning `already_done` in silence. That is the counterpart of the CDX-12 terminal re-publish backstop, which the enrichment terminal never had.

The turn state machine is deliberately unchanged: a failing effect still holds the turn in `finalizing` and still force-dones at the retry budget.

**No `bench migrate`.** SPA rebuild required (`cd frontend && npm run build`).

<details>
<summary>Root cause</summary>

**The label.** Settlement publishes `run:end` with `enrichment_pending`, and `ChatView` kept the message id in a Set until `message:enriched` arrived (`ChatView.vue:4248`, `:7810`, `:7860` on develop). Nothing else cleared that Set: not `loadConversation`, not a reconnect resync, not any timeout. So every one of these left it up permanently on a reply that was already complete:

1. A finalize effect that keeps failing. `finalize.run_finalize` only flips `finalizing -> done` and publishes `message:enriched` once **every** required effect is done (`finalize.py:125` on develop). The usage effect retries by design when the gateway session row is missing or stale (`finalize.py:315`), which is exactly what a live credential change produces. The only retry driver is the `*/5` pump watchdog cron (`hooks.py:275` -> `pump.py:3240`), and the budget is 3 (`turn_state.py:68`), so the floor is roughly 15 minutes, and forever wherever that cron is not running.
2. The push itself lost to a socket blip, a hidden tab, or a client on another route. There is no replay on the socket.
3. `run_finalize` returning early on an already-`done` turn (`finalize.py:70` on develop) without republishing, so nothing could ever heal case 2.

**The stale-credential lane** is a separate defect and does not live in this repo. Written up in full in a comment on this PR, with the fleet-agent and openclaw file references, so it can be filed against the right repo. It is why the client-side bound is the primary fix here rather than a server-side retry tweak: the label must not depend on that being fixed.

</details>

<details>
<summary>Manual test plan</summary>

Rebuild the SPA first (`cd frontend && npm run build`), then hard-reload the chat page.

1. **Happy path is unchanged.** Send a message on a healthy tenant. The reply renders, "Finishing..." appears briefly under it, then disappears within a second or two once enrichment lands, and any late attachment / canvas / title pops in.
2. **The reported bug.** Stall enrichment so `message:enriched` never fires. Easiest reliable way: stop the openclaw container (`docker stop <tenant-container>`) right after a turn's answer finishes rendering, so the finalize usage poll cannot reach the gateway. Expected: "Finishing..." shows, and clears on its own within about two minutes. Before this change it stayed up indefinitely. Restart the container afterwards.
3. **Navigating away does not strand it.** Trigger case 2, switch to another conversation and back before the deadline. The label should still clear on schedule rather than surviving the reload.
4. **Server re-delivery.** For a turn that is already `done`, run `bench --site <site> execute jarvis.chat.finalize.run_finalize --args '["<run_id>"]'`. It should return `{"ok": true, "already_done": true, "republished": true}` and a chat tab open on that conversation should refresh the reply. This also works as a manual un-stick for any client caught by an older build.

</details>

## Pre-merge checklist

- [x] **CI is green** - the `tests` check on this PR passes (never merge on X)

### DIFF
--- a/frontend/src/lib/enrichmentPending.js
+++ b/frontend/src/lib/enrichmentPending.js
@@ -1,0 +1,120 @@
+// SUX-7 / jarvis#681: the "Finishing…" affordance's bookkeeping, extracted as a
+// PLAIN, importable, unit-tested module (ChatView.vue imports it).
+//
+// What the affordance is. Settlement publishes the turn's terminal `run:end` with
+// `enrichment_pending`, which tells the SPA to keep a subtle "Finishing…" line under
+// an otherwise finished reply while the finalize job is still adding late enrichment
+// (attachments, canvas, auto-title). `message:enriched` clears it.
+//
+// The defect this module closes (jarvis#681, e2e finding F26). `message:enriched` is
+// a single best-effort realtime push at the end of a chain of best-effort enrichment
+// effects, and the affordance had NOTHING else that could ever clear it. So all of
+// these left it up permanently on a reply that was already complete and correct:
+//   * a finalize effect that keeps failing. The usage poll retries BY DESIGN when the
+//     gateway session row is missing or stale, which is exactly what a live credential
+//     apply produces, and the turn then sits in `finalizing` until the 5-minute pump
+//     watchdog has burned the 3-attempt budget. That is 15 minutes at best, and
+//     forever on any bench whose scheduler is not running.
+//   * the push itself lost: a socket blip, a hidden tab, a client that was on another
+//     route when it fired. There is no replay on the socket.
+//   * a reload. `loadConversation` replaces the messages but never the pending set,
+//     so navigating away and back re-rendered the same stuck line.
+//
+// A permanent "Finishing…" asserts the answer is incomplete when it is not, which is
+// strictly worse than saying nothing. So the affordance is BOUNDED: every entry
+// carries a deadline, and on expiry it is dropped and the owner is told, so it can
+// refetch the conversation once and still pull in whatever enrichment did land.
+//
+// The bound is deliberately generous. Enrichment normally completes within a couple
+// of seconds; the slowest healthy path is the usage effect's bounded gateway poll
+// (three reads with 1.5s sleeps) plus short-queue latency. Two minutes is far past
+// any healthy run, and short enough that nobody reads the line as permanent.
+export const ENRICHMENT_PENDING_MAX_MS = 120000;
+
+const noop = () => {};
+
+/**
+ * Track which assistant messages are still awaiting `message:enriched`, with a
+ * deadline on each so the affordance can never outlive its usefulness.
+ *
+ * The tracker owns the timers; the caller owns the rendering. `onChange` is handed a
+ * fresh Set on every mutation (so a Vue ref assignment triggers a re-render, matching
+ * how ChatView already replaced the Set rather than mutating it), and `onExpire` is
+ * called once per message that timed out, AFTER `onChange`, so the owner sees the
+ * cleared state before it decides to resync.
+ *
+ * `setTimer` / `clearTimer` are injectable purely so the deadline behaviour is
+ * testable without real time passing.
+ */
+export function createEnrichmentPending(options = {}) {
+	const maxMs = options.maxMs != null ? options.maxMs : ENRICHMENT_PENDING_MAX_MS;
+	const onChange = options.onChange || noop;
+	const onExpire = options.onExpire || noop;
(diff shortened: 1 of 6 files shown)


kavin-114:
## The other half of #681: why a background lane kept stale credentials

Not fixed here, and not fixable here. Recording it so it can be filed as its own issue against the right plane. This PR deliberately does not depend on it: the "Finishing..." bound holds whether or not a lane ever 401s again.

### What the apply actually does

A plain API-key rotation on a live tenant goes through `PUT /v1/containers/{name}/rotate-secret` in `jarvis-fleet-agent/jarvis_fleet_agent/main.py:3815-3876`. It writes `/secrets/llm.key` transactionally, then fires an in-container `secrets.reload` RPC (`secrets_reload.py:45-72`, over `docker exec` because the gateway strips operator scopes from host-side connections). It does **not** restart the container, and it does **not** run `session_hygiene.sweep_stale_main_sessions`. That sweep only runs on the `llm-creds` and `llm-pool` paths, inside their stop window.

So after a rotation, every session that existed before the apply is still alive, untouched, in a process whose in-memory credential was refreshed only if `secrets.reload` genuinely busted every cached client.

### Why that is a live suspicion, not a theory

There is a dated precedent in fleet-agent's own source, `main.py:3301-3306`: openclaw's config file watcher re-applies `openclaw.json` without re-resolving file secrets, so after a pool apply rotated the virtual key "the gateway keeps signing with the REVOKED vk: every turn dies at bifrost as 401 virtual_key_not_found". The fix there was to call `secrets_reload.reload_secrets()` explicitly alongside the watcher path. Same failure class, different entry point.

`session_hygiene.py:1-46` documents the second half: openclaw records a session's provider in the session JSONL and re-applies it on resume, and the built-in heartbeat resumes the persistent `agent:<id>:main` session on every tick. A pre-existing main session therefore survives `rotate-secret` completely untouched and keeps being resumed. That is the best candidate for the "background lane roughly a minute later" in the report, better than a bench short-queue job.

### What is ruled OUT

- `jarvis-openclaw-plugin` holds no credentials, no sessions and no lane concept. `git grep -i lane` over the whole repo returns nothing.
- The bench's post-turn jobs do not reuse the customer's `session_key`. `jarvis/chat/title.py:165-166` and `jarvis/chat/suggestions.py:220-221` each mint a fresh throwaway session per call, so a stale per-session pin cannot explain them directly.
- `jarvis/chat/agent_session_pool.py` is a long-lived object (240s idle ceiling, close to the reported window) but it only carries the bench-to-gateway device-paired transport, not any LLM credential.
- `jarvis/chat/wiki.py maybe_nudge` and `jarvis/chat/proactive.py` make no LLM call at all.

### Where to file it

`jarvis-fleet-agent`, as: `rotate-secret` hot-reloads secrets but never sweeps or invalidates sessions that predate the apply, so a resumed lane can keep 401ing against the old credential. The decisive open question sits in the openclaw image, which is not in any repo on disk: does `secrets.reload` invalidate a per-provider resolved client for sessions and runs that were already open or already warmed by `prewarm.py`? Answering that decides whether the fix is a session sweep on the rotate path, a stronger reload, or an upstream report.

### A smaller, separate bench-side gap

`jarvis/chat/title.py` pins model and provider for the title lane, deliberately giving up its failover chain (the `#531` comment at `title.py:174-177`), and then swallows every failure into `derive_title` (`title.py:213-219`, plus the blanket handler in `autotitle_job` at `title.py:268-272`). `suggestions.py` follows the same shape. So a provider-auth problem that only affects the pinned model is invisible to the customer and shows up for an operator only as an Error Log row. That is arguably correct product behaviour (a title must never break chat) but it does mean this class of credential fault has no signal at all. Worth its own small issue about surfacing it, not a change to make silently inside this PR.

Issue #691, opened 2026-08-05
A ready workspace is blocked behind "Finish setting up Jarvis" because the readiness verdict is memoized and never invalidated on a successful in-place connect


kavin-114:
After successfully connecting an AI model via Retry, the app showed the full-screen setup gate saying the workspace has no AI agent, while the backend reported it fully chat-ready. Observed on `e2e.localhost` immediately after #690's retry succeeded.

## The contradiction

Backend, at the moment the poster was on screen:

```
is_ready_for_chat()   ->  {"ready": true, "reason": null}
llm_provider          ->  "openai_compat"
llm_model             ->  "claude-sonnet-4-6"
llm_api_key           ->  present
last_sync_status      ->  "ok (restart via admin)"
```

Frontend, same moment:

> **Finish setting up Jarvis**
> This workspace isn't connected to an AI agent yet. Complete a short setup to start chatting with Jarvis about your ERPNext data.

No errors logged after 04:30. A hard reload cleared it and went straight to chat.

## Cause

`onboarding/readiness.js` memoizes the verdict in a module-level `readyPromise` so the router guard and the AppShell gate share one round trip. That is deliberate and fine. The problem is invalidation: `forgetReady()` runs on the navigate-to-chat path, but a connect that succeeds WITHOUT navigating leaves the memo holding the `ready:false` answer captured before the save landed. The gate then renders from a verdict that is stale by seconds.

The module comment already anticipates this shape: *"It does NOT cover changes made in-app, which is why forgetReady() below exists: connecting or disconnecting a model from the settings dialog changes this verdict without leaving the page."* The retry path is another such case and does not call it.

## Why it matters

This is the exact mirror of the bug fixed in #676. That one navigated INTO a chat that was not ready; this one BLOCKS a chat that is ready. Same stale-memo root, opposite direction, and this one is arguably worse for a new customer: they have just paid, just connected a model, and are told the workspace is not set up. The only escape is a hard reload, which nothing on screen suggests.

"Complete setup" also loops them back into onboarding they have already finished.

## Suggested direction

Invalidate the memo wherever a connect outcome could change the verdict, not only on the navigate path. Anything that reaches a terminal apply state (ready, rejected, retried-to-success) should call `forgetReady()` before the gate is next consulted. A cheap belt-and-braces would be for the gate itself to re-probe once before rendering the poster, since showing it wrongly is far more costly than one extra read.

https://claude.ai/code/session_0172HQoE7QCUQvEQLK8XQaLz

[...]

Issue #690, opened 2026-08-05
A failed AI-connect save renders as "Still finishing setup", after a 5 minute spinner, when nothing was saved at all


kavin-114:
Connecting an AI model failed on a network error, and the wizard showed the setup animation for five minutes and then told the customer it was "still finishing on its own". Nothing had been saved and nothing was finishing. Observed end to end on `e2e.localhost`.

## What happened

Picked a model, clicked the connect CTA, watched the `SetupNeuralNet` animation for about five minutes, then landed on:

> **Still finishing setup**
> Setup is taking longer than usual. It's still finishing on its own - you can keep waiting or retry.

Meanwhile the save had already hard-failed, from the site's Error Log:

```
04:29:39  admin_client: network error
          url='http://jarvis.admin.v2:8002/api/method/jarvis_admin_v2.api.tenant.rotate_llm_secret'
          ConnectionError(ProtocolError('Connection aborted.',
                          RemoteDisconnected('Remote end closed connection without response')))
04:29:39  Jarvis: admin unreachable
```

And nothing persisted:

```
llm_provider: ""      llm_model: ""      llm_api_key: absent
last_sync_status:     "disconnected"
llm_pool_synced_at:   None
llm_direct_synced_at: None
```

## Why the copy is wrong

"It's still finishing on its own" is the `status.timedOut` branch in `OnboardingView.onTerminal`, whose premise is that the deadline released us while the job continues server-side. That premise is false here: the request never landed, so there is no job. The customer is told to wait for something that does not exist, and the honest answer ("that didn't save, try again") was available at 04:29:39, roughly five minutes earlier.

The five minutes is `createOperationController`'s `deadlineMs` (`5 * 60 * 1000`).

## Suggested direction

A hard transport failure on the SAVE is not the same as a transport failure on a status READ. The read case is deliberately absorbed and retried ("a transport error on the READ is not a verdict", `lib/llmOperation.js`) and that is correct. A connection error on `rotate_llm_secret` should instead surface promptly with a real retry, rather than being absorbed into the poll-until-deadline path.

Worth checking as part of this: whether an apply that never opened an operation can be distinguished from one that did, since only the latter can honestly claim to be finishing on its own.

## Retry does work

Clicking Retry completed the save (`last_sync_status: "ok (restart via admin)"`, `ready: true`), so the underlying error looks transient. The defect is that five minutes and a misleading sentence sit between the failure and the customer learning about it.

[...]

PR #747, opened 2026-08-08
fix(onboarding): clear stale tenant authority on a new admin login

### LINKED ISSUE #693: Pool-to-tenant promotion does not bump the authority generation, so the divergence guard pins the site to the dead pool container forever
When a paid signup is promoted off a pooled container onto its own tenant container, the authority generation is not incremented. The divergence guard therefore sees a new container handle arrive at the SAME generation, correctly refuses to switch, and pins the site to the pool container it no longer owns. Chat then queues forever with no error shown to the customer.

Reproduced end to end on `e2e.localhost` tonight.

## Evidence

Ports, at the moment chat was stuck:

```
site  Jarvis Settings.agent_url        ws://127.0.0.1:19004
                                        ^ jarvis-pool-796302   (pool, created 00:44)
running dedicated tenant               jarvis-tenant-374uk8t72l-773a70
                                        127.0.0.1:19002->18789, 19003->18790  (created 04:28:58)
```

Stored authority, still pointing at the pool container:

```
tenant_authority_handle:      28a001342a7dccb9a3f9c9ae28b3503c   <- pool
tenant_authority_generation:  1
```

Error Log at **04:28:58**, the same second the dedicated container was created:

> `tenant authority: held connection on equal-gen handle divergence` - a different container (`8432565ce1b92b3e110871bae82271fc`) than the one already accepted at that generation (`28a001342a7dccb9a3f9c9ae28b3503c`); held the current connection and will re-poll, never downgraded (review plan 04 P0-5).

## Why this is a contract bug, not a guard bug

The guard behaved exactly as designed. A new handle at an unchanged generation is indistinguishable from a downgrade attempt, so holding the current connection is right. The defect is upstream: **promotion from a pooled container to a dedicated tenant is a legitimate authority change and must advance the generation.** Because it does not, a valid promotion is shaped identically to the thing the guard exists to reject.

"will re-poll" never resolves either: every re-poll returns the same equal-generation divergence, so the site stays pinned indefinitely.

## Blast radius, and why it is not cosmetic

This is on the paid-signup path, right after provisioning. Downstream symptoms, all observed:

1. Every chat turn stays `state: queued` with `creation == modified`. Nothing dispatches it.
2. `AgentSession.connect` targets the pool container, which no longer accepts this site's credentials, so each attempt fails auth.
3. The repeated failures trip the pool container's throttle: `AgentUnreachableError: connect rejected: INVALID_REQUEST: unauthorized: too many failed authentication attempts (retry later)` (`chat/agent_client.py:1213`, via `pump.make_mux`).
4. Restarting the TENANT container does nothing, because the tenant is not the container being contacted. Its log stays clean throughout, which makes this very hard to diagnose from the obvious place.
5. The customer sees only "Queued - ~1 ahead", forever. The real cause never reaches the UI.

Chat demonstrably worked from the pool container earlier (turn `3cd0b4a6160c`, `done` at 01:43), so this breaks a workspace that was previously functional, at the exact moment it becomes a paying one.

## Suggested direction

- Bump the authority generation as part of the pool-to-tenant promotion, so the new handle is adopted rather than refused.
- Make "held on equal-gen divergence" self-limiting: if re-polling returns the same divergence N times, it is not a transient race and needs to escalate rather than hold silently forever.
- `_attempt_connect(..., allow_repair=True)` does not treat this rejection as repairable, so the self-heal path never fires. Related to the dead self-heal classifier noted in the device-token-mismatch work.
- Surface it: a turn that cannot connect should fail visibly instead of sitting in `queued`. Same family as #690 and #691 - a real failure rendered to the customer as progress.

## Not yet traced

Which code path performs the promotion and why it leaves the generation at 1. That is the next step and it is on the admin/fleet side.

Found while e2e-ing the onboarding work (jarvis#676/#687, admin-v2#242/#244/#252).

https://claude.ai/code/session_0172HQoE7QCUQvEQLK8XQaLz

kavin-114:
Fixes #693

## Summary

A bench pointed at a different admin customer no longer gets stuck serving the previous
customer's container. Chat stopped queueing forever with nothing shown to the user.

The reported symptom was real and current. The cause was not where the issue placed it:
admin-v2 bumps the authority generation on every write, without exception. The defect is
bench side.

## What changed

- `write_connection` clears the stored tenant-authority receipt when a new admin
  principal's credentials arrive, alongside the bearer cache and chat-ready markers it
  already reset for the same reason.
- Deliberately narrower than the existing `principal_change`: a standalone
  `customer_password` is re-served on every poll and is not proof of a new login, so
  treating it as one would drop the guard on a healthy connection for no gain.

## Risk and verification

- Five tests. Three fail without the fix, including the realistic two-call shape of
  signup followed by finish payment. Two are regression guards that a plain agent URL
  sync and a standalone password do not clear authority.
- Not run locally: the bench serves the main checkout rather than this worktree, and CI
  is the authority for this repo.

<details><summary>Root cause</summary>

The authority generation is a PER CUSTOMER counter, so every fresh customer's first claim
is generation 1.

A bench that held a real connection (generation 1, handle of container A) and is then
pointed at a different admin customer receives generation 1 again, with the handle of
container B. The guard compares two unrelated customers' counters, sees equal generation
with a different handle, and raises `AuthorityInvariantError`. It can never clear,
because a fresh customer's generation cannot numerically exceed the stale leftover.

Verified against the live admin plane: all three writers of `authoritative_tenant` bump
the generation atomically in the same statement, and every promotion path calls one of
them in the same transaction. There is no missing bump.

The known trigger is a one-sided dev reset, an admin-side customer purge without the
matching bench-side reset, which is exactly the recipe `jarvis/dev.py:reset_onboarding`
documents as two steps. The invariant was wrong for any path that hands a bench a new
principal without a reset, not only that one.

Not implemented here, deliberately: the issue also asks to escalate a persistent
equal-generation hold and surface it in the UI instead of showing Queued forever. This
fix removes the only identified trigger for a permanent hold; the escalation work belongs
with the #690 and #691 silent-failure family.

Two admin-v2 hardening items were found and are NOT fixed here, both latent with no known
trigger: `drain.py:_cutover_under_fence` reads `customer` before the fence and can skip
the authority stamp while the promotion writes above it are unconditional, and
`api/tenant.py:_connection_payload` pairs a generation and a handle that are not
guaranteed to describe the same row.

</details>

### DIFF
--- a/jarvis/onboarding.py
+++ b/jarvis/onboarding.py
@@ -112,6 +112,43 @@ def write_connection(data: dict) -> None:
 	# reconnect that repoints it can be told apart from a daily sync that rewrites
 	# the same URL (which must not disturb an established claim).
 	_old_agent_url = (s.get("agent_url") or "").strip()
+	# Credentials just changed (fresh signup, or a reconnect rotating onto another
+	# account): a bearer minted from the old ones would outlive them, and, the part
+	# that used to be missing, the accepted tenant-authority (generation, handle)
+	# must not outlive them either. The generation is a PER-CUSTOMER namespace
+	# (admin-v2 fleet/_tenant_lookup.py: "the generation is an ASSIGNMENT generation,
+	# per customer"), so a stored value left over from a PREVIOUS principal is not a
+	# stale version of the new principal's count, it is a count from an unrelated
+	# series - every fresh customer's first claim is generation 1, so a leftover
+	# stored 1 (or higher) collides with the new customer's own first claim at the
+	# SAME number but a DIFFERENT container handle. guard() cannot tell that apart
+	# from the real invariant breach it exists to catch, so it HOLDS the stale
+	# connection and re-polls forever: every retry carries the same new-customer
+	# generation 1, which can never numerically exceed the stale leftover (jarvis
+	# #693, reproduced end to end on a dev bench re-signed-up after an admin-side
+	# customer purge without a matching bench-side reset). check_account_reconnect
+	# and _disconnect_agent_transport already clear for exactly this reason on their
+	# own paths; plain signup (and anywhere else a fresh principal's credentials
+	# land here) needs the same guard, cleared BEFORE the connection block below so
+	# a payload carrying both new credentials AND a connection in one call is
+	# covered too.
+	#
+	# Deliberately narrower than the ``principal_change`` used below for the
+	# cached-bearer/chat-ready resets: api_key/api_secret/customer are the strong
+	# signal that this IS a different admin login, but a lone ``customer_password``
+	# is not. admin re-serves that field on every poll while its 60-minute cache
+	# entry lives (billing/signup.py ``_serve_signup_password``, "kept until TTL
+	# rather than deleted on read"), including a poll a fully-connected bench could
+	# still make within that window - and a lone customer_password never arrives
+	# for a GENUINELY new principal either: every real new-principal call also
+	# carries api_key/customer, which already trips this clear. Widening the
+	# trigger to customer_password would drop the P0-5 protection for one window on
+	# an otherwise healthy, unrelated connection for no gain.
+	new_admin_login = any(data.get(k) for k in ("api_key", "api_secret", "customer"))
+	if new_admin_login:
+		from jarvis import tenant_authority
+
+		tenant_authority.clear(s)
 	if data.get("api_key"):
 		set_settings_password(s, "jarvis_admin_api_key", data["api_key"])
 	if data.get("api_secret"):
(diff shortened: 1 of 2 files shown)


kavin-114:
Review note, deliberately not fixed in this PR.

`check_account_reconnect` now clears `tenant_authority` twice: its own explicit clear, plus the implicit one this PR adds via `new_admin_login` in `write_connection`. Clearing 0 to 0 is idempotent so there is no user-visible effect today.

Left as is to avoid another CI cycle while the blocker is waiting to merge. The residual risk is only that the two clear sites are no longer derived from one source of truth, so a later edit narrowing either condition could diverge without a test catching it. Worth folding into whoever next touches this function.

A second candidate was raised and REFUTED on verification: that the new clear fires on a resumed or duplicate signup for an already connected paying customer, opening a window with no guard. That is not the case.

PR #750, opened 2026-08-08
fix(chat): streaming render, typed card approval, duplicate notification

l0gesh29:
Four fix-wise chat-surface commits.

## Commits

**`fix(chat): stop a streaming reply flashing its raw block JSON`**
The block-strip patterns matched on the CLOSING fence, so while a block was still being written nothing matched and marked drew it as a code block. The reader watched a wall of raw JSON type itself out and vanish. It hit all eight block types. Blocks are now held from the opener while streaming and released once settled, so a block the agent never closed is still shown rather than silently swallowed. `stripBlocks` moves to `lib/chatBlocks.js`; `streaming` joins the render cache key.

**`fix(chat): pace streamed text and fence stale tool events`**
`assistant:delta` carries the cumulative reply and the view assigned it straight through, so text moved at whatever rate the socket delivered it. A reveal cursor now paces it, snapping to full on every path that could truncate: `run:end` before `streaming` clears, a recovered-run rewrite, stop/error/conversation switch, and a hidden tab where rAF stops. Painting runs at ~25/s rather than 60/s because each paint costs a full markdown re-parse plus a `v-html` rewrite; skipped frames are paid back so the speed is unchanged.

Also fences stale tool events. The CDX-3 pump fence deliberately lets an epoch-less tool event through, and `run:end` clears `activeTools`, so a straggler `tool:start` pushed a `running` entry no `tool:end` would settle, leaving a spinner nobody opened. These land together because the teardown they both touch is one contiguous insertion.

**`feat(chat): approve a confirmation card by typing, including in bulk`**
"go ahead" now approves the parked card, and with several parked it approves all of them; "confirm 1 and 3" approves exactly those by the number on each card. See the security note below.

**`fix(notify): signal a re-published terminal once`**
A finished reply produced two "Reply ready" toasts. A terminal is published more than once (settlement, then the finalize backstop re-publish). ChatView has always deduped it one-shot; the global notifier had no fence at all. It now uses the same fence from the same module, with its own state.

**`fix(compat): run on Frappe 15 as well as 16`**
`pyproject` declares `frappe >=15,<17`, but three APIs that moved between majors were written against 16 and shipped onto 15 benches, where they raised at call time: `File.get_content(encodings=)`, the openpyxl-to-xlsxwriter `xlsxutils` rewrite, and ERPNext's `get_itemised_tax` signature. The `get_content` one was swallowed by a surrounding `except`, so every chat attachment and `read_file` call silently extracted to empty text. Each shim probes a capability rather than branching on `frappe.__version__`.

## Security note on typed approval

This touches the write-safety gate, so it deserves a careful read.

It is a convenience over the Confirm button, never a widening of it. A message must satisfy all of: the WHOLE message parses as an approval, every number named exists, the confirmation store can answer, and the caller is a human interactive session with no attachments. Anything less falls through to the model as ordinary text. `"yes but change the quantity to 5"` approves nothing.

Matching is whole-message equality against a short fixed list. Substring matching would be a security regression on the one gate between the model and a real ERP write.

Per card, the guarantees are identical to the button: same authenticated session user, same owner-bound single-use `consume`, same `exec_user` scope, same receipt chip. Bulk runs N independent confirmations; it creates no path that skips any of them. The model cannot reach this at all, arriving only through the human session endpoint and never the plugin callback. A typed approval racing a click resolves to one winner inside `consume`.

`confirm_tool`'s body moved to `_confirm_core` so both paths share one implementation and neither can weaken the gate.

## Reviewer notes

- **Ordering is load-bearing** once a number selects a card. The store is a Redis SET with no order, so the server and all three clients sort by `(expires_at, token)`, comparing tokens by code unit to match Python's byte order. `localeCompare` disagrees on mixed-case tokens.
- **All three clients were updated.** The SPA, PWA and Desk widget each render these cards; without handling the confirmed response a typed approval would hang their spinner waiting for a run that never comes.
- **Two loose allowlist entries** worth a second opinion: `all` on its own, and `ok`/`sure`/`y`. Safe only because of whole-message matching. Easy to trim.
- **A known bounded window:** numbers can shift between a glance and a send (TTL expiry, a stopped run clearing its tokens, a resync adding one mid-stack). The ordering makes renumbering consistent rather than arbitrary; `consume` being owner-bound and single-use bounds the worst case to confirming a different card the same user owns and can see.

## Testing

- Frontend: **1000 passing across 66 files**, including 18 block-strip, 23 reveal, 10 typed-approval and 9 notifier tests.
- The notifier tests were verified non-vacuous: with the fence removed, exactly the three dedupe tests fail.
- Bench: 32 tests for typed approval, plus `test_compat_frappe_versions.py` for the shims. **These were not run against a site for this PR** and need CI to confirm.
- Two pre-existing failures in `test_confirm_gate` and `test_pending_confirm` were reproduced on a clean `develop`, so they are unrelated to this branch.

---

**Update:** the Frappe 15/16 compatibility commit was moved OUT of this PR to `fix/frappe-15-16-compat`. It was the only red check: 3 errors in `test_compat_frappe_versions.py` (the v15 permission predicate builds over the unaliased table, so an aliased query fails with `Unknown column 'tabToDo.allocated_to'`; and `itemised_tax` on an unsaved taxless doc hits `_item_wise_tax_details = None`). It will come back as its own PR once fixed. This PR is now the four chat-surface fixes only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,18 @@ jobs:
         # for `node --test`, so this is inherently a one-shot gate.
         run: npm run test:node
 
+      - name: node:test (Desk widget + PWA lib)
+        env:
+          NODE_OPTIONS: --unhandled-rejections=strict
+        # These two suites are plain node:test with only sibling-module imports (no npm deps),
+        # so they run from the repo root with bare node - no `npm ci` needed. They were run by
+        # NOTHING in CI before, which is exactly how a Desk-widget confirmation-card ordering
+        # bug (a dropped field that renumbered a typed "confirm N" onto the wrong ERP write)
+        # shipped under a green build. Gate them here.
+        run: |
+          node --test jarvis/public/js/jarvis_chat/widget/*.test.mjs
+          node --test pwa/src/lib/*.test.js
+
   # The frontend-tests job above runs the SPA's *tests* but never *builds* it — so a Vite/build break
   # would slip through green and only surface at Frappe Cloud deploy time as a 404 on /jarvis. This gate
   # builds the SPA (the /jarvis artifact) the way the deploy does. (The PWA build resolves the bench site
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -205,13 +205,19 @@ export const dismissTool = (token, conversation) =>
 export const listPendingConfirmations = (conversation) =>
 	call(AC + "list_pending_confirmations", { conversation: conversation || "" });
 
-export async function sendMessage(conversation, message, modelOverride, attachments, context) {
+export async function sendMessage(conversation, message, modelOverride, attachments, context, approvalTokens) {
 	// Empty conversation is allowed: the backend creates (or focuses) an empty
 	// conversation itself and returns its id as `conversation_id` - saves the
 	// SPA a createOrFocusEmpty round-trip before the first send (latency plan).
 	const args = { conversation: conversation || "", message };
 	if (modelOverride) args.model_override = modelOverride;
 	if (attachments && attachments.length) args.attachments = JSON.stringify(attachments);
+	// The ordered tokens of the confirmation cards on screen. A typed "confirm 2"
+	// selects by the number the user sees, so the server must resolve that number
+	// against THESE tokens (what was displayed) rather than a list it re-fetches -
+	// otherwise a card expiring between glance and send renumbers the rest and the
+	// wrong card runs. Server ignores it unless the message parses as an approval.
+	if (approvalTokens && approvalTokens.length) args.approval_tokens = JSON.stringify(approvalTokens);
 	// Forward context for a viewing-context doc/report, the one-shot "ground on
 	// wiki" flag (which can arrive without a doc), OR a page marker ("triggers" /
 	// "dashboards") that primes the agent for that surface's flow from the main
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -205,7 +205,14 @@ export const dismissTool = (token, conversation) =>
 export const listPendingConfirmations = (conversation) =>
 	call(AC + "list_pending_confirmations", { conversation: conversation || "" });
 
-export async function sendMessage(conversation, message, modelOverride, attachments, context, approvalTokens) {
+export async function sendMessage(
+	conversation,
+	message,
+	modelOverride,
+	attachments,
+	context,
+	approvalTokens
+) {
 	// Empty conversation is allowed: the backend creates (or focuses) an empty
 	// conversation itself and returns its id as `conversation_id` - saves the
 	// SPA a createOrFocusEmpty round-trip before the first send (latency plan).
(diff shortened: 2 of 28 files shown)


venkat102:
## Review fixes pushed (`1e3c1b0b`)

A multi-lane review of the typed-approval surface found two paths to a **wrong ERP write** plus a set of correctness/observability/test gaps. Fixes are in the commit above; summary below.

### Critical — both closed

**1. Desk widget dropped `expires_at`, so numbered typed approval could run the wrong card.**
All three clients carry the correct `(expires_at, token)` comparator, but the Desk widget's `action:pending` handler and resync map built cards as `{token, tool, summary}` — no `expires_at`. Its sort silently fell back to token order, disagreeing with the server, so two cards minted seconds apart could be numbered in reverse and a typed "confirm 2" ran a different card than shown. Now carried on both Desk paths, with regression tests.

**2. Typed "confirm N" was position-bound, not token-bound.**
The number was resolved against a card list the *server re-fetched at send time*. A card expiring between the user's glance and their send renumbered the rest, so "confirm 1" could land on a more-destructive write. The button never had this — it carries a specific token. Fix: the client now sends the ordered tokens it displayed (`approval_tokens`), and the server confirms *those exact tokens*. A stale/foreign token fails its own card inside the owner-bound single-use `consume` and can never substitute another; if none of the named tokens is live, the message falls through to the model.

### Important — also fixed

- **Hint:** no longer prints "confirm 1 and 3" for a 2-card stack (out-of-range → no-op); now also shipped on **PWA + Desk**, not only the SPA.
- **Desk partial batch:** showed a false "no longer valid, ask again" next to receipts of writes that already succeeded — now surfaces the server's real "N of M went through" message (as the PWA already did).
- **Reveal cursor:** `loadConversation` now flushes it before swapping in resync'd rows, so a missed run terminal can't leave a permanently-truncated answer with no spinner.
- **PartialConfirmation** names which card failed, by its number.
- **Observability:** confirm log titles are entry-point-neutral with a `token`/`conversation` correlator; a store outage that degrades a typed approval now logs one greppable line.
- **Tests/CI:** the card-order comparator is extracted to a shared, unit-tested module per client; the vacuous source-grep "sort test" is replaced with real behavioural tests; the entry gate is extracted and truth-tested; and **the Desk + PWA `node:test` suites now run in CI** — they gated nothing before, which is exactly how the Desk ordering bug shipped under a green build.
- `send_message` docstring documents its two return shapes (`confirmed:true` vs an ordinary send).

### Not verified here (please confirm in CI / flow review)

- **Bench tests** (`test_typed_confirmation.py`) need a live site — the decision logic was verified standalone, but the full suite runs in CI.
- **SPA/PWA vitest** need `npm ci`; the **node** suites ran green locally (Desk 132, PWA 38).
- **Flow review owed:** the new PWA/Desk hints and the message wording are UI I styled with existing tokens but did not render — they want a real browser pass before merge.
- **Residual (follow-up, not done):** a proactive "that card is no longer available" notice when a typed approval falls through (e.g. Redis down / swept card). C2 makes it reach the model and it's now logged, but there's no user-facing signal yet.

PR #751, opened 2026-08-08
billing: one payment renews and arms autopay; return to billing, not the wizard

Navin-S-R:
Tenant half of the mandate-recovery work. Admin side:
Aerele-RnD/jarvis-admin-v2#282 — **merge that first**, this depends on its endpoints.

## What changes for the customer

- **A lapsed customer gets a reactivation grid** — every plan on their billing cycle,
  up *and* down, at full price. A lapsed period has nothing to prorate, so up and down
  are the same purchase. The plan they pick rides the payment intent and is never
  staged on the subscription, where an abandoned attempt would mis-price the next one.
- **Coded billing answers render their own copy and actions** instead of a silent
  reload, so a stuck checkout is recoverable from the page it happened on.
- **The pay page returns here, not to the signup wizard.** The wizard has one code for
  every lapsed status — *"This signup cannot be continued… renew from your account"* —
  and no way to renew. A renewal was sending customers somewhere they could not renew.
  Arriving back after paying runs the healer once, so the page shows truth rather than
  the state they just paid to leave.
- **The paid notice speaks for this surface.** An existing customer who renewed is not
  "continuing their setup" — that is signup-wizard copy, shared through
  `paymentCodes.js`. And `Continue` did nothing (`runCodeAction` had no branch for it,
  so it fell through to a reload); it is now **Back to chat**.

## Also carries the team workflow

`.claude/` — the plans this was built from and the review records for all three
adversarial rounds, including the RED flow verdict this shipped with. Kept in the repo
so the decisions and the known gap are auditable rather than living in a chat log.

## Review status — please read before approving

**Every code review finding is fixed.** **The flow review is RED**: three scenarios
(signup autopay through the new redirect, pay-twice-across-a-lost-confirm, and a REAUTH
mandate through `/return`) could not be executed, because completing them requires
entering card details. Merged on the owner's explicit instruction with that gap named.

Tests: vitest 62 files, node 615.


Navin-S-R:
Adds the onboarding dead-end fix (`ce2f07ef`).

Navigating to `/jarvis/onboarding` on an expired site showed *"…Renew from your account"* above a single **Contact support** button — the copy named the way out and the screen did not provide it. Same defect as the pay-page return already fixed on this branch, reached by typing the URL instead of coming back from the gateway.

`SIGNUP_TERMINAL` now offers **Renew your plan** as its primary action, routing to the billing page. The wizard deliberately does not grow its own plan grid and payment: renewal already exists on billing *with* the cohort rules (`can_reactivate`, the same-cycle grid, the order-vs-mandate shape decision), and a second checkout-initiation path for the same money is how two surfaces drift apart. A test pins that this code never gains `INITIATE`/`RESUME`.

node 675 · vitest 63 files · pinned prettier 2.7.1.

PR #775, opened 2026-08-12
fix(chat): show the full subscription model list in the chat picker

kavin-114:
## Summary

A ChatGPT-subscription tenant only saw the one model id saved in their pool in the chat model picker, even though the catalog carries the whole subscription tier. This shows every subscription-tier model for a connected subscription, so the customer can switch models in chat without re-saving Settings.

Root cause: the picker (and its pin validation) both source catalog models from `_catalog_models_for_pool`, which was api-key only. It gated on `accepts_any_model` (False for subscription rows) and emitted only `tier == "api_key"` rows, so a subscription tenant got an empty catalog and a single-model menu.

The fix widens that one function to also emit `tier == "subscription"` catalog models for every provider the tenant has a connected subscription account for, keyed by the account upstream exactly as the pool build derives it, so the frontend join needs no change. Because `_allowed_pin_models` reads the same function, the picker and pin validation stay in lockstep. Backend-only; adds a metadata-only `agent_provider_for` helper to keep it cheap on the chat hot path.

<details>
<summary>Note on runtime verification</summary>

Unit-tested (picker output + pin validation, including the OpenAI two-tier merge). Not yet exercised against a live ChatGPT-subscription tenant: whether a pinned non-primary subscription model is actually served by cliproxy vs silently misrouted to the primary should be confirmed on a real sub tenant before relying on it.
</details>

## Pre-merge checklist

- [x] New/changed behavior has tests (picker + pin validation + merge case)

### DIFF
--- a/jarvis/chat/api.py
+++ b/jarvis/chat/api.py
@@ -1611,49 +1611,96 @@ def _api_key_models() -> dict[str, list[dict]]:
 
 
 def _catalog_models_for_pool(settings) -> dict[str, list[dict]]:
-	"""Provider id -> api-key catalog models the chat picker may offer, for the
-	providers this tenant has ALREADY configured.
+	"""Provider id -> catalog models the chat picker may offer BEYOND the exact ids
+	saved in the pool, for the providers this tenant has ALREADY configured.
 
-	The point is that a customer with one saved model is not stuck with it. The
-	container holds a credential per provider, and the agent serves a model id the
-	pool spec never named, so switching within a configured provider costs nothing
-	and needs no re-save.
+	The point is that a customer with one saved model is not stuck with it: the
+	container already holds the credential and serves a model id the pool spec
+	never named, so switching within a configured provider costs nothing and needs
+	no re-save. Two kinds of provider contribute, each keyed by the SAME id its
+	``pool_models`` rows carry so the UI joins them without a second lookup:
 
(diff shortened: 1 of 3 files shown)


kavin-114:
Disposition of the automated high-effort review (4 findings, all in `_catalog_models_for_pool`):

- **#0 (subscription key mismatch when a row carries an explicit `provider`)** — **fixed** in `7f43a84a`. `wanted_sub` now mirrors the `pool_models` projection exactly (explicit `provider` wins, else the raw account upstream), and a catalog entry is matched by either its `catalog_id` or its `agent_provider`, so the two projections key identically. The SPA writes `provider=""` for subscriptions today (so the common case was already correct), but the mirror makes the invariant unnecessary. Added a regression test for the explicit-Gemini case.
- **#1 (default blank upstream to `"openai"`)** — **rejected.** The finder mirrored `build_pool_payload` (the wire payload), but the picker joins on the `pool_models` projection, which keys a blank-upstream subscription row as `""`, not `"openai"`. Offering under `"openai"` would never join. Left a comment at the skip so this doesn't get "fixed" later.
- **#2 (double pass over `settings.models`)** — **folded in.** Now a single classifying pass.
- **#3 (per-row decrypt on the send path)** — **accepted as-is.** One AES decrypt of a small blob per subscription row per override-send is sub-millisecond; explicit-provider rows now skip the decrypt entirely, and caching machinery would be riskier than the cost.

PR #777, opened 2026-08-12
fix(subscriptions): show the connected account name for chat subscriptions

kavin-114:
## Summary

A chat subscription never showed *which* account it was, so onboarding displayed "Account connected" with no name, and two ChatGPT subscriptions looked identical in AI Models settings. This surfaces the connected account's name/email in both places.

**1. Backend — the account email was being thrown away.** It rides in the OAuth id_token (pool sign-in requests `openid email profile`), but `_fetch_account_email` only parsed the id_token when a provider had `userinfo: None` — which none do. For OpenAI it always called `api.openai.com/v1/userinfo` with the pool access-token, whose audience is `chatgpt.com/backend-api`, so that call fails and the email came back empty → generic "Account connected". Now the id_token email claim is preferred, with userinfo as the fallback (Google Gemini has no id_token and still uses it). General, not OpenAI-special-cased.

**2. Frontend — the settings list never showed the account.** The collapsed subscription row now shows the connected account name (first account + "+N more"), and `accountLabel()` falls back to a short per-account tag ("Account &lt;last4&gt;", mirroring how Kimi names accounts) when an older account has no stored label/email — so two accounts are never rendered identically.

Tests: 3 new backend tests (id_token-first, no-HTTP-when-claim-present, userinfo fallback, malformed id_token) and 4 new frontend tests (distinct labels for label-less accounts, email shown when present, `SUB_` token never surfaced).

<details>
<summary>Scope note</summary>

This fixes the *display* half of the duplicate-subscription complaint. Whether two same-provider/same-model subscriptions should also be *merged* into one row (the coalesce path) is being diagnosed separately against a live repro on e2e.localhost — not in this PR.
</details>

## Pre-merge checklist

- [x] Branch is **up to date with `main`**
- [x] New/changed behavior has tests (backend + frontend)

### DIFF
--- a/frontend/src/components/LlmPoolEditor.spec.js
+++ b/frontend/src/components/LlmPoolEditor.spec.js
@@ -909,10 +909,9 @@ describe("the verdict and the surrounding controls agree", () => {
  */
 describe("a subscription row's collapsed account identity (jarvis account-name bug)", () => {
 	it("gives two label-less, email-less accounts distinct identifiers via accountLabel()", async () => {
-		// Real refs are secrets.token_hex(8): lowercase hex, letters included. Ending
-		// each in a distinct letter-bearing suffix (not all-digits) is what actually
-		// exercises accountLabel()'s .toUpperCase() - digits uppercase to themselves,
-		// so an all-digit suffix would pass this test with or without that call.
+		// Real refs are secrets.token_hex(8): lowercase hex. The fallback shows the last
+		// 6 chars of the ref (lowercase, matching the server-side Kimi convention), so
+		// two refs that differ within their last 6 render as distinct labels.
 		setPool([
 			subModel("gpt-5", 0, [
 				account("SUB_aaaaaaaaabcd", ""),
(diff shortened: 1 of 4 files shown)


kavin-114:
Disposition of the automated high-effort review (5 findings):

- **#1 (id_token preferred over verified userinfo)** — **rejected.** `_fetch_account_email` runs only from `_exchange_and_build_blob`, i.e. the initial authorization_code exchange, so the id_token was minted by the provider in that same exchange and its email claim is current — there is no refresh path that could hand it a stale token. Both sources also arrive server-side over TLS (the user pastes only the code), so the "unverified vs verified" framing doesn't apply. Flipping to userinfo-first would add a guaranteed-failing HTTP call to every ChatGPT pool connect. Added a docstring note so this ordering isn't reverted later.
- **#2 (last-4 hex collision) + #4 (uppercase vs Kimi's lowercase)** — **fixed together** in `b9f07723`: the fallback now uses the last **6** chars of `account_ref`, **lowercase**, matching the server-side Kimi convention. Spec assertions updated.
- **#5 (redundant `import json`)** — **fixed**; the module already imports `json`.
- **#3 (fix is frontend-only; the data still lacks a distinguishing label)** — **deferred.** Fix A (id_token email) means new connects carry a real email, shrinking the label-less population to legacy rows, which the frontend fallback covers. The durable shape (synthesize a label server-side at capture time, like Kimi already does) is a follow-up, not this PR.

PR #783, opened 2026-08-12
feat(ai-models): show pooled subscription accounts as grouped rows

kavin-114:
## Summary

When a chat-subscription model has **2+ connected accounts**, the AI models list used to collapse them into one terse line ("devhub@aerele.in +1 more"), so you couldn't see or manage the second account without opening Edit. Now those accounts expand into **grouped sub-rows** under the model — each with its avatar, email, primary/backup order, and its own **Disconnect** (same `removeAccount` the Edit panel uses). The model row keeps its reorder arrows + Edit + Reconnect, and hovering highlights the whole card as one unit.

A **single-account** subscription (the common case) is unchanged — one clean row with the email inline.

This is **display only** — accounts still pool into one stored entry, which the fleet routing requires (two rows of the same subscription model are rejected as a duplicate). The grouping just makes the pool legible and gives per-account control.

3 new tests (single-account → one row; 2 accounts → grouped sub-rows with correct emails + primary/backup; sub-row Disconnect → removeAccount with the right index). 73/73 frontend specs pass.

## Pre-merge checklist

### DIFF
--- a/frontend/src/components/LlmPoolEditor.spec.js
+++ b/frontend/src/components/LlmPoolEditor.spec.js
@@ -1027,4 +1027,33 @@ describe("grouped subscription rows (2+ accounts)", () => {
 		expect(spy.mock.calls[0][1]).toBe(1);
 		expect(spy.mock.calls[0][0]).toMatchObject({ _uid: w.vm.rows[0]._uid });
 	});
+
+	// Regression: the grouped model row briefly shipped with only Up/Down/Edit/
+	// Reconnect, losing the ungrouped row's single-click whole-model Remove. Without
+	// it, tearing down a 2+-account model required Disconnecting every sub-row one
+	// at a time (N confirm+apply round-trips, and a partial state if the customer
+	// stopped part-way). Remove and the per-account Disconnect must coexist: this
+	// only asserts the model-row Remove wires to remove(i), not that it replaces
+	// the sub-row Disconnect (covered above).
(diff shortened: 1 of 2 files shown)


kavin-114:
Review disposition (high-effort review, 3 findings):

- **Grouped rows lost the one-click whole-model Remove** (CONFIRMED correctness) — **fixed** in `d9b9aeb6`. Restored the Remove button on the grouped model row with the same `remove(i)` + last-row-Disconnect semantics as the ungrouped row. The per-account sub-row Disconnect still coexists (Remove tears down the whole model atomically; sub-row Disconnect drops one account). Test added.
- **`removeAccount` re-resolves by `_uid` but only checks index bounds, not account identity** (PLAUSIBLE) — **deferred.** The verifier found the specific reorder race doesn't hold (`_uid` changes per reseed); the narrower `finishConnect` splice race is pre-existing (not introduced here) and out of scope for this display PR. Worth a follow-up defensive guard.
- **Grouped branch duplicates the row header + reorder markup** (CONFIRMED cleanup) — **deferred** as a follow-up refactor (factor the shared header/reorder into one sub-component). Kept out of this PR to limit risk; noted so the next edit to row markup touches both branches.

Known UX nit for a follow-up pass: when a grouped row is the pool's *only* connected model, the model-row button and each sub-row's Disconnect all read "Disconnect" (same label, different blast radius). Built to match the ungrouped row's label logic; worth disambiguating later.

PR #786, opened 2026-08-12
fix(billing): usage chart shows real tokens + formatted cost

kavin-114:
## Summary

The "Billing and metering" usage chart rendered **empty** — x-axis stuck at `0→1`, no bars, tooltip reading "Tokens 0" — and the Cost tile showed a raw `$0`. Root cause: `perModelBarSpec` read a `tokens` field that **doesn't exist** on the real payload (rows are `{model, tokens_in, tokens_out, cost_usd}`), so every value collapsed to 0.

This plots **tokens_in + tokens_out** as stacked bars per model (real data, real hover), adds an optional **Cost by model** view off `cost_usd`, and formats the **Cost tile** as proper USD (`$0.00`, 4dp for sub-cent amounts, never raw). An all-zero cost chart renders nothing rather than an empty axis.

Frontend-only. Test fixtures corrected to the real row shape (the old `{tokens}` shape was why this shipped green), with a guard so the bug can't silently return.

> Note: cost is real only for managed-proxy tenants today. Exact cost for **API-key / BYO** tenants needs a per-model pricing table — a separate follow-up (Phase 2). This PR fixes the chart + formatting for the data that already flows.

## Pre-merge checklist

### DIFF
--- a/frontend/src/charts/usageCharts.js
+++ b/frontend/src/charts/usageCharts.js
@@ -56,7 +56,9 @@ export function perModelBarSpec(perModel, metric = "tokens") {
 		// would otherwise render the same "axes with no bars" pathology this file
 		// exists to fix - chartTheme only nulls out a chart when a series has no
 		// data POINTS, and [0, 0, ...] still has points. Skip the chart instead.
-		if (!data.some((v) => v > 0)) return null;
+		// Hide only when there is NO cost signal at all (every value exactly 0).
+		// A negative value (a credit/refund adjustment) is real data, not "empty".
+		if (!data.some((v) => v !== 0)) return null;
 		return {
 			type: "bar",
 			x,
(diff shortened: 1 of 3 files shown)


kavin-114:
Review disposition (4 findings, all around negative cost / credits — which Bifrost doesn't emit today, per-request cost is ≥ 0, but cheap to harden):

- **formatUsd sign placement** (`$-12.30` → `-$12.30`) — **fixed** + test.
- **Cost chart hidden on all-negative data** (`v > 0` guard) — **fixed** to `v !== 0` (hide only when every value is exactly 0) + test.
- **Duplicated per-model emptiness guard** (BillingMeteringPane) and **duplicated numeric coercion** (perModelBarSpec) — **deferred** as minor cleanups; noted for a follow-up refactor.

PR #787, opened 2026-08-12
feat(usage): per-user token cap is now all-time, not monthly

kavin-114:
## Summary

The per-user usage cap ("Monthly limit") reset every month. A token budget can reasonably span longer than a month, so it's now a simple **all-time token cap** — enforcement compares cumulative `total_tokens` against the limit (no monthly rollover), and the control is relabeled **"Token limit"** across the DocType and both usage panes.

Kept the DocType fieldname `monthly_token_limit` (and the `admin_set_user_limit` arg) as the wire contract — no field-rename migration — with a comment that the name is legacy but the cap no longer resets. The **per-model** month-keyed limit is intentionally untouched.

Backend tests updated to total-based enforcement (65 + 8 pass); frontend labels updated (1192 specs pass).

> ⚠️ **Deploy heads-up (behavior change, not a bug):** any user whose all-time `total_tokens` already exceeds a cap an admin set expecting a *monthly* reset will be blocked on their next send once this ships (there's no monthly reset to save them). Admins who set caps under the old monthly semantics may want to review/raise them before/after deploy.

## Pre-merge checklist

### DIFF
--- a/frontend/src/components/settings/UsageAdminPane.vue
+++ b/frontend/src/components/settings/UsageAdminPane.vue
@@ -43,8 +43,8 @@
 				style="grid-template-columns: 1.5fr 1.8fr 1.3fr 0.9fr"
 			>
 				<div>User</div>
-				<div>This month</div>
-				<div>Monthly limit</div>
+				<div>Usage</div>
+				<div>Token limit</div>
 				<div>Last activity</div>
 			</div>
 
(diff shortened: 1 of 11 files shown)


kavin-114:
Review disposition (4 findings):

- **Deploy lockout: existing users over their now-all-time cap** (CONFIRMED, critical) — **fixed** via reconciliation. New patch `v2_14` bumps every existing per-user `monthly_token_limit` by that user's current `total_tokens`, so an admin's intended headroom is preserved going forward (a 100k monthly cap on a user at 500k → 600k). Nobody is retroactively over (no lockout), nobody becomes unlimited (no data loss). Fresh installs have no legacy caps.
- **Chat rejection copy still said "Monthly usage limit reached"** (CONFIRMED) — **fixed** to period-neutral "You've reached your usage limit…". The `usage_limit` reason is shared by the all-time aggregate gate AND the still-monthly per-model gate, so it can't claim either period. Grepped all copy on that reason path; this was the only one.
- **`admin_list_user_usage` still ships unread `month_tokens` fields** (cleanup) — **deferred**; a wire-contract trim, and month fields aren't globally dead (UsagePane still renders "This month").
- **`pct()` duplicated across the two usage panes** (cleanup) — **deferred** as a small shared-helper refactor.

PR #790, opened 2026-08-12
feat(billing): real LLM cost for API-key tenants

kavin-114:
## Summary

API-key / BYO-key tenants saw no cost in the Billing pane — cost was proxy-only (from Bifrost). Now `get_llm_usage()`, for an `api_key` tenant, aggregates **tenant-wide per-model tokens** (SUM over `Jarvis User Model Usage`, current month), multiplies by the admin's **per-model catalog price**, and returns the **same shape** the proxy path emits — so Phase 1's chart renders it with **no frontend change**.

- `pricing.py` — `price_for_model()` reads catalog prices (0 fallback, never raises)
- `usage.tenant_wide_per_model_tokens()` — the `GROUP BY model` aggregation (tenant-wide, no per-user filter)
- `account._direct_llm_usage()` — gated on `llm_auth_mode == "api_key"`

Scoped to API-key tenants only (chat-subscription cost is Aerele's pooled cost, not a per-tenant $/token). 61/61 tests pass.

> **Pairs with jarvis-admin-v2 #307** (the catalog price fields). Until that merges and an admin prices a model, cost reads $0 (the price helper falls back to 0) — so this is safe to merge independently; it just shows real numbers once prices are entered.

<details><summary>Two cosmetic notes</summary>

- `period` is `"YYYY-MM"` on the direct path vs the proxy path's opaque string (e.g. "1M"); the pane renders it verbatim either way.
- Gating is a strict `== "api_key"`; a blank/unset mode returns `applicable:false` (the field is `reqd` with default `api_key`, so this is an edge).
</details>

## Pre-merge checklist

### DIFF
--- a/jarvis/account.py
+++ b/jarvis/account.py
@@ -1060,7 +1060,11 @@ def get_llm_usage() -> dict:
 	require_jarvis_admin()
 	settings = frappe.get_single("Jarvis Settings")
 	if not getattr(settings, "proxy_active", 0):
-		if (getattr(settings, "llm_auth_mode", "") or "") == "api_key":
+		# Blank/unset llm_auth_mode coerces to "api_key" — the field is reqd with
+		# that default, and the two other reads in this file (account.py:688, :1511)
+		# treat a blank the same way, so a legacy row with no auth mode must still
+		# get its BYO-key cost, not fall through to the empty shape.
+		if (getattr(settings, "llm_auth_mode", "") or "api_key").strip() == "api_key":
 			return _direct_llm_usage()
 		return {
 			"applicable": False,
--- a/jarvis/account.py
+++ b/jarvis/account.py
@@ -1048,13 +1048,20 @@ def _persist_operation_probe_verdicts(status: dict) -> None:
 
 @frappe.whitelist()
 def get_llm_usage() -> dict:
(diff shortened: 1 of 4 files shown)


kavin-114:
Review disposition (5 findings):

- **Blank `llm_auth_mode` falls through to empty instead of direct cost** (CONFIRMED, correctness) — **fixed**: the gate now coerces blank → `api_key`, matching account.py:688/:1511 and the field's own default. Test added for the blank-mode routing.
- **`price_for_model` always returns 0 in production** (CONFIRMED against develop) — **by design / disclosed**: it's resolved by the paired **jarvis-admin-v2 #307**, which adds the price fields to the catalog + projection. Until #307 merges, cost reads $0 (already stated in the PR body). The bundled offline fallback stays $0 (no live prices offline) — graceful, not a regression.
- **`tenant_wide_per_model_tokens` duplicates `_per_model_totals`'s SQL** (cleanup) — **deferred**: generalizing `_per_model_totals(users=None)` touches another module; noted as a follow-up.
- **Duplicated test setUp/tearDown + price-cache-clear boilerplate** (cleanup ×2) — **deferred**: a shared test mixin, follow-up.

PR #864, opened 2026-08-15
Expose parent_document_type on create_dashboard_chart

### LINKED ISSUE #862: create_dashboard_chart cannot chart child-table fields (missing parent_document_type)
Charting a child-table field fails: grouping Sales Order Item by item_code returned InvalidArgumentError "Parent document type is required to create a dashboard chart". Frappe supports this natively via the Dashboard Chart parent_document_type field, but the Jarvis tool wrapper does not expose that parameter, so the agent cannot pass it. The agent then wrongly told the user this was a Frappe limitation.

**Steps:** 1. In chat, ask for a dashboard including an item-wise order value chart. 2. Agent calls create_dashboard_chart on Sales Order Item grouped by item_code.
**Expected:** chart created with the parent doctype set.
**Actual:** InvalidArgumentError; chart dropped.
**Repro:** conversation ves2j96q81, tool call seq 54, e2e 2026-08-15.
**Suspected location:** jarvis/tools/create_dashboard_chart.py, parameter surface.

Proposal: expose parent_document_type and pass it through; a fix is being attempted in parallel.

kavin-114:
Fixes #862

Charting a child-table field (e.g. Sales Order Item grouped by item_code) always failed with "Parent document type is required to create a dashboard chart" because the tool never exposed the field Frappe's Dashboard Chart doctype needs for it.

Adds an optional `parent_document_type` param, validated with Frappe's own `get_parent_doctypes()` lookup (real DocType, actual parent of `document_type`, read-permitted on that parent through Frappe's parent-qualified `has_permission`), threaded through to the created chart. The schema/description in `jarvis-openclaw-plugin` mirrors this tool's params, so a matching PR is up there too (Aerele-RnD/jarvis-openclaw-plugin#73, this PR merges first: the plugin's advertised param is dead until the Python side accepts it, and jarvis-side alone is already safe since dispatch filters unknown args to the function signature).

Note: creating any Dashboard Chart on a child-table `document_type` still requires the core Frappe `System Manager` role for non-admin users - Frappe's own `has_permission` hook for Dashboard Chart only bypasses its doctype-read check for that literal role, and a child table never carries its own DocPerm rows to satisfy the check otherwise. This is pre-existing Frappe behavior, unrelated to and unaffected by this PR.

## Pre-merge checklist

- [x] CI is green — the `tests` check on this PR passes (never merge on ❌)
- [x] Branch is up to date with `develop` (so it is tested against the latest code)

### DIFF
--- a/jarvis/tests/test_dashboard_tools.py
+++ b/jarvis/tests/test_dashboard_tools.py
@@ -55,6 +55,64 @@ class TestCreateDashboardChart(FrappeTestCase):
 				based_on="creation",
 			)
 
+	def test_group_by_child_table_needs_parent_document_type(self):
+		# Contact/Contact Email is a core Frappe parent/child pair - istable,
+		# no ERPNext fixture needed. Without parent_document_type Frappe's own
+		# Dashboard Chart.validate() would reject the insert; the tool should
+		# catch this itself and name the valid parent in the message.
+		with self.assertRaises(InvalidArgumentError) as ctx:
+			create_dashboard_chart(
+				chart_name=_h("JT Child NoParent"),
+				document_type="Contact Email",
+				chart_type="Group By",
+				group_by_based_on="email_id",
+			)
+		self.assertIn("Contact", str(ctx.exception))
+
+	def test_group_by_child_table_with_parent_document_type(self):
+		res = create_dashboard_chart(
+			chart_name=_h("JT Child Parent"),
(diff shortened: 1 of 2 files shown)


kavin-114:
Plugin-side schema mirror: Aerele-RnD/jarvis-openclaw-plugin#73 (merges after this one).

Issue #192, opened 2026-07-03
test: 9 settings_on_update tests fail on a fresh CI site (Frappe Singles snapshot + __Auth pollution)


kavin-114:
## Summary
When `jarvis`'s test suite runs on a **fresh site** (as in CI — see PR adding GitHub Actions), 9 tests in `jarvis/tests/test_settings_on_update.py` fail. They pass on a long-lived dev bench because that site already has the `Jarvis Settings` singleton populated. **These are quarantined with `@unittest.skip` referencing this issue** so CI can gate the rest of the suite; please un-skip as you fix.

## Failing tests (9)
- TestOnUpdateClassification: `test_max_output_tokens_change_is_noop`, `test_no_change_is_noop`, `test_only_key_change_triggers_reload`, `test_temperature_change_is_noop`
- TestOnUpdateAlwaysAdminPath: `test_on_update_routes_to_admin`
- TestOnUpdateStatus: `test_records_ok_reload_via_admin_on_success`, `test_save_succeeds_even_when_admin_call_fails`
- TestOnUpdateAsyncPending: `test_pending_status_written_before_enqueue_for_reload`
- TestEnqueueDedupAndJobId: `test_enqueue_carries_action_keyed_job_id_and_dedupe_flag`

## Root cause
`_classify_llm_change` (jarvis_settings.py) returns `restart` for **every** save because `flags.llm_auth_mode_changed = has_value_changed('llm_auth_mode')` is `True`. On a fresh site the `llm_auth_mode` row is absent in `tabSingles`, so `get_single` applies the `api_key` default **in memory**, while the `for_update` snapshot loaded during `save()` reads it back as `None` → `has_value_changed` is `True` → misclassified as structural `restart` (tests expect `reload`/`noop`). Compounded by `__Auth` password pollution: a sibling test (`test_onboarding_sync`) writes `sk-test` to `llm_api_key`; `db_set` on a Password field doesn't touch `__Auth`, so `get_password` keeps returning `sk-test` (signature `'sk-test' != 'sk-persisted'`).

## Attempted fix that did NOT hold in CI
Pinning `llm_auth_mode='api_key'` in `_reset_settings` + rewriting password fields via `remove/set_encrypted_password` fixed it on a **populated** dev site but had zero effect on a fresh CI site (the `for_update` snapshot still reads the field as absent). Left in place as reasonable test hygiene; insufficient alone.

## Recommended fix (team decision — changes prod behavior)
[...]

PR #874, opened 2026-08-15
Live dashboard-build progress card + click-through thumbnail in chat

kavin-114:
## Summary

Main chat now shows a real "Building dashboard..." progress card while the agent builds a Jarvis Dashboard, and the finished canvas renders as a compact clickable thumbnail whose whole card opens the Dashboards builder with that canvas loaded.

Two new pure functions in `frontend/src/lib/dashboardBuildCard.js` (unit tested) decide the turn gate, the phase from real tool events, and the thumbnail scale; `ChatView.vue` wires them in without touching backend files. Both pieces are mutually exclusive with their existing generic counterparts (the "Working on it..." line, the generic file card), so no other flow changes.

The embed-marker assumption held, contrary to my first pass below; the deep-link assumption held outright; the render-component assumption did not hold. Details below, including a correction made after the first review round.

<details>
<summary>Investigation findings (with a correction)</summary>

1. **Corrected.** The `[embed ref="<document-id>" title="..." /]` marker is real and handled: `jarvis/chat/canvas.py:70-71,74-75,197-251` (`_EMBED_REF`, `_embed_ref_name`, `persist_canvases`). My first pass missed it (grepped only the file's first ~50 lines and a literal substring that did not match the regex) and concluded main chat could never originate a build. That was wrong in an important way: `persist_canvases` runs unconditionally at the end of **every** turn in **every** conversation (`jarvis/chat/turn_handler.py:106-127`, no origin check), and the `jarvis-persona` skill that produces a dashboard build (`skills/jarvis-dashboards/SKILL.md`) is discoverable by ordinary description match in any chat, independent of the builder-only `context.page === "dashboards"` prompt gate (`turn_handler.py:1617`, sent only by `frontend/src/api/dashboards.js:99`). Open persona PR #192 ships a main-chat first-cut build using this exact path. So the thumbnail and its click-through now gate on canvas presence (`isDashboardCanvas`, keyed on the hosted-embed `documents/` name prefix that only the dashboards skill emits) OR the pre-existing builder-origin rule, not origin alone. The live progress card stays origin-only: a main-chat build has no honest pre-canvas signal to key on, so it shows no card and the thumbnail simply appears when the canvas lands, rather than guessing a phase it never observed.
2. `dashboardOpenRoute` is reused unchanged for the thumbnail's click-through, exactly as assumed.
3. `DashboardCanvas.vue` wires the live query bridge (`callDashboardTool`/`runDashboardSource`) in both its modes, which `dashboardOpen.js` documents main chat deliberately never runs. Reusing it for an auto-rendered thumbnail would fire live queries per rendered message, so the thumbnail instead reuses the same static `cvOf`/`ensureCanvas` srcdoc main chat already fetches for the artifact preview panel, scaled down via CSS transform.

Piece A's phases come from the same live signals `ChatView.vue` already keeps for its generic activity line (`activeTools`, `statusPhase`, `waiting` from `tool:start`/`tool:end`), never a timer. A turn joined already in flight (no `activeTools` seen yet) reports no phase rather than guessing "Understanding." `activeTools` is never backfilled from history, confirmed at `ChatView.vue:5430-5435`.

Known limit, now narrower than the first pass: the live progress card still gates on conversation origin only, so a plain message typed into main chat's own composer on a builder-origin conversation also shows "Building dashboard..." even though that specific turn never triggers the dashboards skill. It resolves cleanly with no canvas once the turn ends.

</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (never merge on FAIL)
- [x] New/changed behavior has tests (`frontend/src/lib/dashboardBuildCard.test.js`, `dashboardOpen.test.js` updated)

### DIFF
--- a/frontend/src/lib/dashboardBuildCard.js
+++ b/frontend/src/lib/dashboardBuildCard.js
@@ -1,38 +1,90 @@
 // The "Building dashboard…" live card + finished thumbnail, for a dashboard
 // build watched from MAIN chat (issue #858).
 //
-// Two things this file decides, both pure and both fenced by
-// dashboardBuildCard.test.js so the wiring in ChatView.vue stays honest:
+// Corrected understanding (this file originally assumed main chat could
+// never originate a build at all — wrong, see below): three things this
+// file decides, all pure and all fenced by dashboardBuildCard.test.js so
+// the wiring in ChatView.vue stays honest:
 //
 //   1. Whether the turn CURRENTLY streaming into this conversation is a
-//      dashboard build, so the live progress card only ever shows for one.
+//      BUILDER-ORIGIN dashboard build, so the live progress card only ever
+//      shows for one.
 //   2. What phase that build is honestly in RIGHT NOW, from the same
 //      real events ChatView already keeps for its generic "Working on it…"
 //      indicator (tool:start/tool:end -> activeTools, statusPhase). Never a
 //      timer: a phase this mount has not seen evidence for is reported as
 //      `null` (indeterminate) rather than guessed.
+//   3. Whether a FINISHED canvas item is a dashboard build regardless of
+//      which conversation it landed in — the signal the thumbnail and its
+//      click-through gate on, since origin alone misses a real path.
 //
-// The turn-level gate reuses the exact origin binding
+// The turn-level gate (1) reuses the exact origin binding
 // `canOpenInDashboards` (dashboardOpen.js) already uses for the finished
 // artifact's "Open in Dashboards" button: `origin_page` is stamped
 // server-side only when the message that opened this turn was sent with the
 // Dashboards builder's `context: {page: "dashboards"}` (jarvis/chat/api.py).
-// A turn earns the live card under precisely the same rule a build earns the
-// promote button — main chat itself never originates one (api/dashboards.js
-// is the only sender of that context), so both gates read the same fact:
-// "this conversation's current activity came from the builder".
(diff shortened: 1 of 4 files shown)


kavin-114:
Independent review pass: one finding, a theme toggle stranded the new inline thumbnail on its loading shimmer since cvOf's cache key carries the dark flag and nothing re-fetched it for a passive read. Fixed in 6cd6853d with a watch on effectiveDark scoped to canOpenDash messages, plus a regression test. Everything else verified clean: sandbox attributes match the existing artifact panel, the two v-if/v-else chains are exact complements (no double-render), reduced-motion disables the new shimmer, accessibility (single native button, decorative iframe excluded via aria-hidden/tabindex), no DashboardCanvas reuse, no dead code.

kavin-114:
Correction to finding 1 after cross-checking against jarvis-persona: main chat CAN originate a dashboard build. persist_canvases (jarvis/chat/canvas.py) runs for every conversation with no origin check, and the jarvis-dashboards skill is discoverable in ordinary chat independent of the builder-only prompt gate; open persona PR #192 ships a main-chat first-cut build on this exact path. Reworked in fb615483: the thumbnail and its click-through now gate on canvas presence (a hosted-embed name prefix only the dashboards skill emits) OR the existing builder-origin rule, so a main-chat build gets the same affordance once #192 ships. The live progress card stays origin-only, since a main-chat build has no pre-canvas signal to key on; documented in dashboardBuildCard.js and the PR body.

PR #880, opened 2026-08-15
fix(chat): stop persisting the agent runtime's own web UI as a dashboard canvas

kavin-114:
## Summary

Main-chat dashboard builds were rendering a dark "Control UI did not start" error page branded with the agent runtime's own product name instead of the built dashboard. Live repro on e2e.localhost (conversation `3hepjr1j4s`): the runtime published a real document with genuine dashboard content, but the gateway route that is supposed to serve it did not intercept the request on this build, so a GET for that path (and a fabricated, nonexistent one) both came back HTTP 200 with the runtime's own web-UI shell. `fetch_canvas()` only checked for a 200 with a body, so it saved and rendered that shell as the dashboard.

`persist_canvases` now probes the same route with a name that cannot exist before fetching any real artifact this turn. A healthy gateway 404s it; one that fell through to its own UI answers with something else, which is exactly what the live repro showed. When that happens, persistence is skipped for the whole turn (logged once) so a fetch failure degrades to no-canvas instead of a branded error page.

Known limit: main-chat dashboard thumbnails will not render until the runtime exposes a backend-fetchable document path for this gateway build. The Dashboards builder page is unaffected: it renders via its own live query bridge, not this persisted-canvas path.

## Pre-merge checklist

- [x] **CI is green** - the `tests` check on this PR passes (never merge on FAIL)
- [x] New/changed behavior has tests (`jarvis/tests/test_canvas.py`, `jarvis/tests/test_canvas_embeds.py`)

### DIFF
--- a/jarvis/chat/canvas.py
+++ b/jarvis/chat/canvas.py
@@ -67,6 +67,13 @@ _CANVAS_BARE = re.compile(rf"\S*?canvas/{_PATH}\.(?:{_EXTS})(?![\w])", re.IGNORE
 # ``[embed url="/__openclaw__/canvas/..." /]``). The gateway serves ref
 # documents at ``canvas/documents/<id>/index.html``, so refs fold into the
 # same fetch/persist path as plain canvas file references.
+#
+# On at least one observed runtime build, that route did not intercept these
+# requests at all (nonexistent AND real refs both came back HTTP 200 from the
+# runtime's own web UI, falling through as a catch-all) even though the
+# published document existed on disk with the right content. See
+# ``_gateway_fakes_missing_canvas_as_ok`` — persistence is skipped for the
+# whole turn when that is detected, rather than saving that page.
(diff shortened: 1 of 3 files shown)


kavin-114:
Live e2e on `e2e.localhost` (2026-08-15, integration branch of #880+#881) PASSED:

- Real main-chat dashboard build: runtime published `documents/cv_6f68.../index.html`, probe detected the fallback (fabricated path answered non-404), persistence skipped, exactly one `chat canvas: gateway fallback detected` Error Log, `canvas` field empty, `[embed]` marker stripped, reply rendered as clean prose. No branded page anywhere.

One correction to this description: the claim that the Dashboards builder is unaffected because it "renders via its own live query bridge" is not accurate. The builder canvas is fed by the `kind:"canvas"` frame published in `jarvis/chat/turn_handler.py:111-127` only when `persist_canvases` returns items, so on this gateway build the builder canvas now stays empty after a build (it previously rendered the branded shell, so this is still the better failure). Tracked in #884.

PR #881, opened 2026-08-15
fix(dashboards): stop double-dispatching turns and ask answers

kavin-114:
Repeat Enter/click in the Dashboards builder chat could fire duplicate turns and duplicate ask answers; both now stay locked out until the current turn actually finishes.

Live repro: three 1-char messages sent 18:57:00-18:57:02 each spawned a turn and wedged the pump. Root cause: `DashboardChatPane.vue`'s `send()`/`sendText()` only gated on `sending`, which clears as soon as the POST resolves, not when the turn (`runActive`) actually ends - so a fast repeat Enter/click landed in the window between "POST done" and "turn over." Both dispatch entry points and the Send button now also gate on `runActive`.

Separately, `AskCard.vue`'s Submit button never locked after a click; single-submit was left entirely to the host, and the dashboards host's guard only covered the POST window, so a second click could dispatch a duplicate answer while the turn's tail was still running. `AskCard` now locks itself optimistically (`answered`) right after a successful dispatch, and takes a `busy` prop so it refuses to lock when the host would silently swallow the emit instead - this closes a race the `runActive` fix above widens (the busy window is now longer). AskCard is the shared renderer for both ChatView and the Dashboards pane, so both surfaces get the fix.

Note: the existing test "the picks survive Submit, so a host that refuses the text loses nothing" previously asserted a second click re-emits the same answers - that was the owner-reported bug. Its retry-on-host-refusal intent is now covered by the `busy` prop (the card refuses to lock while the host would swallow the emit) rather than by allowing re-clicks; the test is updated accordingly, plus new tests for the lock and the busy-gated refusal.

## Pre-merge checklist

- [ ] CI is green - the `tests` check on this PR passes (never merge on ❌)
- [x] Branch is up to date with `develop` (branched from `upstream/develop` just before this PR)

### DIFF
--- a/frontend/src/components/chat/AskCard.vue
+++ b/frontend/src/components/chat/AskCard.vue
@@ -12,6 +12,7 @@
 					:key="li"
 					class="jv-ask-opt"
 					:class="{ on: isPicked(qi, lbl) }"
+					:disabled="answered"
 					@click="toggleSingle(qi, lbl)"
 				>
 					<span v-if="isPicked(qi, lbl)" class="jv-ask-tick">✓</span>{{ lbl }}
@@ -24,6 +25,7 @@
 					:key="oi"
 					class="jv-ask-opt"
 					:class="{ on: isPicked(qi, opt) }"
+					:disabled="answered"
 					@click="q.type === 'multi' ? toggleMulti(qi, opt) : toggleSingle(qi, opt)"
 				>
 					<span v-if="isPicked(qi, opt)" class="jv-ask-tick">✓</span>{{ opt }}
@@ -35,6 +37,7 @@
 				type="date"
 				class="jv-ask-field"
 				:value="sel[qi] || ''"
+				:disabled="answered"
 				@input="pickSingle(qi, $event.target.value)"
 			/>
 			<input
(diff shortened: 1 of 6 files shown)


kavin-114:
Self-review pass (quality-code-review checklist, correctness/concurrency focus) found one real gap, now fixed in the latest commit: the ask card's link-search suggestion dropdown had no `answered` gate of its own — it only closed via the input's `@blur`, which is timing-fragile against a click that lands before the blur fires. A click racing that timing could still pick into an already-answered card. `v-if` on the dropdown now requires `!answered` directly, with a mount-level test covering it.

No other correctness, security, or concurrency issues found. Everything else (dispatch guards, the `busy` prop race, test coverage) matches the design discussed in the PR description.

kavin-114:
Live e2e on `e2e.localhost` (2026-08-15, integration branch of #880+#881) PASSED:

- Dashboards builder: three Enter presses within 49 ms produced exactly one `send_message` POST and one user row; the discriminating check is that Send stayed disabled well after the POST resolved, for the whole run (`runActive` window), and re-enabled once the turn ended, including after an errored turn.
- AskCard (dashboards host): three rapid Submit clicks produced exactly one answer dispatch; the card locked immediately after the first successful dispatch. Server showed exactly 2 user messages (prompt + one answer).
- ChatView host: could not coax the model into emitting an ask card live, so that surface is covered by the unit tests only (shared renderer + `busy` prop).

Issue #884, opened 2026-08-15
Dashboards must use the integrated dashboard feature, never the openclaw canvas


kavin-114:
Dashboards must run entirely on the integrated dashboard feature. The openclaw canvas channel (agent publishes HTML, gateway serves it at `/__openclaw__/canvas/...`, backend fetches and persists it) must not be part of the dashboard flow at all. Today it is the only delivery path for built dashboards, and on the current gateway build that path is dead, so builds finish with nothing visible anywhere.

**Target behavior (owner-defined):**
1. A dashboard request in main chat routes the user to the Dashboards page and builds there. No in-chat first cut.
2. The build progress animation (phase ticks from real skill/tool events, already implemented in `dashboardBuildCard.js`) runs whenever the dashboards skill or its tool calls are active.
3. The built dashboard is delivered by a tool call into the Jarvis-owned store (`save_dashboard` / `Jarvis Dashboard` doctype, which already holds full html + query sources and renders with no gateway involved). No hosted-embed markers, no gateway fetch.

**Folds in:** #858 (intent routing to the page, this issue adopts its proposal) and #877 (progress card; solved by building on the page where the card already works). Touches jarvis-persona `skills/jarvis-dashboards` (stop publishing hosted embeds), jarvis backend (`jarvis/chat/canvas.py` persist path no longer used for dashboards), and the SPA (intent route).

<details><summary>Evidence that the canvas path is dead (2026-08-15 live run, e2e.localhost)</summary>

- Agent built the dashboard: `/home/node/.openclaw/canvas/documents/cv_6f683877caf3495088c3f8ea457958be/index.html` exists in `jarvis-pool-68b37b`, title "Total Items", real content.
- Gateway GET for that exact path returns 200 with the runtime's own web UI shell. A fabricated path returns the identical page. The canvas route is not intercepting; requests fall through to the runtime UI as a catch-all.
- Backend behavior (post #880, working as designed): one `chat canvas: gateway fallback detected` Error Log per build turn, persistence skipped, embed marker stripped.
- Visible result: main chat shows prose with no thumbnail; Dashboards builder canvas stays on "Nothing on the canvas yet" after a successful build.
- The openclaw image is unpinned (`ghcr.io/openclaw/openclaw:latest`); this arrived with an image roll. Under this issue's direction, the gateway route no longer matters for dashboards; pinning the image is a separate infra concern.

</details>

kavin-114:
[...]

PR #886, opened 2026-08-15
feat(dashboards): redirect main chat to the builder, deliver builds via save_dashboard (#884)

kavin-114:
## Summary

Implements #884: dashboard builds never touch the agent-runtime canvas again. Main chat now redirects a dashboard ask to the Dashboards builder (new `jarvis-goto` fence, seeded prompt, one-shot auto-navigate), the builder shows the phase-tick build animation from live tool events, and the built HTML is delivered by a new `save_dashboard` registry tool that saves a real Jarvis Dashboard and publishes a `dashboard` frame the builder renders via its saved-row path.

Merge order: this PR first, then jarvis-openclaw-plugin (tool descriptor), then jarvis-persona (skill rewrite). Old canvas machinery stays for old transcripts; removing it is a later cleanup.

Live e2e (e2e.localhost, integration of all three branches): redirect fired once, interview asked only the undecided question, phase ticks ran, `Jarvis Dashboard` row saved Connected with `source_conversation` stamped, revision updated in place and re-rendered live, zero canvas events, zero fallback Error Logs, zero `[embed]` markers. Known follow-up: the very first event-driven adoption can render blank until reload (revision re-renders fine).

## Pre-merge checklist

- [x] Branch is **up to date with `develop`**
- [x] New/changed behavior has tests (`test_save_dashboard_tool.py`, `chatGoto.test.js`, `dashboardPrefill.spec.js`, `DashboardChatPane.spec.js`, extended block/build-card suites)

### DIFF
--- a/frontend/src/composables/dashboardPrefill.js
+++ b/frontend/src/composables/dashboardPrefill.js
@@ -1,0 +1,24 @@
+import { ref } from "vue";
+
+// Cross-view hand-off for a ```jarvis-goto redirect (main chat -> Dashboards
+// builder, issue #884). Main chat no longer builds dashboards itself: the
+// agent restates the request and points here with a fenced block, ChatView
+// stashes it and router.push("/dashboards"); DashboardsPage consumes it ONCE
+// on mount (and clears it on EVERY mount so a stale prompt can never fire
+// later): when `autoSend` is set it hands the text to the chat pane's
+// sendText(), which sends it as the next message on the builder's thread.
+// Shape:
+//   { text: string, autoSend: true }
+export const pendingDashboardPrefill = ref(null);
+
+export function setDashboardPrefill(payload) {
+	pendingDashboardPrefill.value = payload || null;
+}
+
+// Read-and-clear: returns the pending payload (or null) and empties the slot
+// so a later plain visit to Dashboards doesn't re-fill a stale prompt.
+export function takeDashboardPrefill() {
+	const v = pendingDashboardPrefill.value;
+	pendingDashboardPrefill.value = null;
+	return v;
+}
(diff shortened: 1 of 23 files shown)


kavin-114:
Automatic high-effort review ran on this PR; outcomes (commit f2698df9):

- Fixed: `save_dashboard` added to `_WRITE_TOOLS` (agent saves are now audited + savepointed like every sibling write tool; still ungated, same class as `save_agent_dashboard`).
- Fixed: `onDashboardSaved` now adopts a same-name revision directly but routes a DIFFERENT incoming row through `confirmDiscard`, restoring the "every path that drops an unsaved canvas asks first" invariant (confirmDiscard no-ops when nothing is at stake, so the normal build flow is unchanged).
- Fixed: id-less `tool:end` frames now fall back to the latest running entry with the same tool name, so the phase ticks cannot freeze on a tool that already finished.
- Fixed: removed the dead `in_test` commit guard and its misleading comment in the tool (the shared save path commits unconditionally; tests clean up by hand and say so).
- Fixed: `jarvis:goto-fired` stamps now carry timestamps and are pruned to the newest 50.
- Skipped deliberately: factoring the session/html/error guards shared with `save_agent_dashboard` into a helper; that touches the shipped delegate tool and belongs in its own cleanup, not this PR.

Still open (follow-up, also in the PR body): the very first event-driven adoption in the builder can render blank until reload; revisions re-render correctly.

kavin-114:
Added per owner direction (commits 315514b2, 9e3ab661, cf2ed55d): (1) pre-redirect morph line in main chat - the working line morphs into sliding chevrons + "Taking you to the Dashboards builder" once the goto block lands, with a ~900ms hold before navigation so the animation has real screen time (skipped under prefers-reduced-motion); (2) the #874 dashboard-only live card is generalized into a common artifact activity card (PDF / spreadsheet / image / dashboard kinds, adaptive icon+title, same four phase ticks from real tool events; run_report alone deliberately never triggers it). Pure CSS transform/opacity animation, zero layout shift, theme tokens only, latch+timer hygiene on every terminal. Verified live: both animations captured on e2e.localhost.

PR #906, opened 2026-08-17
fix(list): make the Sort field picker actually selectable (+ Created datetime sort)

Navin-S-R:
## Problem
On every list (Skills, Agents, Files, Macros, Dashboards, Support, Triggers) the **Sort** control only ever let you sort by the page default (e.g. `Name`). The "Sort by" field picker was a frappe-ui `FormControl type="select"` **inside a `Popover`**, and frappe-ui `Select` renders its options through a reka `SelectPortal` — whose click counts as an *outside-click* and **dismisses the popover before a field can be picked**. So the datetime columns (Updated, etc.) were wired end-to-end but unreachable. (`SortButton`'s own comment already flagged this for the direction toggle, which they'd moved out; the field select was left trapped.)

Reported on Skills: *"Unable to use column other than Name for sorting; it has to have date/datetime options; the UX is not good."*

## Fix
- **`SortButton.vue`**: replace the portal-based select with a **portal-free `role="menu"` list of `<button role="menuitemradio">` rows** rendered in the popover's own DOM. A click now stays inside the popover and applies via `pickField`. The public contract (props `sortOptions`/`sort`/`defaultSort`, emit `update:sort {field,dir}`), the always-visible direction toggle, and Clear Sort are unchanged. Shared component → **fixes sorting on every list page**. Adds an `aria-labelledby` group label.
- **Skills**: add a **`Created` (creation) datetime** sort option + whitelist `creation` in the backend `_SKILLS_SORTABLE`. `order_by` stays whitelist-only (no injection; `creation` is a real column, valid in `ORDER BY` without being in the `SELECT`).

## Tests
- `SortButton.spec.js`: renders every option as a menuitemradio, picks each (incl. Updated/Created) → emits the right field; direction toggle flips dir; Clear Sort resets. Frontend suite: **1337 green**.
- `test_list_filters_pilots.py`: `creation`/`modified` are honored sortable columns, an unknown field falls back to Name, and an **end-to-end** case creating 3 skills with distinct creation asserts real row-order for `sort_field=creation` (asc/desc). Backend module: 24 green. Mutation-verified.

## Verification note
This is an interaction fix; jsdom can't exercise the reka-portal outside-click, so the click-verification (open Sort → pick Updated/Created → list re-sorts, popover stays open) is a manual browser check. Follow-up: the **Filter** panel has the same reka-select-in-popover pattern (separate change).

### DIFF
--- a/frontend/src/components/list/SortButton.spec.js
+++ b/frontend/src/components/list/SortButton.spec.js
@@ -1,0 +1,98 @@
+import { describe, it, expect, vi } from "vitest";
+import { mount } from "@vue/test-utils";
+
+// frappe-ui's ESM entry doesn't resolve under vitest; stub the three components
+// SortButton uses. The Popover stub renders BOTH slots so the picker body (which
+// used to be trapped behind a portal-select) is queryable in jsdom.
+vi.mock("frappe-ui", () => ({
+	Popover: {
+		name: "Popover",
+		template: `<div>
+			<slot name="target" :togglePopover="() => {}" />
+			<slot name="body" :close="() => {}" />
+		</div>`,
+	},
+	Button: {
+		name: "Button",
+		props: ["label", "icon", "tooltip", "variant"],
+		emits: ["click"],
+		template: `<button :data-icon="icon" @click="$emit('click')">{{ label }}</button>`,
+	},
+	FeatherIcon: {
+		name: "FeatherIcon",
+		props: ["name"],
+		template: `<span class="feather" :data-name="name" />`,
+	},
+}));
+
+import SortButton from "./SortButton.vue";
+
+const OPTIONS = [
+	{ label: "Name", value: "skill_name" },
+	{ label: "Updated", value: "modified" },
+	{ label: "Created", value: "creation" },
+];
+const DEFAULT = { field: "skill_name", dir: "asc" };
+
+function mountBtn(sort = { field: "", dir: "" }) {
+	return mount(SortButton, {
+		props: { sortOptions: OPTIONS, sort, defaultSort: DEFAULT },
+	});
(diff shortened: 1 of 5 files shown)


tech-specsmakers:
## Don't merge yet: this fixes Sort, but two of the reported issues remain

The fix itself looks right and is verified against an independent diagnosis: the root cause (a reka `SelectPortal` click counts as an outside click and dismisses the frappe-ui `Popover`) is exactly what we traced on latest develop, the portal-free in-popover menu is the correct pattern, and since `SortButton.vue` is shared it repairs sorting on every list page. The label-jump fix and the `creation` whitelist + tests also look good. Nice work.

Please hold the merge, because the original report covers two more problems that this PR leaves open (the PR notes the first one itself):

1. **The Filter panel has the same portal bug.** The field, "equals", and "Yes" dropdowns never open because the popover dismisses itself the moment the select renders. `FilterGroup.vue` and `FilterValueControl.vue` need the same portal-free treatment this PR gave `SortButton.vue`, on every page that uses them.
2. **Thick black focus border** on the filter value controls (see the "Yes" pill in the screenshot). That heavy ring is not the Frappe UI design language and shows up wherever these controls render.
Suggested path: stack the Filter panel + focus-ring fix on top of this branch (or extend this PR), verify all three issues in one browser pass on the Skills page, then merge together so develop never ships a half-fixed filter bar.

Reference screenshot from the report (Skills page, latest develop; filter dropdowns stuck at their defaults, black ring on "Yes"):

pasted-1

Navin-S-R:
Superseded by #914, which combines this Sort fix with the Filter/dropdown portal-free fix on top of current develop.

PR #936, opened 2026-08-19
fix(v15): route set_delimiters_flag through compat probe (unbreak CI)

kavin-114:
## Summary

Fixes the version-15 branch's CI, which went red for every PR after the shared `jarvis-ci-base:v16` image was rebuilt today with a Frappe that no longer exposes `DataImport.set_delimiters_flag`. `jarvis/tools/_import_preview.py` called that method directly and raised `AttributeError`, failing ~66 import/preview tests and destabilising the whole run.

`develop` already routes this call through a probe-based shim (`jarvis.compat.set_delimiters_flag`) and stays green on the exact same image; the shim was simply never backported to version-15. This adds it verbatim from develop and swaps the two call sites.

## What changed
- `jarvis/compat.py`: add `set_delimiters_flag(doc)` (verbatim from develop) - `getattr`-probes the method and no-ops when the framework lacks it
- `jarvis/tools/_import_preview.py`: call `compat.set_delimiters_flag(doc)` instead of `doc.set_delimiters_flag()` at both sites; add the `compat` import
- `jarvis/tests/test_compat_frappe_versions.py`: unit-test both probe branches (calls through when present, no-ops when absent)

## Risk and verification
- Behavior-preserving on a Frappe that has the method (the shim calls straight through); the only change is that a framework lacking it no longer raises
- Matches develop, which is green on the same `:v16` image (run for `develop@2a408487` passed at 14:34 today while base `version-15` failed identically)

<details><summary>Root cause evidence</summary>

- Base `version-15` (run 32260184316) fails with `AttributeError: 'DataImport' object has no attribute 'set_delimiters_flag'` x66, plus cascade `db down` / `no such site` and 3 `orjson` errors (orjson is present in the image - develop imports it and passes - so those clear once the run stops falling over).
- Timeline: this branch was green at 13:21 today, red from ~13:48 - the window the base image was rebuilt. No version-15 code changed in between.
- develop's `jarvis/compat.py` docstring documents the same class of Frappe 15/16 drift and the probe-not-version-string doctrine.

</details>

## Pre-merge checklist

- [ ] **CI is green** - the `tests` check on this PR passes (this PR is the fix for it)
- [ ] Branch is **up to date with `main`** (n/a: targets `version-15`)

https://claude.ai/code/session_01WZJ3nhp1CqHuWzhGibypRa

### DIFF
--- a/jarvis/compat.py
+++ b/jarvis/compat.py
@@ -29,6 +29,19 @@ import inspect
 import frappe
 
 
+def set_delimiters_flag(data_import_doc) -> None:
+	"""Apply a Data Import's custom CSV delimiter, where the framework supports it.
+
+	Frappe 16 added ``DataImport.set_delimiters_flag()`` (and the custom-delimiter
+	feature it drives); Frappe 15 has neither and always parses with a comma.
+	Calling it unconditionally was an ``AttributeError`` on 15, so skip it there:
+	with no delimiter feature, the default comma parse is already correct.
+	"""
+	fn = getattr(data_import_doc, "set_delimiters_flag", None)
+	if fn is not None:
+		fn()
+
+
 @functools.cache
 def _get_content_takes_encodings(cls) -> bool:
 	"""Does this File class accept ``get_content(encodings=...)``? (Frappe 16)"""
--- a/jarvis/tests/test_compat_frappe_versions.py
+++ b/jarvis/tests/test_compat_frappe_versions.py
@@ -284,3 +284,35 @@ class TestItemisedTaxAcrossMajors(FrappeTestCase):
 			self.assertIs(seen["target"], doc)
 		else:
 			self.assertEqual(list(seen["target"]), list(doc.get("taxes")))
+
+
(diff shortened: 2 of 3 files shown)


kavin-114:
Closing as redundant. This backported the compat.set_delimiters_flag shim, but batch #932 (merged 15:09) already landed the identical change on version-15 base: jarvis/compat.py has the shim and jarvis/tools/_import_preview.py already calls compat.set_delimiters_flag(doc) at both sites. I'd branched off 88fc5f6e (pre-#932) and diagnosed the symptom without realizing the batched Frappe-15 compat backport (batches 1-2 in #932, batch 3 in #937) already owns this. The remaining failures on this branch's run (pypika .run, orjson test-infra) are exactly what #937 covers.

One salvageable bit: this PR added a unit test TestSetDelimitersFlag (probe-calls-through / no-ops-when-absent) in jarvis/tests/test_compat_frappe_versions.py. If you want that coverage, happy to fold it into #937 or a follow-up; otherwise it goes away with this branch.

PR #951, opened 2026-08-20
DO NOT MERGE - debug(v15): probe issue #5 (cross-install detach)

kavin-114:
Diagnostic-only draft to capture one v15 CI round. Adds probe output to the failing assert message in test_cross_install_recurrence_bump_is_detached_not_deleted so we can see whether f_b.first_seen_run survived in the DB (detach misread it) or was nulled. Close after reading the tests-shard log; never merge.


kavin-114:
Probe served its purpose: v15 get_all omits first_seen_run from the projection (raw DB row had it). Root cause fixed in #953. Closing.

PR #958, opened 2026-08-20
Hold a calm "confirming" screen after checkout instead of the panic card

kavin-114:
## Summary

After a customer completes checkout, the "Confirming your payment" spinner currently flips **straight to a recovery card that reads like a failure** ("We have not confirmed this payment. Check the status before doing anything else." + Check / Continue / Contact support) whenever the post-return reconcile comes back still-pending. In production that pending answer is the normal case: Razorpay's confirmation webhook lags the browser redirect by a few seconds. So a customer who paid successfully is shown a panic screen during a routine wait.

This keeps a calm "Confirming your payment, we're checking automatically" hold on screen while the **already-existing** background auto-poll (~2 min) confirms, and only escalates to the recovery card once that poll actually gives up (`pendingPollStuck`) or a real failure code arrives. The hold is gated strictly on the completed-checkout return (`?pay=done`), so a failed or abandoned return still gets the recovery card immediately, and the copy says "confirming", never "confirmed" — money truth stays server-owned. The auto-poll's first re-check is also moved early (4s) so the common few-seconds webhook lag resolves while the customer is still watching the calm screen.

No backend or state-machine change; this is view-level gating over machinery that already exists.

### Visual — before vs after (same normal webhook-lag moment)

| Before (shown instantly) | After (calm hold) |
|---|---|
| ⚠️ "We have not confirmed this payment." + Check / Continue / Support | "Confirming your payment… we're checking automatically" spinner |

_Before/after preview image attached below._

<details>
<summary>Root cause & mechanism</summary>

- On return from the admin-hosted checkout, the pay page appends `?pay=done\|failed\|pending`. `OnboardingView.vue`'s `onMounted` shows the one-shot "Confirming your payment" spinner (`confirmingReturn`) and runs a **single** reconcile (`get_onboarding_state` + a fresh-mount `check_signup_payment_status` heal) — never a poll.
- The moment that single round trip resolves, `confirmingReturn` flips false and whatever state the machine landed in renders. A completed checkout whose webhook has not landed yet lands on `PAYMENT_CONFIRMATION_PENDING` → `S.UNKNOWN` → `showRecovery` → the alarming card, immediately.
- A background auto-poll (`runPendingAutoPoll`, `PENDING_CHECK_ATTEMPTS` × `PENDING_CHECK_INTERVAL_MS` ≈ 2 min) was already running *underneath* that card. This change puts a calm screen in front of it instead of the panic card until it either confirms or exhausts.

**Why gated on `?pay=done` and not `S.UNKNOWN` alone:** `S.UNKNOWN` also covers a checkout the customer never touched (see the `PAYMENT_CONFIRMATION_PENDING` comment in `paymentCodes.js`). Only the `done` return proves completion, so only it softens the copy.

**Scope:** deliberately not extended to `PAYMENT_AUTHORIZED_PENDING_CONFIRM` (`S.CONFIRM_REQUIRED`) — an e-NACH mandate can pend indefinitely with no auto-poll behind it, so a calm spinner there would hang forever; its own copy is already non-alarming.

**Known edge (acceptable for v1):** a hard page reload during the wait strips `?pay=done`, so a reload drops to the recovery card. That is honest, not a regression; persisting the return across reloads (sessionStorage) is deliberately out of scope.
</details>

## Test evidence

- `frontend/src/views/OnboardingView.spec.js`: +4 tests — calm hold shows on a completed return, `?pay=failed` still gets recovery, escalation to recovery once the poll gives up, and the hold releases the moment the payment confirms.
- Full frontend suite green on Node 24: **1310 passed / 84 files**. The existing auto-poll timing tests were verified unaffected by the 4s early check.
- `prettier` (pinned 2.7.1 via pre-commit) + `eslint` pass.

## Pre-merge checklist

- [ ] Branch is **up to date with `develop`**
- [x] New/changed behavior has tests (4 new view specs)

https://claude.ai/code/session_01Ui9kZy3ZCgzkwz2jRb5uci

### DIFF
--- a/frontend/src/views/OnboardingView.spec.js
+++ b/frontend/src/views/OnboardingView.spec.js
@@ -1430,4 +1430,40 @@ describe("post-checkout settling hold (calm wait, not the panic card)", () => {
 		expect(wrapper.vm.pay.value).not.toBe(STATES.UNKNOWN);
 		expect(wrapper.vm.showPaymentSettling).toBe(false);
 	});
+
+	// Finding 3: the early 4s first re-check is scoped to the completed-checkout
+	// case (a webhook expected within seconds). An in-app pending wait that never
+	// left the page keeps the ordinary 15s first interval and does not hit the
+	// rate-limited endpoint any sooner than before.
+	it("re-checks early (~4s) on the first auto-poll tick after a completed checkout", async () => {
+		pendingAfterDoneReturn();
+		vi.useFakeTimers();
+		mountView();
+		await flushPromises();
+		// The mount-time return-heal already fired one check; clear it so we are
+		// measuring the auto-poll's own first tick.
+		api.onboardingPaymentApi.checkSignupPaymentStatus.mockClear();
+		await vi.advanceTimersByTimeAsync(4_000);
+		expect(api.onboardingPaymentApi.checkSignupPaymentStatus).toHaveBeenCalledTimes(1);
+	});
+
+	it("keeps the ordinary 15s first interval for an in-app pending wait (no completed checkout)", async () => {
+		// No ?pay= return, so returnedFromCompletedCheckout stays false.
+		api.onboardingPaymentApi.getOnboardingState.mockResolvedValue(
+			ENVELOPE({ code: "PAYMENT_CONFIRMATION_PENDING", attempt_id: "att_s", generation: 1 })
+		);
+		api.onboardingPaymentApi.checkSignupPaymentStatus.mockResolvedValue(
+			ENVELOPE({ code: "PAYMENT_CONFIRMATION_PENDING", attempt_id: "att_s", generation: 1 })
+		);
+		vi.useFakeTimers();
+		mountView();
+		await flushPromises();
+		expect(api.onboardingPaymentApi.checkSignupPaymentStatus).not.toHaveBeenCalled();
+		// 4s must NOT be enough for the first tick here (unlike the checkout-return case).
+		await vi.advanceTimersByTimeAsync(4_000);
+		expect(api.onboardingPaymentApi.checkSignupPaymentStatus).not.toHaveBeenCalled();
+		// The ordinary 15s interval fires it.
+		await vi.advanceTimersByTimeAsync(11_000);
+		expect(api.onboardingPaymentApi.checkSignupPaymentStatus).toHaveBeenCalledTimes(1);
+	});
 });
--- a/frontend/src/views/OnboardingView.spec.js
+++ b/frontend/src/views/OnboardingView.spec.js
@@ -1324,3 +1324,110 @@ describe("pending-payment auto-poll", () => {
 		expect(api.onboardingPaymentApi.checkSignupPaymentStatus).not.toHaveBeenCalled();
 	});
 });
+
+// The panic-page fix: a customer who COMPLETED checkout (?pay=done) and lands on
+// the coded PAYMENT_CONFIRMATION_PENDING wait (UNKNOWN) - because Razorpay's
+// confirmation webhook lags the browser redirect by seconds - must see the calm
+// "Confirming your payment / we're checking automatically" settling hold, NOT the
+// alarming "We have not confirmed this payment / Check the status before doing
+// anything else" recovery card, while the auto-poll works. The recovery card is
+// reached only once the poll gives up (pendingPollStuck). The gate is the completed
+// return, never UNKNOWN alone, so a failed/abandoned return still gets recovery.
+const SETTLING_TEXT = "You've completed checkout";
+const RECOVERY_PENDING_TEXT = "We have not confirmed this payment";
+describe("post-checkout settling hold (calm wait, not the panic card)", () => {
+	function atSearch(search) {
+		window.history.pushState(null, "", "/onboarding" + search);
+	}
+	afterEach(() => {
+		window.history.pushState(null, "", "/onboarding");
+		vi.useRealTimers();
+	});
+
+	// A completed checkout whose webhook has not landed yet: both the passive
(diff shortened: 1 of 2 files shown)


kavin-114:
## Code review addressed + verification

**Review findings (all 4 confirmed) resolved:**

- **Findings 0+2 (bfcache/tab-focus return never set the completion flag)** and **finding 1 (stale flag)** are fixed by one change: the completion proof (`returnedFromCompletedCheckout`) is now **reset whenever a checkout (re)opens** (`CHECKOUT_OPEN` watch). Only a fresh `?pay=done` full-navigation return re-arms the calm hold. The signal-less return paths (bfcache restore / tab-focus, which carry no `?pay=` outcome) therefore stay on the honest recovery card rather than being told "you've completed checkout" without proof, and proof from one attempt can never leak into a later, unrelated wait. This keeps the advisor's safety invariant: never soften the copy without a real completion signal.
- **Finding 3 (early 4s re-check applied to every pending wait):** scoped to the completed-checkout case only. In-app pending waits keep the ordinary 15s first interval, so they don't hit the rate-limited endpoint any sooner than before. +2 tests pin both directions.
- **Finding 4 (settling/confirming template duplication):** deliberately left as two parallel blocks. They carry different copy, different roles, and the file's convention is explicit per-state blocks; extracting would also churn the already-shipped confirming block. Noted, not changed.

**Verification:**
- Full frontend suite green on Node 24: **1312 passed / 84 files** (70 in `OnboardingView.spec.js`, incl. 6 for the settling hold + scoping).
- Live browser render of the real onboarding SPA (this branch's build) in the webhook-lag state: the calm "Confirming your payment / we're checking automatically" screen shows in place of the panic card. Escalation-after-ceiling is covered by the deterministic `pendingPollStuck` test (the live 2-min escalation shot was not captured due to a static-serve harness artifact — the un-rendered jinja boot block — not a product issue).

Visual evidence (before/after + live calm render) shared with the author.

kavin-114:
## Manual testing found the primary case — extended the fix

A live onboarding run surfaced that the standard trial / auto-pay signup completes the ₹0 authorization and lands on **`PAYMENT_AUTHORIZED_PENDING_CONFIRM`** (bank confirming the e-NACH mandate) — **not** the `PAYMENT_CONFIRMATION_PENDING` webhook-lag state the first two commits covered. That mandate-confirm state still rendered the alarming recovery-card chrome ("Check payment status / Contact support / we checked just now, nothing has changed yet") right after a successful authorization. It's almost certainly the *primary* panic in production.

Extended the same bounded calm hold + background auto-poll to it, via one shared `isAutoPollable` predicate (loop guards, `restartPendingAutoPoll`, and the settling computed all read it, so they can't drift). Copy is now state-keyed: the genuinely-authorized state shows **"Your payment is authorized"** (server truth), the webhook-lag state keeps "Confirming your payment". After the ~2 min ceiling, both escalate to the existing recovery card unchanged — the right exit for an e-NACH that can pend by design. +3 mirror tests.

Verified: full suite **1315 pass / 84 files** on Node 24; prettier + eslint clean; live render of the new authorized-calm screen confirmed. (An earlier advisor note had excluded this state on the assumption its copy was "already calm" — the live screenshot disproved that at the rendering level, since the calm headline sat inside the recovery card's chrome.)

kavin-114:
## Live run found the real gate bug: the mandate returns `?pay=pending`, not `?pay=done`

A from-scratch onboarding run still showed the recovery card after the auto-pay mandate authorization. Root cause: the admin pay page (`billing/checkout/shell.py` `outcomeFor`) returns **`?pay=pending`** — not `?pay=done` — for the "authorized, awaiting confirmation" case (`next_step` poll/contact_support), which is the `PAYMENT_AUTHORIZED_PENDING_CONFIRM` mandate path this hold exists for. The gate accepted only `done`, so it rejected the primary path.

Fix: accept `done` OR `pending` as proof the customer finished the Razorpay flow (both do; only the outcome differs). `failed` / bare reload still fall to the recovery card. The mandate tests now use the real `?pay=pending` return, and re-verified live against the built bundle: the calm "Your payment is authorized" screen now renders for the mandate path.

Note on process: the earlier live check passed only because the harness stubbed `?pay=done` — the wrong param. Fixed the harness and the tests to use the param the real pay page emits.

Suite still green: **1315 pass / 84 files**, prettier + eslint clean.

kavin-114:
## Live testing found a worse failure: the calm hold could hang forever

A live run left a customer on the calm "confirming" spinner for **~15 minutes**. Root cause: the calm hold's only escape was the auto-poll's `pendingPollStuck` (its 8-check ceiling), but that ceiling is **not reliably ~2 min** — the server rate-limits `check_signup_payment_status` (every 429 adds a cooldown the poll waits out) and browsers throttle background-tab timers, so "8 checks" can stretch far past 2 minutes. When it does, the calm screen — which has **no buttons** — traps the customer, which is worse than the panic card it replaced.

Fix: a hard **wall-clock deadline** (`SETTLING_MAX_MS = 120s`). The calm hold escalates to the recovery card (with its manual "Check payment status" button) after ~2 min **regardless of what the poll is doing**. Escalation is now `pendingPollStuck` OR the deadline, whichever comes first — so the hold can never trap the customer. A new test pins this with every check rate-limited (so `pendingPollStuck` can never fire).

Diagnosed via an instrumented build: the poll reliably stalls after ~2 checks under timer throttling / rate-limit cooldowns. Suite green: **1316 pass / 84 files**.

_Note: on the local bench the Razorpay confirm webhook can't reach `localhost`, so payments never actually confirm here and the hold always rides to the ~2 min escalation — expected locally; in prod the webhook confirms in seconds/minutes and the flow advances to Connect._

Navin-S-R (approved):

PR #975, opened 2026-08-24
fix(widget): lazy-mount the Desk chat panel so no-access users see no…

venkat102:
… popup

Combines #969 and #966. The floating chat FAB rendered <Panel> eagerly on every Desk page, so Panel's onMounted fired get_chat_ui_settings()/readiness on every full page load. For a System User without the Jarvis User role those endpoints 403 via require_jarvis_access(), and Frappe surfaced the throw as a red "You need the Jarvis User role" dialog on every Desk page.

Gate the Panel behind v-if="panelMounted" so it mounts only on the first open and then stays mounted (keep-mounted, so conversation/scroll/draft survive a close->reopen). onFabClick already redirects no-access users to /jarvis-no-access before opening, so a role-less user mounts nothing.

Two pure helpers, unit-tested with node:test (positive + negative):

  - fabAction()       routes the tap: no-access / full-SPA / toggle.
  - panelTogglePlan() drives the lazily mounted Panel. On the FIRST open the
                      Panel mounts CLOSED and reveals on the next tick so
                      Panel.vue's non-immediate watch(props.open) observes a
                      false->true transition and runs its first-open load()
                      (conversation restore), ensureRealtime() and focus.
                      Revealing in the same tick would mount the Panel
                      already-open, skip that watch, and open it blank (the
                      first send minting a NEW conversation). The reveal is
                      scheduled before the throw-prone readDeskContext() so a
                      context-read failure cannot strand the panel unopened.

Supersedes #969 (first-open correctness) and #966 (fabAction extraction + tests); both edit the same Widget.vue lines and should be closed.

## Pre-merge checklist

> ℹ️ These repos are private on the GitHub **Free** plan, so branch protection is
> **not enforced** — CI cannot hard-block a merge. Honoring this checklist is what keeps
> broken changes out of UAT. See [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [ ] Branch is **up to date with `main`** (so it is tested against the latest code)
- [ ] New/changed behavior has tests (the coverage gate still passes)
- [ ] I self-reviewed the diff

### DIFF
--- a/jarvis/public/js/jarvis_chat/widget/Widget.vue
+++ b/jarvis/public/js/jarvis_chat/widget/Widget.vue
@@ -54,7 +54,13 @@
 			:class="{ 'jvw-backdrop--show': leaving }"
 			aria-hidden="true"
 		></div>
+		<!-- Lazily mounted: the Panel does not exist until the first open (v-if on
+		     panelMounted), so its onMounted fetches (get_chat_ui_settings /
+		     readiness) never fire on plain Desk page loads - and never pop the
+		     "You need the Jarvis User role" dialog for a role-less System User.
+		     It STAYS mounted after that; see onFabClick for the mount->reveal. -->
 		<Panel
+			v-if="panelMounted"
 			ref="panelRef"
 			:open="panelOpen"
 			:context="effectiveContext"
@@ -70,13 +76,14 @@
 </template>
 
 <script setup>
-import { ref, computed, watch, onMounted, onBeforeUnmount } from "vue";
+import { ref, computed, watch, nextTick, onMounted, onBeforeUnmount } from "vue";
 import { FULL_CHAT_URL, conversationUrl, PANEL_MIN_VIEWPORT_PX } from "./config.mjs";
 import { contextFromRoute } from "./desk_context.mjs";
 import { panelLayout } from "./panel_anchor.mjs";
 import * as panelSize from "./panel_size.mjs";
 import { isDarkNow, watchTheme } from "./desk_theme.mjs";
 import * as fabPos from "./fab_position.mjs";
+import { fabAction, panelTogglePlan } from "./fab_action.mjs";
 import Panel from "./Panel.vue";
 
 // ---- FAB: draggable, edge-snapping, idle-fading launcher button.
@@ -159,6 +166,11 @@ const panelRef = ref(null);
 const isDark = ref(false);
 let unwatchTheme = null;
 const panelOpen = ref(false);
+// The Panel is v-if'd on this: it mounts on the FIRST open and then STAYS mounted
+// (keep-mounted, so the conversation/scroll/draft survive a close->reopen - a bare
+// v-if="panelOpen" would tear all that down on every close). Latches true on the
+// first open; onFabClick owns the two-phase mount->reveal.
+const panelMounted = ref(false);
 const deskContext = ref(null);
 const contextDismissed = ref(false);
 
@@ -255,7 +267,7 @@ function readDeskContext() {
 }
 
 function closePanel() {
-	panelOpen.value = false;
+	panelOpen.value = false; // hide only - panelMounted stays true (the latch)
 	fabEl.value?.focus();
 }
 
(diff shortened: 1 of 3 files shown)


kavin-114:
Reviewed — the fix is correct (verified the two-tick reveal against Panel.vue's non-immediate `watch(props.open)`, `readDeskContext` is pure client-side so no-access users hit no server call, and `fabAction` fails closed). PR tests 7/7 and all hosted CI green. 👍

One non-blocking nit: the code comments are quite long. The logic is subtle so *some* explanation is warranted, but a few blocks restate the same 'mount closed → reveal next tick → non-immediate watch → else blank panel' point three times over. Suggest trimming:

- **fab_action.mjs**: the module header (lines ~1–18) and the `panelTogglePlan` `deferReveal` docstring both explain the full mount/watch rationale in detail — collapse one into a one-line pointer to the other.
- **Widget.vue onFabClick**: the ~8-line `deferReveal` comment duplicates what `fab_action.mjs` already documents. A one-liner ('first open: mount closed, reveal next tick so Panel's watch fires — see fab_action.mjs') is enough here.
- The `panelMounted` ref comment and the template `v-if` comment also overlap; one can point to the other.

Net: keep one authoritative explanation (in fab_action.mjs, next to the pure logic) and make the call sites one-liners that reference it. Not a blocker.

kavin-114:
## Design/UX question: should the FAB render at all for users without Jarvis access?

The panel fix here is correct and I'm not asking to change it — killing the on-mount 403 popup is the right call. But it papers over a more basic question, so raising it here.

**Traced the actual behavior for a role-less System User:**

| State | What they get |
|---|---|
| Onboarding incomplete (no Company) → `jarvis_site_setup_complete === false` | Widget hidden for everyone (`host.style.display = none`). ✅ correct |
| Onboarding complete, user lacks `Jarvis User` role | FAB **renders**. Tap → `fabAction` returns `no-access` → full-page redirect to `/jarvis-no-access` → 'You need access to {agent}' + only a 'Back to Home' button. Dead end. |

Nothing in the bundle (`jarvis_widget.bundle.js`) or `widget_visibility.mjs` gates on `jarvis_has_access` — the FAB mounts for all authenticated desk users once the site is set up.

**The concern:** a System User who can never use Jarvis (say an accountant without the role) sees a floating button on every Desk page. Clicking it — the natural thing to do with a prominent FAB — abandons whatever form/report they're on and dumps them on a dead-end 'no access' screen with no way to request access. That's friction with no upside for them.

**The only upside of mounting-for-everyone** is the self-heal: `jarvis_no_access.py` re-checks `has_jarvis_access()` live and redirects to `/jarvis` if access was granted after boot. But that only helps the rare mid-session-grant case, and a page reload (which re-runs boot) achieves the same thing. We're trading a permanent papercut for all role-less users against a rare convenience.

**Suggestion — pick one, don't sit in the middle:**
- **(A) Hide the FAB for no-access users** — gate the widget on `frappe.boot.jarvis_has_access` (a `v-if` on `jvw-root`, or skip mount in the bundle). A user who can't use Jarvis sees nothing. The just-granted user's FAB appears on their next navigation/reload. This is my recommendation — cleanest, and it makes `fabAction`'s `no-access` branch a pure defensive fallback rather than an expected path.
- **(B) If discoverability for no-access users is an actual product goal**, then `/jarvis-no-access` needs a real 'Request access' action (notify an admin), not just 'Back to Home'. Otherwise the button is a tease.

Current state is the worst of both: visible button → dead-end page. Not a blocker for this PR (which fixes a real bug), but worth a follow-up decision before we call the no-access UX done.

kavin-114:
**Decision: go with option A.**

Hide the Jarvis button for users who don't have access. No point showing a button that only leads to a dead-end page.

We don't need to build anything for the "user opens /jarvis directly" case — the server already redirects role-less users to the no-access banner (and self-heals if they were just granted access). So that stays exactly as it is.

**Follow-up change (small):** gate the widget on `frappe.boot.jarvis_has_access` — a `v-if` on the widget root, or skip the mount in `jarvis_widget.bundle.js`. Keep `fabAction`'s `no-access` branch as a safety net in case the boot flag is ever stale.

Not blocking this PR — the panel fix here is good to merge on its own.

kavin-114 (approved):

PR #984, opened 2026-08-25
fix(wiki): collapse missing-vs-private into one error to close existence oracle (#733)

### LINKED ISSUE #733: wiki: read_wiki distinguishes "page missing" from "page private", leaking whether a colleague has a private page on a topic
`read_wiki` answers "that page does not exist" and "that page exists but is not yours" with two different, machine-distinguishable errors. Since User-scope slugs embed the target's email local part, that turns the tool into an existence oracle for other people's private pages.

## Where

`jarvis/tools/read_wiki.py:54-65`:

```python
name = frappe.db.get_value(WIKI, {"slug": slug}, "name")
if not name:
    raise InvalidArgumentError(f"unknown wiki page: {slug}")
if not frappe.has_permission(WIKI, ptype="read", doc=name):
    raise PermissionDeniedError(f"no read permission on {WIKI} {name}")
```

`_run_tool` puts the exception class name on the wire as the `code` field, deliberately, so callers can branch without parsing prose (`jarvis/api.py:1049-1054`). So the two cases differ in `code`, in `message`, and the permission message additionally discloses the internal docname.

## Why that is exploitable

The audience suffix is derived from the target user's email local part. `jarvis/jarvis/doctype/jarvis_wiki_page/jarvis_wiki_page.py:148`:

```python
return f"--u-{local}" if local else ""
```

and `:173`: "`--u-<localpart>` keys on the email LOCAL PART only".

So the slug for a colleague's private page is guessable: `<topic>--u-<their-localpart>`. Ask the agent to read it and the error tells you which of the two it was:

- `InvalidArgumentError` means no such page
- `PermissionDeniedError` means the page exists and you may not see it

Repeat across topics to map what a colleague keeps private notes about. You learn existence and topic, never content.

## The same shape on the write path

`update_wiki` behaves comparably. `_find_existing` (`jarvis/tools/update_wiki.py:58-74`) with a non-User scope returns any row matching the slug, and the caller then raises `PermissionDeniedError` at `:119` when the found page is non-Org and unwritable. A slug with no row instead falls through to the create branch. So probing distinguishes the two there too. Worth fixing in the same pass.

## Severity

Low. Content never leaks and this needs an authenticated desk user. It is an information-disclosure paper cut, not a boundary break. Filing it because the rest of this permission model is careful enough that this stands out, and because the fix is small.

## Suggested fix

Collapse both branches to one indistinguishable error, which is the standard treatment. Something like:

```python
name = frappe.db.get_value(WIKI, {"slug": slug}, "name")
doc = frappe.get_doc(WIKI, name) if name else None
if not doc or not wiki_permissions.can_read_page(doc, frappe.session.user):
    raise InvalidArgumentError(f"unknown wiki page: {slug}")
```

Two things to keep in mind:

- Drop the docname from the message. It is internal and it is the second disclosure here.
- The error text is read by the LLM, so it should still steer the model somewhere useful rather than inviting a retry loop. The `FeatureDisabledError` message added in PR #729 is a good model for that tone: say plainly that retrying or changing arguments will not help.

Keep whatever wording is chosen identical across both branches, including the `code`, or the oracle survives in a new form.

## Tests

Assert the two cases are **identical** on `code` and on `message`, not merely that each is an error. A test that only checks "raises something" passes without the fix.

Found while auditing the wiki permission model on 2026-08-08.

kavin-114:
## Summary

**Security / info-leak.** Closes an existence oracle in the wiki tools: `read_wiki` and `update_wiki` no longer let a caller tell "page does not exist" apart from "page exists but is another user's private page", which leaked whether a colleague keeps a private page on a topic.

<details><summary>Root cause / notes</summary>

Read path raised `InvalidArgumentError("unknown wiki page")` for a missing row but a distinct `PermissionDeniedError` (disclosing the docname) for an existing-but-unreadable row. Both now raise the byte-identical "unknown" error. On the write path a forbidden-existing slug is explicitly intercepted and masked as "unknown" rather than falling through to `doc.insert()` (whose unique-slug duplicate-key error would leak existence even louder). The tool error `code` for these two denial paths changes `PermissionDeniedError`→`InvalidArgumentError` (intended). A residual success-vs-error side-effect oracle remains and is out of scope.
</details>

## Pre-merge checklist
- [ ] **CI is green**
- [ ] Branch up to date with `develop`
- [x] Tests updated to assert byte-identical messages on both paths
- [x] Self-reviewed the diff

Local regression: `test_wiki_scopes_api` 39/39, `test_wiki` 30/30, `test_action_error_detail` 18/18.

Fixes #733

https://claude.ai/code/session_01HpbkcUnbmLZCH2ohy2q1pA

### DIFF
--- a/jarvis/tests/test_wiki_scopes_api.py
+++ b/jarvis/tests/test_wiki_scopes_api.py
@@ -15,7 +15,7 @@ from frappe.tests.utils import FrappeTestCase
 from frappe.utils import add_to_date, now_datetime
 
 from jarvis.chat import wiki
-from jarvis.exceptions import PermissionDeniedError
+from jarvis.exceptions import InvalidArgumentError
 from jarvis.permissions import JARVIS_ADMIN_ROLE, JARVIS_USER_ROLE
 from jarvis.tools.read_wiki import read_wiki
 from jarvis.tools.update_wiki import update_wiki
@@ -549,8 +549,15 @@ class TestWikiToolVisibility(_WikiScopeFixture):
 		frappe.set_user(PLAIN_USER)
 		page = read_wiki(slug=slugs["org"])
 		self.assertEqual(page["slug"], slugs["org"])
-		with self.assertRaises(PermissionDeniedError):
+		# #733: a foreign private page and a missing page raise the
+		# byte-identical error, so the exception can't be used to learn
+		# a private page exists.
+		with self.assertRaises(InvalidArgumentError) as forbidden:
 			read_wiki(slug=slugs["kwu_user"])
+		self.assertEqual(str(forbidden.exception), f"unknown wiki page: {slugs['kwu_user']}")
+		with self.assertRaises(InvalidArgumentError) as missing:
+			read_wiki(slug="wkscope-missing-slug")
+		self.assertEqual(str(missing.exception), "unknown wiki page: wkscope-missing-slug")
 		# The owner still reads their own page.
 		frappe.set_user(KW_USER)
 		page = read_wiki(slug=slugs["kwu_user"])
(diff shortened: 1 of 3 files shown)


kavin-114:
Adversarial review note (from an independent pass): the read path is fully closed, but the **write path still has a residual existence oracle** — when a caller supplies both `title` and `page_type`, a forbidden-existing Role/User page raises `unknown wiki page` while a genuinely-missing slug with the same args **succeeds** (creates the page). So success-vs-exception still distinguishes existence on the write path via `update_wiki(slug=<guess>, title=..., page_type=..., append_md=...)`.

This is the known success-vs-error side-effect oracle noted as out-of-scope: closing it fully would require either not creating on a miss (breaking page creation) or faking success for forbidden pages. Not a regression — that distinction predates this PR, which still closes the read-path class/message leak and masks the write-path *message*. Flagging so the residual is a deliberate, recorded decision. Consider a follow-up if enumeration via create-probing is in scope.

Navin-S-R (approved):

Issue #1082, opened 2026-09-02
Agents: tenant-level installations so admins provision once and users just run


kavin-114:
Child of #1058 (model B, target governance design). Today every user has their own Jarvis Agent Installation (own schedule, config, run-as, activation state), while the container roster, the activation ceiling and the catalog install counts are tenant-level. That mismatch produces: the per-user activation ceiling (`_activation_ceiling(doc.owner)`, ten users get ten live modules against a stated ceiling of one); Installed and Activity tabs scoped to the current user while cards show tenant-wide install counts; every user needing their own shadow to live sign-off; the dead `sync_status` and `synced_at` Installation fields the Admin tab renders as "-".

**Expected (owner model):** an admin installs, configures, trials in shadow and promotes an agent once per tenant, allows it for roles and named users, and users see only allowed live agents and run them (execution identity stays per run-as). The ceiling, install counts, Installed and Activity are then tenant facts.
**Suspected location:** `Jarvis Agent Installation` ownership model, `jarvis/chat/agents_api.py` (installs, activation, ceiling), Agents SPA tabs.

Interim A (deny by default, allowed users, admin allow pushes the roster, reviewer-gated promote) ships first; this issue is the design for the full model.

PR #1077, opened 2026-09-02
feat(agents): Stop button, live run progress, and role-scoped marketplace

kavin-114:
Fixes #1062 (child of #1058). Stacked on #1064; retarget to `develop` once #1064 merges.

## Summary

Operators get a Stop button and live progress on a running agent run, users without apply rights are no longer told to click a button they cannot see, and the marketplace only shows a user the Published agents their roles permit.

- Runs pane: Stop button while running (calls the new stop endpoint); Open Chat only once the run is terminal; `stopped` renders as its own state; ticking elapsed time plus the recent activity trail replace the static placeholder; the runs list polls every 10s while any run is running (paused on hidden tab).
- "Not loaded on your container" error now says "Ask your administrator to apply catalog changes" to non-reviewers; reviewers keep the click-through copy.
- Catalog visibility is scoped at the single enrichment choke point: non-admins see rows they are allowed AND Published, plus anything they installed themselves (an install survives a later role revocation so uninstall stays reachable); `allowed_roles` is stripped from non-admin list payloads; `get_agent` deep links get the same gate. Admins keep full parity. The locked "Available to: ..." cards are gone.

Risk: `get_agent` still returns `allowed_roles` (an existing test pins it); only the list endpoints strip it. Progress is status/elapsed/activity, not a per-step timeline (that needs delegate-side telemetry, tracked separately).

## Before / After

**Before** (develop): during a running run the pane offers Open Chat and nothing else - no Stop, no progress, a static placeholder.
before: running run on develop

**After**: Stop button, elapsed timer, activity trail; Open Chat only once terminal (see Evidence below).

## Evidence (integration branch on e2e.localhost, real dispatched run)

Running run: Stop present, no Open Chat, elapsed ticking, activity trail.
running run with Stop button

After clicking Stop in the UI: row `stopped` with "Stopped by operator.", Open Chat back, Stop gone; activity row `run_stopped`; the poll sweep afterwards left it untouched.
stopped run

Admin catalog parity (32 listings). A plain Jarvis User on the same site sees 30, all Published, no `allowed_roles` key; hidden: the admin-only `custom-app-learning` and the unpublished `ticket-triage-operator`.
admin catalog

Frontend: 90 files / 1393 tests green on Node 24. Backend CI ran green on this head via workflow_dispatch before the PR.

## Pre-merge checklist

- [ ] **CI is green** — the `tests` check on this PR passes (never merge on ❌)
- [x] Branch is up to date with its base (rebased onto the 1061 head)
[...]

Issue #1058, opened 2026-09-01
No way to stop a running agent run; wedged runs hang 3h until reaper


kavin-114:
A running agent run has no Stop control anywhere in the UI or API, so a run that will never finish sits `running` until the 3h stale-run reaper ends it. There is no manual override.

**Steps:**
1. Trigger an auditor whose delegate is not in the container roster (seen on e2e.localhost, `close-auditor`; container roster was empty).
2. Dispatch returns 202; the container rejects the turn (`INVALID_REQUEST: unknown agent id "agent-close-auditor"`) but nothing writes back.
3. The run shows `running` on the Runs tab with only "Run Now" / "Open Chat", no Stop.

**Expected:** an operator can cancel a running agent run and have it terminalize immediately.
**Actual:** no cancel endpoint, no `stopped` status, only `reap_stale_agent_runs` (3h) ends it. Macros (`stop_macro_run`) and chat (`stop_run`) both have a stop path; agent runs do not.

**Suspected location:** `jarvis/chat/agents_api.py` (no cancel endpoint), `Jarvis Agent Run` status options (`running/completed/partial/failed`), `jarvis/chat/agent_scheduler.py` (dispatch is detached).

**Proposal (not the required design):** add a whitelisted `stop_agent_run` mirroring `stop_macro_run` that aborts the gateway session for the run's `session_key` and terminalizes it, add a `stopped` status, and surface a Stop button.

<details><summary>Related defects from the same incident</summary>

1. **Wedge instead of fail-fast.** The gateway rejects an unknown agent id synchronously in ~1ms, but fleet returns 202 before that validation, so the deterministic rejection never propagates and the run hangs for 3h instead of failing immediately.
2. **Pre-dispatch check drift.** The bench "agent is loaded on your container" gate passed while the container delegate roster was actually empty (`delegate_ids=[]`); it trusts admin/bench state instead of the live roster, so it dispatched a turn the container then rejected.

</details>

kavin-114:
Two more UX gaps on the same Runs surface, seen while a run is in progress:

**1. "Open Chat" is shown during a running run.** It should appear only once the run reaches a terminal state (completed / partial / failed). The agent turn is detached and server-side, so opening a chat mid-run shows the operator nothing actionable and reads as if the run is interactive when it is not. Hide the button while `status == running`.

[...]

PR #1084, opened 2026-09-02
feat(agents): deny-by-default access, admin allow by roles and users

kavin-114:
Part of #1058 (governance decisions of 2026-09-02; supersedes the deferred D1). Stacked on #1077; retarget to `develop` once it merges.

## Summary

Admins now decide exactly which agents each role or named user may see, install and run; everything else is admin-only. Users never hit the "ask your administrator to apply" dead end, because the tenant container is loaded when the admin grants access, not when a user installs. Existing installs keep working through an upgrade patch.

- Access is deny by default: a listing with no allowed roles and no allowed users is visible only to admins. New `allowed_users` table next to `allowed_roles`; one `set_agent_access` endpoint (`set_agent_roles` kept as a shim).
- The pushed roster is the union of allowed Published listings and enabled installs, so an allowed agent is already on the container before anyone installs it.
- Admin tab gets one Access editor (two-column layout with the Installs table): role picker, frappe-ui Autocomplete people picker (same component as the roles picker) that lists users on open, unsaved indicator, single Save that always loads the agent on the workspace (restart only when the loaded set changes, stated in the helper text), with progress and done/failed toasts.
- Promote to live requires the reviewer role set (Skill Reviewer / Jarvis Admin / System Manager); plain users see no shadow or attestation language at all.
- Jarvis Admin can now disable, stop, uninstall and re-target other users' installs.
- Users with nothing allowed get "No agents have been made available to you yet. Ask your administrator." on every tab.

Risk: the inversion closes every unrestricted listing on upgrade; the `v2_18` patch grants each existing install's owner and run-as user by name so nothing stops running. 33 new backend tests + 27 new specs; frontend 1433 green.

## Before / After

**Before**: a plain Jarvis User saw the entire catalog (every unrestricted listing was installable by everyone), and the admin roles picker was a two-step control whose selection was silently lost without a second Save.
before: plain user sees full catalog
before: two-step roles picker

**After**: see Evidence below (deny by default, one Access editor with apply-now, plain user sees only what was granted).

## Evidence (integration branch on e2e.localhost)

Admin Access editor (two-column Admin tab, people picker open):
admin access editor

Plain user with nothing allowed (Featured tab):
plain user empty state

After the admin allowed one agent for the user's role (container restarted once, on the admin's save): the user sees exactly that agent, installs it with no shadow or promote language, and Run Now dispatches immediately.
plain user sees one agent
installed, no shadow copy
run dispatched

<details><summary>Design notes</summary>

Installations are per user while the container roster, activation ceiling and install counts are per tenant; that mismatch is the full model B tracked in #1082. This PR keeps per-user installs and moves the only tenant-wide side effect (the restart) to the admin. A user holding only Jarvis Skill Reviewer can still promote (same reasoning as `apply_agents`). The `legacy_empty_allows` reading is used only for the enabled-install leg of the roster so a tenant that upgrades before the patch runs keeps its agents loaded (dispatch still refuses until access is granted).

</details>

## Pre-merge checklist


kavin-114:
Code review (high effort, 6 findings) disposition, all addressed on the current head:

1. **Fixed (correctness)**: the Access editor reset unsaved drafts on any props refresh (e.g. toggling Enabled reloads the agent). A refresh now updates only the saved baseline; dirty drafts survive. 3 specs.
2. **Fixed (correctness)**: out-of-order people-search responses could overwrite newer results; a monotonic request sequence now drops stale responses and stale errors. 1 spec.
3. **Fixed (correctness)**: the roster's enabled-install leg admitted agents dispatch would refuse (`legacy_empty_allows`). Removed; the roster now uses exactly the dispatch admission (run-as user passes the deny-by-default gate). Tests flipped accordingly, plus a grandfathered-install-ships test for the real upgrade path.
4. Declined: extracting the sync-status poll into a shared composable (refactor scope); the failure toast now reuses `humaniseSyncStatus` so copy cannot drift.
5. **Fixed**: user validation in `set_agent_access` is one batched query.
6. **Fixed**: both roster legs share one roles/users map pair; the per-agent helper is no longer called per row.

Also fixed while greening CI: `search_users` registered as a non-list endpoint; the Jarvis Admin write path on other users' installs is now stated in app code (`_check_installation_write`) because User Permissions on link fields defeat the DocPerm alone; DocType JSON `modified` stamps bumped so `bench migrate` actually delivers the new field and perm rows to existing tenants.

kavin-114:
Fresh code review on the final head (5 findings) disposition, fixes in `cd7eb64d`:

1. **Fixed (correctness)**: `set_agent_access` now validates only users being ADDED; existing grants pass through unchanged (a deleted User row is dropped with a log line since the child Link cannot save it; a disabled user is kept, disabling is reversible). The `set_agent_roles` shim no longer fails a role-only save on a listing with a stale grant. 3 tests.
2. **Fixed**: grandfather patch prefetches User existence in one query instead of one `exists()` per identity.
3. Verified on the stacked head (#1085): the Administrator empty-state branch is wired from AgentsList (`isAdministrator` from the session, single `emptyState` computed).
4. Declined: `feedback_gate.mjs` duplication belongs to #1088 on develop, not this PR.
5. Declined: shared apply-poll composable (refactor scope), as before.

Navin-S-R (approved):

Issue #1083, opened 2026-09-02
Published operator agents can be installed and scheduled but can never run


kavin-114:
Child of #1058. The registry publishes operator agents (AP 3-way match, AR collections, bank reconciliation, cycle-count planner, reorder replenishment) but no dispatch path exists for the Operator nature: `run_agent_now` refuses them and the cron sweep silently advances the slot with no run row, activity or notification, while Configure still offers "Run automatically" and prints a next run date, and Apply still pushes their bundles (forcing restarts for zero function).

**Expected:** either an Approval-Board dispatch path for operators, or they are not offered for install.
**Actual:** a user installs, enables, schedules, and waits forever.
**Suspected location:** `jarvis/chat/agents_api.py` run_agent_now nature gate, `jarvis/chat/agent_scheduler.py` _sweep_one, `jarvis/agents/registry.json`.

Interim: the UX polish branch marks these listings Coming Soon so they stop being installable; this issue tracks the real dispatch path.

PR #1085, opened 2026-09-02
chore(agents): UX polish batch from the agent-flow review

kavin-114:
Part of #1058 (product/UX review of 2026-09-02). Stacked on #1084; retarget to `develop` once it merges.

## Summary

Fourteen small, reviewed fixes that make the Agents surface honest and readable for an accountant: shadow runs are visibly marked as previews, dead ends explain themselves, install failures stay on screen, and the catalog stops leaking internals.

- Runs: the "Preview / not a compliant attestation" pill and "Open dashboard" now render (two fields were missing from the run list); a stopped run says "Stopped by operator." instead of "No findings"; activity labels for stopped, promoted, demoted.
- Detail page: visible one-line hints when Run Now is disabled (enable first, shadow scribe); "Reads these records" lists the DocTypes an agent reads; a failed install shows a persistent inline error, not a 2-second toast; `allowed_roles` no longer sent to non-admins.
- Copy: em dashes and curly quotes out of user-facing strings; three over-long errors trimmed; not-found never leaks the DocType name; uninstall warning says dashboards are kept; take-to-chat surfaces failures instead of opening an empty chat.
- Catalog: categories derived from the registry domain with real labels (Accounts Payable, Bank and Reconciliation, HR and Payroll, GST Compliance…); the six operator listings with no dispatch path become Coming Soon (tracked in #1083).
- Plain users see two tabs only, "Agents" and "Activity" (Featured and Installed are redundant once access is deny by default); admins and reviewers keep all four. Deep links to hidden tabs fall back to Agents.
- `set_enabled` no longer marks the catalog dirty on a no-op toggle (root cause of the spurious "leave site?" prompt).

Risk: mostly copy and read-path fields; the operator downgrade and category mapping change what customers see in the catalog (owner ratified). Frontend 1452 tests green.

## Before / After

**Before**: an install failure was a 2-second toast with nothing left on screen, and the Overview never said which records an agent reads.
before: transient install toast

**After** (integration bench on e2e.localhost):
after: reads these records chips
after: persistent inline install error

Plain user's catalog after the change: Agents and Activity only (deep link to #installed lands on Agents).
after: plain user two tabs

Also verified live: operators show Coming Soon with proper category labels; the stopped-run copy and status filter; the Preview pill on shadow runs.

## Pre-merge checklist


kavin-114:
Code review (high effort, 3 findings) disposition, fixed in `544c44b2` and `d37c0ebd`:

1. **Fixed (correctness)**: `take_finding_to_chat` committed the conversation before seeding it, so a failed seed (no model configured) left an orphaned empty "Finding: ..." conversation on every click now that the SPA no longer navigates on failure. The server now deletes the conversation and any partial message before returning `ok: false` with the reason. Two tests (failure leaves no row; success path unchanged).
2. **Fixed**: list-field parsing extracted to `lib/parseListField.js`, used by both `needs` and `readsRecords`.
3. **Fixed**: `categoryTitle` extracted to `lib/agentCategory.js`, imported by both AgentsList and AgentDetail.

kavin-114:
Fresh code review on the audit-fix head (5 findings) disposition, fixes in `d80dcf91`, `eb319baf`, `0ba5475e`:

1. **Fixed (correctness)**: breadcrumb no longer sticks on a blank skeleton when the agent fetch fails; it falls back to the slug.
2. **Fixed (correctness)**: the DocType.field extractor no longer matches capitalised hosts inside markdown link targets (URL guards added; the multiword DocType allowance was tried and reverted because it swallowed sentence-initial verbs).
3. **Fixed (correctness)**: `get_doc` on a Single with `names` returns the documented batch envelope; plain calls keep the flat document.
4. **Fixed**: the three copies of the debounced, sequence-guarded link search are now one `useLinkSearch` composable (ConfigForm, AgentAccessEditor, FilterValueControl, the last now with its first spec).
5. **Fixed**: finding display extraction is computed once per finding (memo Map) instead of 2-3 times per render.
Skipped: `formatTime12h` via dayjs (not a direct dependency of the frontend package).

Navin-S-R (approved):

PR #1087, opened 2026-09-02
feat(agents): live step timeline for agent runs

kavin-114:
Part of #1058 (item 5, full scope). Stacked on #1085; retarget to `develop` once it merges.

## Summary

A running agent run now shows what the agent is actually doing, step by step, instead of a static "run in progress" line: every read, query, report and balance check the agent makes against the bench is recorded as a step and rendered as a live timeline in the run pane (also inspectable after the run ends).

- New `Jarvis Agent Run Step` DocType (Link-free like Activity), written from the plugin tool-call path when the calling session is an agent run: humanised labels ("Queried Warehouse, 4 rows", "Ran report Stock Balance", "Checked balance for Stock In Hand"), durations, and the tool's error text on failed steps. The agent's own config reads show as "Read engagement configuration". Never records row contents, only DocType names, counts and report names.
- "Dispatched to the agent" is recorded at launch and "Recorded N findings" at writeback, so a run that never starts or never writes back is visually obvious.
- `list_run_steps` endpoint (owner or Jarvis Admin), ordered by time with stable numbering; SPA `RunStepTimeline` built from frappe-ui Badge/FeatherIcon in the Activity tab's row style, polled every 5 s while running (visibility-guarded, single timer), consecutive identical steps folded into one row with a count.

Risk: one indexed lookup per plugin tool call to resolve the run; the step insert is best-effort and can never fail a tool call. 49 backend test cases, frontend 1487 green.

## Before / After

**Before**: the running pane only said "Run in progress, findings appear when it completes" with a loose activity box.
before: static running placeholder

**After** (integration bench, live run): header with elapsed time and step count, then the timeline.
after: live step timeline

<details><summary>Design notes</summary>

Steps come from the bench side of every `jarvis__*` tool call (authenticated by the run session), so no container or plugin change is needed. Check-level progress reported by the delegate itself ("evaluating check 2 of 5") would need a plugin tool and bundle protocol; that is a follow-up. `record_agent_run` records its own writeback step because only it knows how many findings persisted. The raise-path step is lost on a real 500 (transaction rollback); the `ok: false` envelope path, which is how tool errors normally arrive, is covered.

</details>

## Pre-merge checklist

### DIFF
--- a/frontend/src/api/agents.js
+++ b/frontend/src/api/agents.js
@@ -55,6 +55,13 @@ export const listAgentActivityPage = (p = {}) =>
 		page_length: p.page_length || 20,
 	});
 
+// One run's STEP TIMELINE, oldest first (jarvis#1062): what the bench observed
+// the run do - the launch dispatch, each jarvis__* tool the delegate called back
+// with, and the findings writeback. Steps carry shapes (DocType names, report
+// names, counts), never row contents. Ownership-gated server-side; a run the
+// caller cannot read returns an empty timeline. -> { steps: [...], count }
+export const listRunSteps = (run) => call(AG + "list_run_steps", { run });
+
 // Seed a new conversation from a finding and land the user in live chat.
 // -> { ok, conversation, run_id, reason }
 export const takeFindingToChat = (finding) => call(AG + "take_finding_to_chat", { finding });
--- a/frontend/src/pages/agents/FindingsPanel.spec.js
+++ b/frontend/src/pages/agents/FindingsPanel.spec.js
@@ -369,7 +369,7 @@ describe("C3b: a finished run keeps its timeline, collapsed", () => {
 		expect(apiAgents.listRunSteps.mock.calls.length).toBe(afterFlip);
 	});
 
-	it("shows an error step in the red tone", async () => {
+	it("shows an error step in the red tone, with the tool's message", async () => {
 		apiAgents.listRunSteps.mockResolvedValue({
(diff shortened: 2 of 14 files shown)


kavin-114:
Code review (high effort, 4 findings + 1 refuted item taken anyway) disposition, all in `87d03f12`:

1. **Fixed (correctness)**: batch `get_doc` labelled a failed read as "Read X, N documents" from the requested count; now gated on the call resolving, prefers the returned count, and falls back to "Looked up X" with the error in detail.
2. **Fixed (correctness)**: a measured 0 ms rendered nothing; "0ms" now renders, and folding two steps without timings no longer fabricates a 0.
3. **Fixed**: one run-row read per delegate tool call instead of two; the capability resolution now carries owner/status and is reused by the step hook (tests count the reads).
4. **Fixed**: dead `record_agent_run` branch removed; the writeback step is tested end to end through the real tool, sharing the `plural` helper.
5. Also taken (the refuted candidate): a gated write parked as pending confirmation is labelled "Proposed <action>, awaiting confirmation" instead of reading as done.

kavin-114:
Fresh code review on the re-stacked head: no correctness findings; the one cleanup finding (a third copy of the debounced, sequence-guarded link search) is addressed on the base branch (#1085 `d80dcf91`: shared `useLinkSearch` composable) and lands here on the final re-stack.

Navin-S-R (approved):

PR #1089, opened 2026-09-03
ci: adopt version-N-hotfix branches; stable only takes release PRs

kavin-114:
## Summary

Adopts the Frappe-style hotfix layer: backport PRs now target `version-15-hotfix` / `version-16-hotfix`, and the stable branches `version-15` / `version-16` only accept the release PR that merges their hotfix branch. Anyone opening a backport notices: a PR straight into a stable branch now fails the new `release-source` check with a message telling them where to retarget.

What changes on `develop`:

- `release-guard.yml` (new): `release-source` job fails any PR into `version-N` whose head is not `version-N-hotfix` in this repo, and on a release PR also fails unless `__version__` moved up on the right line. Runs on `pull_request_target` so the base branch's copy is what runs (nothing checked out, no head code). Advisory until listed as a required check.
- `ci.yml`: push trigger covers `version-16` and `version-16-hotfix`; auto-cancel keys on `pull_request`. Deliberately no `edited` trigger; a retargeted PR is failed by the guard until pushed to or reopened.
- `jarvis/tests/test_release_notice.py`, `test_www_release_notice.py`: fixed `99.0.x` sentinel versions instead of literal `0.0.2`, so the hotfix PRs' version bump does not flip five tests; new direct tests for the version parser.
- `.pre-commit-config.yaml`: `no-commit-to-branch` guards `develop` plus `version-N` and `version-N-hotfix`; drops the deleted `beta`.
- `.mergify.yml`: backports retarget to the hotfix branches.
- `CONTRIBUTING.md`: new "Branches and releases" section. PR template no longer says "up to date with `main`".

Counterparts: #1090 (version-16-hotfix) and #1091 (version-15-hotfix) are cherry-picks of this branch's commits `44869a77`, `5ffd17c5`, `4288b3cf`, `981977e1`, `a8ed19b4`, `501a4bd3` plus their version bump. Re-diff them against this PR before merging if anything else lands here.

Rollout order: merge this PR first (Mergify reads `.mergify.yml` from `develop` only, and its Workflow Automation product is not activated yet, so no automated backport can land in the window either way), then #1090 and #1091 with merge commits. Rulesets are a manual follow-up in repo settings: merge-only on version-16 and a `version-*-hotfix` ruleset before the first release; `release-source` as a required check only after the first release PR has merged the guard file onto each stable branch.

## Pre-merge checklist

- [x] CI is green
- [x] Branch is up to date with its base
- [ ] New/changed behavior has tests (workflow and docs only; no test surface)
[...]

PR #1091, opened 2026-09-03
ci: adopt version-15-hotfix; stable only takes release PRs; bump to 15.2.0

kavin-114:
## Summary

version-15 counterpart of #1089: brings the hotfix branch model to the `version-15` line and bumps `__version__` to `15.2.0`, so the first release PR (`version-15-hotfix` -> `version-15`) ships a real version instead of `0.0.1`. Anyone running `bench version` on a v15 site notices.

- Hand-applied to version-15's own `ci.yml` (it is a separate file from develop's): push trigger covers `version-15-hotfix`, auto-cancel keys on the event type. `release-guard.yml`, the retarget-aware ci.yml jobs and the sentinel-version tests are cherry-picked from #1089 (`5ffd17c5`, `4288b3cf`, `981977e1`, `a8ed19b4`, `501a4bd3`).
- `.pre-commit-config.yaml`, `CONTRIBUTING.md`, PR template: same as #1089.
- `jarvis/__init__.py`: `0.0.1` -> `15.2.0` (10 backports since v15.1.0 include feature work, hence minor).

Merge #1089 first, then this into `version-15-hotfix`, then open the release PR from `version-15-hotfix` into `version-15`. Do not squash.

## Pre-merge checklist

- [x] CI is green
- [x] Branch is up to date with its base
- [ ] New/changed behavior has tests (workflow, docs and a version string)

https://claude.ai/code/session_01BhuU9nWobCxQe8JWiCzuo6


kavin-114:
Review addressed in the latest commit on all three PRs (#1089, #1090, #1091 carry the same change).

**Adopted**
- #2 and alternative A: `release-guard.yml` now triggers on `pull_request_target`, so the base branch's copy runs. A head that predates the file, or a fork that deletes it, can no longer produce "no check". Nothing is checked out and no head code runs, so the usual `pull_request_target` hazard does not apply.
- #4: on a release PR (base `version-N`, head `version-N-hotfix`) the job also fails unless `__version__` increased and its major matches the line. Reads the file through the API, no checkout. Logic exercised against the real branches: 0.0.1 -> 16.2.0 passes, 15.2.0 into version-16 fails as wrong line, identical tips fail as not bumped.
- #1 and #3: header and CONTRIBUTING wording no longer assert enforcement the rulesets do not yet provide. The check blocks a merge only once it is listed as a required status check.

**Repo settings, not the diff** (owner action): merge-only on version-16, rulesets for `version-*-hotfix`, required checks on stable and hotfix. One ordering note that follows from the base-run trigger: `release-source` must be added as a required check only after the first release PR has merged the guard file onto each stable branch, otherwise the release PR itself can never report it and stays unmergeable.

**Not adopted**
- Alternative B (sentinel `999.0.0` versions): kept the derived versions. They also pin the exact installed+1 boundary, and the private `_version` import fails loudly on rename.
- #5 (shell tests for the guard): skipped for now, no repo convention for testing Actions shell.

Consequence to be aware of: the guard no longer reports on these three PRs, because none of their base branches has the file yet. It first runs on PRs opened after #1089 lands on develop.

kavin-114:
Correction to the previous comment, found by live test. Adding `edited` to `ci.yml` with job-level skips was wrong: a body edit produced an all-skipped run that shared the concurrency group with the in-flight run and cancelled it on all three PRs (the cancelled runs you can see on the earlier head commits). The deeper problem: a skipped job still posts a "skipped" check on the head commit, GitHub reads the latest check per name, and skipped satisfies a required check, so a title edit after a red run would have masked the failure.

Reverted in `501a4bd3` (#1090 `2fa163bb`, #1091 `4a02fbc0`). `ci.yml` is back to the default PR event types. The retarget case (I1) now lives in `release-guard.yml`: an `edited` event that carries `changes.base` fails the guard with "push a commit or close and reopen", and either of those makes `ci.yml` run against the new base and re-runs the guard.

venkat102 (approved):

PR #1098, opened 2026-09-03
ci: tag and publish a GitHub Release automatically on stable-branch merges

kavin-114:
## Summary

Automates the last manual step of a release. When a release PR merges into `version-15` or `version-16`, the new `Release` workflow tags the merge commit `vN.x.y` from `__version__` and publishes a GitHub Release with notes generated from the previous tag on that line. Whoever cuts a release notices: no more hand-made tags.

- Idempotent (existing tag = no-op) and re-runnable from the Actions tab via `workflow_dispatch`.
- "Latest" is given only to the highest `version-N` line, so a v15 release never displaces v16.
- Refuses a `__version__` that is not on the branch's line (belt to `release-guard.yml`'s braces).
- Dry-run against the real branches: v16.2.0 on version-16 (previous v16.1.0, latest=true), v15.2.0 on version-15 (previous v15.1.0, latest=false), re-run with the tag present = no-op.

Repo settings: the "Release tags" ruleset has no tag `creation` rule (GitHub refuses the built-in Actions app as a bypass actor), only deletion, update and the `vN.N.N` name pattern. Done 2026-09-03.

Backports to both hotfix branches follow; the open release PRs #1096 and #1097 pick them up automatically.

## Pre-merge checklist

- [x] CI is green
- [x] Branch is up to date with its base
- [ ] New/changed behavior has tests (workflow only; logic dry-run as above)

https://claude.ai/code/session_01BhuU9nWobCxQe8JWiCzuo6

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,64 @@
+name: Release
+
+# Tags and publishes a release whenever a stable branch (version-N) moves. Under the hotfix
+# model the only thing that lands on a stable branch is the release PR from version-N-hotfix,
+# and that PR carries the __version__ bump (release-guard.yml refuses it otherwise), so every
+# push here IS a release: read __version__, tag the merge commit vN.x.y, publish a GitHub
+# Release with notes generated from the previous tag on the same line.
+#
+# Idempotent: if the tag already exists nothing happens, so a failed run can be re-run from
+# the Actions tab (workflow_dispatch) without side effects. "Latest" goes only to the highest
+# version-N line, so a v15 release never displaces the v16 one.
+#
+# The token is github-actions[bot]; the "Release tags" ruleset must list the GitHub Actions
+# app (Integration 15368) as a bypass actor or the tag push is refused.
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


kavin-114:
Review addressed in the latest commit (also cherry-picked to #1099 and #1100).

- Finding 1 (not resumable after a partial failure): tag and Release are now checked separately. A re-run after a pushed tag with no Release publishes the Release; both present is a no-op. The previous-tag lookup excludes the tag being released so notes still start from the right place. Dry-run of all four paths (fresh, tag-only, both, wrong branch) behaves as intended. CONTRIBUTING wording corrected from "idempotent" to "resumes whatever step was missing".
- Finding 3 (loose glob): anchored regex `^version-[0-9]+$`, so a hotfix branch under workflow_dispatch gets the intended error.
- Finding 2 (ruleset bypass actor for the GitHub Actions app, id 15368): repo-settings step for the owner, called out in the PR body and the workflow header. Until it is added the first run fails at the tag push and a re-run after the fix completes it.

Navin-S-R (approved):

PR #1136, opened 2026-09-05
feat(chat): context meter and Compact with hint

kavin-114:
## Summary

Every chat now shows how full the assistant's context window is, and the user can compact the conversation, optionally saying what to keep ("keep the invoice inputs"), before the model hits its limit.

## What changed
- A context ring next to the model picker fills in the Jarvis gradient and turns terracotta at 80 percent; clicking it opens a popover with the exact numbers (in use, auto-compact point, last reply in/out, compactions) and a Compact button.
- Compact from the ring's popover, the top-right menu, or by typing `/compact <what to keep>`. A folding-lines banner shows while it runs (about 6 s) and the composer is locked; a chip confirms when done.
- Sends are refused at the front door while a compaction is in flight; queued or macro turns wait and dispatch when it clears.
- Automatic mid-turn compaction by the runtime now shows a status on both transports instead of nothing.
- Settings "This chat" shows the measured context reading with the measured figures, not under the estimate badge.

## Risk and verification
- Compact runs as a background job on one pooled gateway socket with a 300 s lock, so it cannot starve turns. The runtime learns the new size only on the next reply, so the meter says so rather than guessing.
- Backend: 13 touched modules green on the local site (two pre-existing local-state failures reproduce identically on develop). Frontend: 1738 specs green on Node 24, build clean. Live: real compact job with a hint against the e2e tenant, 7.6 s, count and lock verified.
- Live UI check on e2e.localhost through a sidecar deployment of this branch (the tenant's provider was rate-limited, so the compaction itself was declined; the failure path shows honest copy). Animation preview: https://claude.ai/code/artifact/73a7b69c-a4ef-4564-b570-9b5f2b0b4d55.

<details><summary>Runtime facts this relies on</summary>

`sessions.list` rows carry `totalTokens` (current prompt size) and `contextTokens` (capacity); the bench already snapshots both per turn. The only way to pass a hint is the runtime's text command `/compact <hint>` over `chat.send`; the `sessions.compact` RPC has no hint field and does not refuse a running turn, so locking lives bench-side. The command path emits one `chat` final event and no lifecycle frames; the bench waits for that only. Verified on image 2026.6.8 (live) and 2026.7.1-2 (pinned).

</details>

## Before / After

**Before** (composer has no context meter)

before-composer

**After** (context ring next to the model picker)

after-context-ring

Click the ring: details popover with Compact

after-context-popover

Compact dialog with the optional hint:

after-compact-dialog

While compacting (banner, locked composer, pill drain):

after-compacting-banner

Settings, General and Usage panes:

after-settings-general
after-settings-usage

## Pre-merge checklist

> ℹ️ A red check only blocks the merge where the branch ruleset lists it as required.
> Honoring this checklist is what keeps broken changes out of UAT. See
> [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] Branch is **up to date with its base** (`develop`, or `version-N-hotfix` for a backport)
- [ ] If the base is `version-N`: this is the release PR from `version-N-hotfix`, `__version__` is bumped, and `release-source` is green

https://claude.ai/code/session_01B3y7PDPWZZtNVRRQbjTCmD


Navin-S-R (approved):

kavin-114:
@mergify backport to version-15 and version-16

kavin-114:
@mergify backport version-15-hotfix version-16-hotfix

PR #1207, opened 2026-09-10
feat(chat): classify turn failures into actionable headlines and hints

kavin-114:
## Summary

Chat turn failures now get an accurate headline, a next-step hint, and a provider status link where one is identifiable, instead of a raw or generic error string. Each card is a title plus one guidance line (status link and raw-error Details inline), the shape ChatGPT/Claude/Copilot use for a failed turn. Retry is hidden for failures that need input, config, or account changes rather than a retry (for example a server file-permission error).

- One shared, ordered classification ruleset (JS + Python) replaces three previously drifting per-surface copies (`frontend/src/lib/errors.js`, the Desk `Panel.vue`, `turn_handler.py`).
- Wired into desktop ChatView, DashboardChatPane, Desk Panel, and mobile ChatView.
- Fixed during review: Desk Panel and mobile ChatView now also carry the live `run:error` event's classification through to their error UI, instead of only reclassifying from the persisted text (this closed a same-screen mismatch between the banner and the message card for the same failure).
- 88 frontend tests and the Python classification tests (fixture-shared between both languages) pass. Ruff is clean.

<details><summary>Verification</summary>

- Hosted CI green on the rebased branch: lint, frontend-build, frontend-tests, all four backend shards (incl. `test_turn_handler.py`, `test_pump_pipeline.py`), coverage.
- Local: 88 frontend tests, Python classification fixtures (shared with JS), ruff, Vue script syntax.
- Independent `/code-review high`: 7 findings, 5 fixed, 2 by design; per-finding status in the PR comment.
- Deployed on local `integration/chat-error-messages` (e2e.localhost) for owner testing; Before/After, dark mode and mobile screenshots below are from that deploy.

</details>

## Before / After

Same three persisted failures on e2e.localhost (EPERM on the agent's database file; Anthropic 529 overloaded with `provider=anthropic`; OpenAI 429 `insufficient_quota`).

**Before** (develop: two generic buckets, Retry always offered, no status link; rendered by applying the develop build's headline/hint copy to the same rows)

before-desktop-light

**After** (this branch, live build): title, one guidance line with the status link and a Details toggle inline (expanded on the Claude card), Retry only where it applies.

after-desktop-light

<details><summary>After: dark mode, mobile (PWA), Desk panel on a live failure</summary>

after-desktop-dark
after-mobile-light
after-desk-panel-live-failure

</details>

## Pre-merge checklist

> ℹ️ A red check only blocks the merge where the branch ruleset lists it as required.
> Honoring this checklist is what keeps broken changes out of UAT. See
> [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] Branch is **up to date with its base** (`develop`, or `version-N-hotfix` for a backport)
- [ ] If the base is `version-N`: this is the release PR from `version-N-hotfix`, `__version__` is bumped, and `release-source` is green

https://claude.ai/code/session_01DC4kzTDkGKnA6ZpCzehvKG


kavin-114:
**Independent code review (`/code-review high`) on `60da7fbe`: 9 candidates, 7 distinct findings, 0 refuted. Status after `c96caf12`:**

| # | Finding | Status |
|---|---|---|
| 1 | Retry button gated on the classification (`ChatView.vue`, `Panel.vue`) | **By design.** The spec hides immediate Retry for failures that need input, configuration, access, or an account change (e.g. EPERM). The review's "loose substrings" concern does not hold against the actual rules: `billing` needs `insufficient credit`/`credit balance`/`out of credits`, `permission` needs `permission denied`/`access denied`, `authentication` has no bare `invalid`. |
| 2 | `AgentUnreachableError` override narrowed to gateway/timeout text (`turn_handler.py`) | **By design**, documented in `docs/chat-errors.md`, covered by `test_specific_rejection_survives_transport_wrapper`. Known tradeoff: a gateway-level 429 on the WS handshake reads as a model request limit (wrong noun, right action: retryable, wait). |
| 3 | Dashboard chat pane never used the live `run:error` code | **Fixed.** Same `errorMeta` pattern as ChatView and the PWA. |
| 4 | Bare word "together" matched Together AI, and outranked a named Anthropic | **Fixed** (suffix mandatory) + 4 new JS assertions. |
| 5 | Desk widget dropped the provider context, so its status link was weaker | **Fixed.** The widget now passes the failed row's persisted `provider` (from `get_conversation`) like the other surfaces. |
| 6 | `errorNote()` called 5x per bubble in PWA and dashboard templates | **Fixed.** Classified once in the `items`/`bubbles` computed. |
| 7 | Rules file parsed by string-split in Python (fragile) | **Hardened.** Anchored regex with a clear `RuntimeError`; imported at module load so a broken rules file fails at deploy time, not in the error-reporting path. Switching to a `.json` was left out on purpose (bundler compatibility noted in the doc). |

Not re-run through the full review workflow after the fixes; happy to do a second pass on request.

venkat102:
## Review — multi-dimension panel (12 lanes)

Ran our standing Phase-1-floor review loop over this branch (correctness, security, perf/concurrency, resilience, edge-cases, testing, operability, maintainability, frontend, UX, chat-safety, + an approach/simplification pass). Every lane returned an explicit verdict. Findings below are deduplicated across lanes; the bracket tags show which lanes independently converged on each one.

### Verdict: no Critical, but a cluster of **Important** defects worth addressing before merge

The design is sound and the consolidation is real — three drifting per-surface classifiers collapse into one ordered ruleset, security is clean (XSS / link-injection / prototype-pollution / info-leak / auth all confirmed clean), the 52-case fixture proves **52/52 JS↔Python parity on ASCII**, the three modified existing tests were **verified correct, not test-bent** (checked against rule text + docs + the untouched transcript fixture), and the cross-tree imports resolve in all three build systems. The Important findings cluster in the two *new* behaviors this PR introduces — retry-gating and cross-surface classification.

---

## Important

**1. Rule-order collisions: a bare HTTP code / broad token outranks a specific reason → wrong remedy + wrong Retry state** `[resilience, edge-cases]`
`jarvis/public/js/turn_error_rules.mjs`: `rate-limit` (L93) precedes `model-unavailable` (L125); `not-found` (L149) precedes `service-unavailable` (L157) / `timeout` (L165). Because the backend classifier is authoritative (its `code` isn't `gateway`/`provider`, so the SPA never re-derives), the wrong code sticks:
- `"429 model_not_found"` → `rate-limit` (retryable **true**, "wait and retry") instead of `model-unavailable` (retryable false). A decommissioned model gets a Retry that can't help. (`404 model_not_found` correctly → `model-unavailable`; the ordering was fixed for 404 but not 429.)
- `"503 … upstream not found"` → `not-found` (retryable **false**) instead of `service-unavailable` (retryable true) — a transient 5xx loses its Retry. Realistic for CDN/proxy 503/504 bodies.

Both verified in JS **and** Python. Fix: move the specific-reason rules ahead of the numeric-HTTP rules they collide with (or negative-lookahead the numeric branches); add fixture cases so the drift-guard pins it.

**2. `cancelled` renders as a red error alert (+ a dangling "· Details") on 3–4 surfaces, and the Desk widget loses Retry for aged-out turns** `[resilience, ux, edge-cases, frontend]`
Only desktop `ChatView.vue:992` mutes `code === 'cancelled'` (`role="status"`). `DashboardChatPane.vue:72`, Desk `Panel.vue:436` (`role="alert"`) and `pwa/.../ChatView.vue:849` (`role="alert"`) fall through to the alarm-red failure card. Triggered by ordinary paths (`USER_CANCEL_REASON` / `AGE_OUT_REASON`, `admission.py:67-68`).
- `cancelled` is the only rule with `hint: ""`, so the separator+Details fragment renders with no lead-in → a **leading dangling `· Details`** on all four surfaces.
- **Regression (diff-confirmed):** Desk's Retry block is now `v-if="retryable"`, and `cancelled` is retryable=false — so the widget hides Retry for the aged-out case whose own text says *"Please try again."* The `cancelled` rule conflates user-cancel (no retry) with queue age-out (retry wanted).

Fix: port the muted `cancelled` branch to the other three surfaces; gate the `· Details` separator on `hint` being non-empty; split/relabel age-out so it stays retryable.

**3. ReDoS regression — unbounded, model-supplied error text through O(n²) regex** `[perf, security]`
`turn_error_rules.mjs:46` (`device.*(pairing|signature).*invalid`) and `:70` (`input.*exceeds.*token`) — two unbounded `.*` gaps backtrack quadratically on both engines. The old `_classify_error` used linear substring checks, so this is a regression, and `err_text` (`event.get("error")`) is never length-capped before classify/persist (`turn_handler.py:2213`). Measured: 160 KB → ~7.6 s (Python) / ~1.7 s (Node), the latter ×~10 unmemoized calls per desktop render (Finding 8) → multi-second tab freeze for anyone viewing the conversation. HTTP-code alternations measured linear (safe). Fix: cap input in both consumers (`text[:8192]` in `classify_error_text`, `.slice(0,8192)` in `turnErrorInfo`) before matching; optionally bound the two patterns (`.{0,80}`).

**4. PWA banner disagrees with its own inline card for the same event — reintroduces the #702 mismatch** `[ux, frontend, chat-safety]`
`pwa/.../ChatView.vue:617` computes the live banner via `turnErrorInfo(p.error, p.code)` with no provider context (and before `load()` refetches the row), while the card (`errorNote(m)`, L191) passes `m.provider`. For `service-unavailable`/`timeout`/`connection` (the "Claude via OpenRouter" case the code comments call out) the banner shows generic copy + no link while the card shows provider-specific copy + a status link — both on screen, disagreeing. Fix: give the banner the same provider context, or reuse the card's copy once available.

**5. Transient transport failures now classify to non-retryable codes and hide Retry** `[resilience, correctness]`
The narrowed override (`turn_handler.py:2208`, rescues only `gateway`/`timeout`) lets `AgentUnreachableError("No endpoints found")` → `model-unavailable` (retryable false, though OpenRouter "no endpoints" is often transient) and `"all providers failed"` → `models-exhausted` (retryable false); a recv-path SSL-cert error → `connection`, also not overridden. The pre-ack handler's own comment calls these "real, retriable." Fix: make transient strings like `no endpoints found` retryable, and/or widen the override so an `AgentUnreachableError` carrying no upstream `exc.code` always maps to a retryable code. (`models-exhausted`'s non-retryable is partly by-design — worth an explicit sign-off.)

**6. A benign edit to the rules file silently degrades all classification to `gateway`, undetected** `[operability]`
`error_taxonomy.py:20-27` guards only the `export default [ … ];` wrapper. `export default [];` parses fine → `_RULES = []` → every error classifies `gateway` forever, no exception/log/metric — the "check that silently stopped checking" failure mode. Also `_load_rules` gives a friendly error for only 1 of 4 malformed cases (bad JSON → bare `JSONDecodeError`; missing key → `KeyError`; bad regex → `re.error` with no which-rule). Fix (cheap): `assert len(rules) >= N` + a `frappe.logger(...).info("loaded %d rules")` at import, and wrap the parse to re-raise a named error identifying the offending rule.

**7. The provider status link is driven by text-scanning on error rows, and that scan picks a provider by array order** `[edge-cases, approach]`
`message.provider` is stamped **only on the success path** (`_effect_usage`→`_stamp_reply_model`); errored/cancelled turns never run the `usage` effect (`settlement.py`), so `context.provider` is essentially always empty on the failed rows this feature serves — `providerFor` falls to its text-scan branch as the *normal* case. That branch's OpenRouter/Groq/Together short-circuit (`turn_errors.mjs:69-76`) resolves by **declaration order**, not by which provider is implicated: `"mistral and groq both timed out"` → a Groq link, breaking the deliberate "ambiguous → no link" safety for those three. Fix: drop the short-circuit (let them obey `matches.length === 1`) or scope it to aggregator + model-brand; consider stamping a best-known provider onto error rows.

**8. Test gaps on the two behaviors the PR is about** `[testing, frontend]`
- `retryable: true` is **never asserted anywhere** — mutation-verified that flipping `service-unavailable` to non-retryable passes the whole suite. The headline UX change is unproven.
- No component/interaction tests for desktop / Desk / PWA (only one narrow Dashboard case) — retry-gating, banner/card consistency, `rawOpen` untested at component level.
- The `service-unavailable` provider-copy override (headline/hint) and the `explicitCode === "provider"` refinement leg are untested.
- a11y / perf: desktop `errorInfo(m)` is unmemoized (~10 calls/render, amplifies #3 — precompute like the other surfaces); PWA adds `role="alert"` to cards rendered from **history** (announced assertively on load); Dashboard's richer card has **no role at all**.

**9. An in-scope surface in the same PWA app was not migrated** `[chat-safety]`
`pwa/src/lib/notifications.js:73` still does `body: e.error || …` — raw provider/JSON text (e.g. the `EPERM …/agent.sqlite` path) lands verbatim in a persisted notification a user opens without visiting chat, one file over from the migrated `ChatView.vue`. Fix: `turnErrorInfo(e.error, e.code)` → use `info.headline`. (Separately, `jarvis_mobile` is a 5th `run:error` surface still showing raw text — cross-repo, reasonably a follow-up, but the "one taxonomy" thesis is incomplete until it's ported.)

---

## Minor (non-blocking)

- **Cross-language non-ASCII divergence** `[correctness (executed both engines), testing, approach]`: JS (ASCII `\b`/`\w`, simple case-fold) vs Python (Unicode, full fold) diverge on non-ASCII adjacent to a keyword/digit — e.g. Turkish `İnvalid api key` → live `authentication`/no-retry vs reload `gateway`/retry. Low frequency (error text is usually ASCII; would bite CJK providers like Moonshot/Kimi), but a real reintroduction of the target bug the ASCII fixture can't see. At minimum add non-ASCII fixture cases + document ASCII-only classification. (This is the strongest motivation for the alternative below.)
- **Copy/actor** `[ux]`: `safety` hint says "contact the model provider" (a B2B end user has no such relationship — siblings route to admin/account owner); `internal`/`gateway` name *only* support with no admin fallback; `billing`/`recovery-expired` hints reference "trying again" while Retry behaves inconsistently.
- `internal` and `gateway` share a byte-identical headline `[operability]` — on-call can't tell them apart from a screenshot.
- **Maintainability** `[maintainability, frontend]`: PWA passes the whole `m` as context (others pass `{provider}`) — latent collision risk; nested-ternary code selection (`turn_errors.mjs:78`) → guard clauses; `.jv-err-link` CSS triplicated across build trees (add a "kept in sync" comment); missing type hint + docstrings on `classify_error_text`/`_load_rules`; `ActionError.vue` still says "Show details" (native `<details>`) so the new "mirrors ActionError" comment is now inaccurate.
- `docs/previews/jarvis-error-messages.html` hardcodes a snapshot of the copy `[chat-safety]` — same drift class, relocated to docs.
- Retiring the `provider` catch-all narrowed two trigger words `[chat-safety]`: `"insufficient funds"` / `"your credit expired"` now → `gateway` (retryable) instead of a billing bucket — likely deliberate, worth a fixture + sign-off.
- ↗ arrow not `aria-hidden`; Details button lacks `aria-controls` `[frontend]`. `cancelled` `^`-anchored patterns coupled to exact `admission.py` wording, untested `[resilience]`. `classify_error_text` not total for exotic/circular dicts (unreachable today) `[correctness]`. Dashboard drops a failed row with empty `error`+`content` (latent) `[edge-cases]`. Silent 1000-char error truncation with no marker `[ux]`.

---

## Worth discussing: persist the server-computed `error_code`, delete the client-side JS classifier `[approach]`

Today two regex engines compile the same patterns, and on reload each JS surface re-classifies the persisted text itself (guarded only by the ASCII fixture). Alternative: add an `error_code` field (as `provider` was added in jarvis#560), centralize in `_mark_errored(name, err, code)`, add it to the `api.py` select; the client passes `m.error_code` as `explicitCode` (the signature already prefers it) and the `compiled = rules.map(new RegExp…)` + `compiled.find(…)` block is **deleted** — JS becomes a pure `code → copy` lookup + provider link.

This eliminates the dual-engine drift class (incl. the non-ASCII divergence), the cross-engine `\\x22`/`\\x27` escaping requirement, and the existing live-vs-reload mismatch for exception-based refinements (`turn-timeout`/`unreachable`) — and gives ops a queryable `code`. Cost: a doctype field + migrate + a small backfill + threading `code` through ~6 emit sites. The other two probes (the JS-as-JSON scheme; the provider text-scrape fallback) were verified to have real constraints — **keep them**. Raising this for a decision, not proposing a silent rewrite.

---

*Panel review generated with [Claude Code](https://claude.com/claude-code).*

kavin-114:
Thanks @venkat102 — validated each finding against the code before changing anything. Addressed in `48dc4149`:

| # | Finding | Verdict | What changed |
|---|---|---|---|
| 1 | Rule-order collisions (`429 model_not_found`, `503 … upstream not found`) | **Valid** (reproduced in JS and Python) | `model-unavailable` now precedes the numeric 429 rule; `not-found` moved after the 5xx/timeout/connection rules. Both cases are fixtures now. |
| 2 | `cancelled` as a red card + dangling `· Details`; Desk Retry | **Valid** | PWA, dashboard and Desk panel render `cancelled` as a muted `role="status"` note like desktop; the `·` separator only follows a non-empty hint. Age-out stays non-retryable on purpose: the muted note has no Retry on desktop either, and the composer is open to resend. |
| 3 | ReDoS | **Valid in principle** (didn't reproduce at 100 KB with my input, but the shape is real) | Both classifiers cap input at 8 KB; the two open-ended `.*` gaps are bounded to `.{0,80}`. A pathological-input timing assertion is in the JS tests. |
| 4 | PWA banner vs card provider context | **Partly** | Today error rows carry no `provider` (see 7), so banner and card both infer from text and agree; kept as is, no code change beyond 7. |
| 5 | Transient failures non-retryable | **Valid** | `no endpoints found` is now a retryable `service-unavailable`; `models-exhausted` is retryable. The narrowed `AgentUnreachableError` override stays (documented, tested): a specific rejection carried by the transport must not be relabelled "unreachable". |
| 6 | Rules file silently degrading to `[]` | **Valid** | Loader refuses fewer than 20 rules (or missing `gateway`/`internal`), names the offending rule on bad JSON/keys/regex, logs the count at import. |
| 7 | Provider text-scan ordering; no stamp on error rows | **Valid** (stamp is success-path only, confirmed in `finalize.py`) | A link is inferred only when exactly one routing host is named, or exactly one provider and no host (`"openrouter and groq"` → none; `"mistral and groq"` → Groq). Stamping a provider onto error rows is a backend change I'd rather take separately; documented in `docs/chat-errors.md`. |
| 8 | Test gaps / a11y / desktop memo | **Valid** | JS tests now pin `retryable` per code, the provider copy override, the legacy `provider` refinement leg, host ambiguity; desktop `errorInfo()` is memoised; history cards drop `role="alert"`; arrow is `aria-hidden`, Details has `aria-controls`. Component-level tests beyond the dashboard spec are still open. |
| 9 | `notifications.js` raw text | **Valid** | Feed body uses the classified headline + hint. `jarvis_mobile` is cross-repo, not in this PR. |

Minor: non-ASCII parity fixed (Python matches with `re.ASCII`, fixture `İnvalid api key` pinned on both engines); `insufficient funds` / expired credit → billing (fixtures); safety hint routes to the administrator; `internal` and `gateway` have distinct headlines; PWA passes `{ provider }` only.

Not done here, for a decision: persisting a server-computed `error_code` (your approach section). I agree it removes the dual-engine class of drift; it needs a field, a migrate and threading through the emit sites, so I'd take it as its own PR if the owner wants it.

Navin-S-R (approved):

PR #1210, opened 2026-09-10
feat(chat): session feedback popup + monthly business pulse survey

kavin-114:
## Summary

Adds a once-per-session feedback popup (after about 10 turns) and a monthly business-pulse survey in chat, so the team gets systematic signal on session quality and whether Jarvis is solving a tenant's business problem. Responses are stored only in the admin fleet dashboard, no local copy on the customer bench.

## What changed

- Session popup: "How did this session go?" fires once per conversation after about 10 turns, five preset reactions, an optional note only for negative reactions.
- Monthly pulse survey: star rating, a dynamically detected feature-usage question (only features the user actually used that month), and an open "is Jarvis solving your business use case" question.
- The existing thumbs up/down bar is suppressed on the same reply the session popup fires, so a user is never shown both at once.
- New `agent_initiated` flag excludes automated agent-run conversations (audits, scheduled reports) from both prompts.
- New `via_voice` flag on chat messages, records whether a message came from mic dictation, feeds the pulse survey's feature list.

## Screenshots

Session popup (after ~10 turns):

session-popup

Monthly business pulse survey:

pulse-dialog

## Risk and verification

- Verified live end to end on e2e.localhost: both dialogs render, submit correctly, and the responses land in the admin dashboard. Session popup confirmed not to re-fire after answering.
- The turn counter is a maintained counter incremented in its own transaction after settlement commits, avoiding a documented lock-order hazard in the settlement path.
- No new scheduled jobs, both prompts trigger from existing chat-turn and chat-open events.
- Review round (Codex, four findings, all confirmed and fixed with regression tests in 137b963b): settlement's turn-count bump can no longer raise when the error log itself fails, the two admin feedback pushes use a 10s timeout instead of 150s, the monthly pulse offer is claimed with a compare-and-set so an answer can never be overwritten back into a re-offer, and the client gate re-checks dismissal after its request returns. 26724385 fixes a pre-existing test-isolation leak in `test_skill_tools` that the resharding exposed.

<details><summary>Design spec and implementation plan</summary>

Full design spec and task-by-task implementation plan available on request, this PR was built through a spec, brainstorm, and reviewed implementation plan process with per-task and whole-branch review passes.

</details>


jarvis/chat/settlement.py
```diff
@@ -187,6 +192,95 @@ def invoke_settlement(
...
+		# row lock is released before the publish, and is never held across it.
+		frappe.db.commit()
+	except Exception:
+		# Guarded like turn_state.lease_lost_exit's: on a connection-level fault the
+		# rollback itself can raise, and "never raises" has to hold literally.
+		try:
+			frappe.db.rollback()
+		except Exception:
+			pass
+		frappe.log_error(title="settlement.bump_turn_count", message=frappe.get_traceback())
```
kavin-114: [P1] Keep logging failures out of chat settlement

`frappe.log_error()` writes to the database and can raise during the same database outage that caused the counter update to fail. This escapes `_bump_turn_count()` after settlement has committed, skipping the immediate `run:end` publish and finalize enqueue. A feedback counter should not interrupt reply completion.

Guard the logging call too. Add a test where both the counter read and error logging fail, and verify terminal publication and finalization still run.

jarvis/admin_client.py
```diff
@@ -1827,6 +1827,20 @@ def push_chat_feedback(item: dict) -> dict:
 	return _post(path=_m("api.tenant.ingest_chat_feedback"), body={"item": item})
 
 
+def push_session_feedback(item: dict) -> dict:
+	"""Push one once-per-session popup response to admin. Called best-effort
+	from jarvis.chat.feedback._forward_session, which swallows failures -- a
+	lost response is acceptable, a blocked popup is not.
+	Raises AdminAuthError / AdminUnreachableError / AdminValidationError."""
+	return _post(path=_m("api.tenant.ingest_session_feedback"), body={"item": item})
```
kavin-114: [P1] Give feedback forwarding a short timeout

Both new push functions inherit `_post()`'s 150-second timeout and run inside the web request. Closing the dialog does not release that server worker. If admin is slow, several feedback submissions can occupy the customer's web workers and delay chat and ERP requests.

Use a short feedback timeout, including the authentication path. Add an admin-timeout test that verifies the request finishes promptly without retrying for minutes.

jarvis/chat/feedback.py
```diff
@@ -93,3 +121,394 @@ def _forward(payload: dict) -> None:
...
+	# A user chatting before their settings row exists is normal (the row is
+	# created lazily by whichever surface needs it first), and an offer has to be
+	# recorded somewhere, so this is where the create belongs.
+	if row:
+		settings_name = row.name
+	else:
+		from jarvis.chat.usage import get_or_create_user_settings
+
+		settings_name = get_or_create_user_settings(user).name
+	_store_pulse_state(settings_name, current_key, offer_count + 1)
```
kavin-114: [P2] Make the monthly offer claim atomic

A pending `pulse_context()` can read count 1, pause while collecting features, then overwrite count 3 written by a completed survey submission with count 2. The user is offered the survey again after answering. Concurrent opens can also lose increments and exceed the three-offer cap.

Use an atomic claim that rechecks the period and cap. Add concurrent-open and offer-versus-submit tests. I reproduced the answered count changing from 3 back to 2.

frontend/src/lib/pulseFeedbackGate.js
```diff
@@ -0,0 +1,37 @@
...
+// pulse_context() again, which is still due -- reopening the survey and
+// burning another one of the three monthly offers. A page reload resets this
+// (module state, not persisted), which is the intended boundary.
+let _dismissedThisLoad = false;
+
+export async function maybeOpenPulseFeedback() {
+	if (_dismissedThisLoad) return;
+	try {
+		const ctx = await pulseContext();
+		if (ctx && ctx.due) {
```
kavin-114, line 30: [P2] Ignore pending survey responses after dismissal

Two chat opens can start two `pulseContext()` requests. If the first response opens the survey and the user clicks “Maybe later,” the second response still opens it again: `_dismissedThisLoad` is checked only before the await.

Deduplicate pending checks and recheck dismissal after the response. Add a deferred-response test: start two checks, dismiss the first dialog, then resolve the second. I reproduced the dialog reopening.

kavin-114 (commented):
Please hold merge until the issues below are fixed and covered by regression tests.

Verified at `29ba59d4`: 1,867 frontend tests pass on Node 24, the desktop build passes, and Ruff passes on the reviewed backend files. GitHub CI is green. Both dialog screenshots reviewed.

I reproduced the two survey races and the settlement error path in isolated tests. These cases are missing from the PR tests. I did not run database-backed tests locally or test on customer sites.

jarvis/chat/settlement.py
kavin-114: Confirmed and fixed in 137b963b. `frappe.log_error` in the except branch is now guarded the same way the rollback is: it INSERTs an Error Log row over the connection that just failed, so during a DB outage it raised too and escaped a function documented as never raising, right before the run:end publish. Regression test: `test_bump_never_raises_even_when_the_error_log_itself_fails` (sql, rollback and log_error all raising; rollback/commit stubbed so the test's own transaction survives).

jarvis/admin_client.py
kavin-114: Confirmed and fixed in 137b963b. Both pushes run synchronously inside the submit endpoint, so the inherited 150s default could pin a web worker per submit. New `_FEEDBACK_TIMEOUT_S = 10` on `push_session_feedback` and `push_pulse_feedback`, same rationale as `_MODEL_CATALOG_TIMEOUT_S`; both callers already treat any failure as a lost, acceptable response, so a timeout is the same outcome reached sooner. Scoped deliberately: the pre-existing `push_chat_feedback` has the identical shape and keeps the default, noted in the comment rather than changed here. Regression test: `test_both_pushes_use_a_short_timeout_not_the_150s_default`.

jarvis/chat/feedback.py
kavin-114: Confirmed, reproduced by trace, and fixed in 137b963b. The offer write is now `_claim_pulse_offer`, a compare-and-set pinned to the key and count that were read (same idiom as `_claim_session_feedback`, rowcount read before any commit). Losing the race returns not due, so an answer landing during the feature sweep can no longer be dragged from 3 back to 2, and two chat opens racing burn one offer instead of two. Two predicates: same period compares key and count; rollover (or a just-created row with a NULL key) compares only the key and resets to 1. Regression tests: `test_an_answer_during_the_sweep_is_never_overwritten_by_the_offer` (your 3 -> 2 repro), `test_two_chat_opens_racing_burn_exactly_one_offer`, `test_a_rollover_claim_is_conditional_too`.

frontend/src/lib/pulseFeedbackGate.js
kavin-114: Confirmed and fixed in 137b963b. `_dismissedThisLoad` is re-checked after the `await pulseContext()`, so a dismissal while a second check is in flight no longer reopens the survey. Regression test in `pulseFeedbackGate.spec.js` holds the mock promise open, dismisses, then resolves with due: true and asserts the dialog stays closed; it fails 1/6 against the previous gate and passes now. Together with the server-side CAS above this also closes the second monthly offer such a race used to burn.

kavin-114:
Review round status. All four inline findings are fixed in 137b963b with regression tests (replies on each thread).

CI on 137b963b went red on `tests (4)`: `test_turn_usage.test_top_tools_tenant_wide_cap`, which is not a file this PR touches. Cause: the three new test files resharded `test_skill_tools` ahead of `test_turn_usage` (shard 4 on the green 29ba59d4 run differs by exactly that module), and `TestReceiptRefStamping` there leaked a committed `run_report` receipt: `persist_tool_receipt` commits for durability, while the class tearDown's conversation delete was uncommitted and rolled back. 26724385 makes that tearDown delete the receipts too and commit the cleanup, same shape as the other conversation fixtures. Fresh run in progress on 26724385.

https://claude.ai/code/session_0121RjwTjC4SaFQMn1vHTbvB

Navin-S-R (approved):

frontend/src/lib/pulseFeedbackGate.js
kavin-114: Correction: the spec I cited here was verified locally (fails 1/6 on the pre-fix gate) but was left out of commit 137b963b, so it is not in the merge. It is in #1213 (test only, one new case in pulseFeedbackGate.spec.js). The fix itself is in the merge.

PR #1256, opened 2026-09-11
feat(onboarding): GSTIN party-details autofill via the tenant's own India Compliance

Navin-S-R:
## What

On the onboarding **Details** step, a **"Fetch details"** button prefills the billing party fields (legal/trade name, registered address, state) from the customer's **GSTIN**, using **this tenant site's own India Compliance** — in-process, same-site. No control-plane or books-site hop, no guest endpoint; the customer's own IC + API credits (free on Frappe Cloud).

The button is offered **only where this site's IC autofill API is usable** — `india_compliance` installed **and** `is_production_api_enabled()` (Enable API + a credential, incl. the FC-injected `frappe.conf.ic_api_secret`, + not sandbox). Everywhere else the form is plain manual entry.

## Why it's safe

- **Fail-open, always.** A malformed GSTIN, a GSP hiccup, a slow call (9s client-side timeout), an absent/mapping fault, or IC being unusable → a soft miss + manual entry. The lookup never blocks signup and writes nothing server-side.
- **Auth.** Both endpoints (`gstin_autofill_available`, `gstin_autofill`) are `require_jarvis_admin`-gated (same as `start_signup`); IC is called in-process behind lazy-import seams so a non-IC site never imports it. Registry values are stripped of the Unicode categories a party save rejects and length-clamped.
- **Prefill UX.** Overwrites the fetched fields but keeps a **one-shot Undo**; values are user-owned (win the stale-response fence); `business_name` maps to the invoicing party; the state resolves to a valid option (or the existing value stays and is validated); country coerces to India and the India/Overseas validators re-run; a non-Active GSTIN shows a non-blocking advisory.

## Tests

- Backend: 15 tests — auth gate + invalid-vs-unavailable + mapping-fault fail-open **mutation-verified**; IC stubbed at the seams so an IC-less `test_site` passes.
- Frontend: vitest for the state resolver + `applyGstinPrefill`/undo; full onboarding vitest **168/168**; `vite build` clean; prettier@2.7.1.

## Flow review — deferred to staging (read before merge)

A local dev site has **no Frappe-Cloud `ic_api_secret`** (and jarvis.local is in GST sandbox + already onboarded), so `is_production_api_enabled()` is False locally and the **live happy-path prefill cannot be exercised on a dev bench**. The wiring was verified by review + unit tests + a clean build, but the **real-browser flow review — button appears on an IC-enabled site, prefill populates the fields, Undo restores, and the button is hidden where IC isn't usable — must be run on FC/staging before merge.**

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -133,6 +133,12 @@ export const workspaceResetState = () => call("jarvis.onboarding.workspace_reset
 export const resetOnboarding = (wipeData = true) =>
 	call("jarvis.onboarding.reset_onboarding", { wipe_data: wipeData ? 1 : 0 });
 
+// GSTIN party autofill on the Details step, using THIS site's own India Compliance (offered only
+// where its lookup API is usable). `available` gates the button; `gstin_autofill` returns the party
+// details ({found, ...}) or a soft miss — the form always falls back to manual entry.
+export const gstinAutofillAvailable = () => call("jarvis.onboarding.gstin_autofill_available");
+export const gstinAutofill = (gstin) => call("jarvis.onboarding.gstin_autofill", { gstin });
+
 // --- Per-user chat settings + real (measured) usage tracking, incl. the
 // tenant-admin usage table (design doc §4/§6). All return the house
 // {ok, data} / {ok:false, reason} envelope — unlike getUsage above, which is
--- a/frontend/src/onboarding/indianStates.js
+++ b/frontend/src/onboarding/indianStates.js
@@ -339,3 +339,24 @@ export function isIndia(country) {
 export function isValidIndianState(state) {
 	return _STATE_SET.has((state || "").trim().toLowerCase());
 }
+
+// India Compliance derives a party's state from the GSTIN (titlecase of the registry `stcd`), so
+// the incoming string is already GSTIN-authoritative — but its spelling may not be verbatim one of
+// INDIAN_STATES (e.g. "Lakshadweep Islands" vs our "Lakshadweep", or a "& "/"Pondicherry" variant).
+// Resolve to the exact option a case-insensitive match or a known alias yields, else "" — a GSTIN
+// autofill must land a value the State <select> can actually show, never inject an unshowable one
+// that then blocks the form; on no confident match the field is left blank for manual pick.
+const _STATE_LC = new Map(INDIAN_STATES.map((s) => [s.toLowerCase(), s]));
+const STATE_ALIASES = {
+	"lakshadweep islands": "Lakshadweep",
+	"daman and diu": "Dadra and Nagar Haveli and Daman and Diu",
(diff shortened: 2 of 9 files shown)


Navin-S-R:
Folded two things onto this branch (`d026d50c`):
- **Foreign customers**: GSTIN + its autofill are now hidden (`v-if="billingIsIndia"`, mirroring the State field), and a `watch(billingIsIndia)` clears the value/error/lookup-state on any switch to a non-India country — so a foreign customer never sees or submits a GSTIN.
- **Blur-trigger**: replaced the "Fetch details" button with an on-blur trigger (`onGstinBlur`). It runs only for a new, checksum-valid GSTIN the tenant's IC can resolve (`canFetchGstin`) and only when it changed since the last settled lookup (new `_gstinLastLookedUp` marker), so re-blurring an unchanged value never re-hits the GST API; the existing stale-response fence still applies. A single derived status line (`gstinAutofillNote`) replaces the button's affordances (pre-fetch cue / fetching… / result / "filled from your GSTIN"); scoped Undo stays.

Note: prettier-clean and all referenced symbols verified present; local `vite build` couldn't run in the isolated worktree (node-20 bin resolution) so CI is the build gate here.

Navin-S-R:
### Browser-verified locally (frontend UX)

Verified the onboarding Details step end-to-end in Chrome on an un-onboarded local tenant (`jarvis_2.local`), by temporarily building this branch's frontend into the served assets with a **mocked** IC lookup, then reverting:

- **GSTIN leads Invoicing details** — first field of the block. ✅
- **Blur autofill (no button)** — typed a valid GSTIN (`33ABCDE1234F1Z7`) and tabbed out → prefilled Invoicing company name (`Acme Registered Pvt Ltd`), Address 1/2, City (`Chennai`), Postal (`600001`), State (`Tamil Nadu`), with the "Company details filled from your GSTIN." note and the scoped **Undo autofill**. ✅
- **Foreign hide/clear** — switching Country to **United States** removed the GSTIN field + autofill entirely, flipped State to free-text "Region", and cleared the GSTIN value (prefilled address retained for editing). ✅
- `vite build` clean.

**Mocked, and why:** the two API calls (`gstinAutofillAvailable`, `gstinAutofill`) were stubbed with canned data because no local tenant has a **production GST API credential** — `jarvis.local` is `sandbox_mode=1` + `api_secret=null` (capability off) *and* onboarded (wizard guarded); `jarvis_2.local` has IC installed but is likewise sandbox/no-secret. So this exercises the **frontend UX** (blur-trigger, prefill wiring, note, Undo, foreign hide/clear); the **real** IC/GSP call is covered by `test_onboarding_gstin_autofill.py` (which mocks IC) and still needs a live-credential pass on **staging/FC**.

venkat102 (approved):

PR #1337, opened 2026-09-21
feat(filebox): skill discovery + wiki read/fenced-write-back on file drop

Navin-S-R:
## What
When a file is dropped into the File Box, the unattended run now:
1. **Classifies** the document.
2. **Discovers a matching processing skill** (`find_skills` / a user-authored `Jarvis Custom Skill`, else the always-on `ocr-data-entry` fallback) — this is how a user says "when a file like X arrives, do Y".
3. **Consults the wiki** (`read_wiki`) for existing content before drafting, building on it instead of starting from scratch.
4. **Records findings back to the wiki** after the draft.

The **drafts-only / never-ask / ambiguity→Approval** safety envelope now explicitly **overrides** any discovered skill.

## Wiki write-back is fenced (the load-bearing safety design)
A dropped file is untrusted, and a File Box run is unattended (no confirm card). So `update_wiki` in a `file_box` conversation is **not** routed to the raw tool (which defaults `scope=Org` + `ignore_permissions` and can full-body-replace a curated page). Instead a new **file_box-only** dispatch branch routes it through the append-only, provenance-fenced funnel `apply_extracted_page_updates(source="file-box:", provenance_prefix="file-box", allow_body_replace=False, preserve_curated=True)` — the same machinery the app-learning scribe uses. Net effect: an untrusted file can only **create or append to File-Box's own pages**, is **refused** on a slug colliding with a human/curated/other-feature page, and **never overwrites**. `replace_body_md` is downgraded to an append.

- **Scoped:** the branch keys on `and file_box` — attended chat and admin `auto_apply` still park a confirm card for wiki edits; the wiki kill-switch still refuses a wiki-off write first; destructive tools still park.
- **Audited:** both the `tool_audit` log line and the durable `Jarvis Agent Write` manager-board row are filed (the fenced branch returns before `_dispatch_and_wrap`, so it files them itself).
- **Best-effort:** a transaction-fatal wiki error rolls back and returns a soft `{ok:false}` — it never fails the unattended run.

## Reviewed
Front-gate (plan-check) caught the original raw-`update_wiki` design as an org-poisoning vector; this fenced design is the fix. Back-gate (review-loop): security / correctness / resilience (opus) + testing (sonnet) lanes all CONFIRMED-clean, no Critical/Important; the LOW/MODERATE findings (best-effort txn-fatal, manager-board audit parity, extra tests) are folded in.

## Tests (on the jarvis test site)
`test_confirm_gate` +13 file-box cases (fenced routing exact kwargs, replace→append, refusal, normal/admin-auto_apply park, wiki-off refused, destructive parks, create/update auto-apply unregressed, both audit rows, txn-fatal best-effort, empty-outcomes) — mutation-verified the `and file_box` guard. `test_filebox` intent guard. Regression: `test_api` (23), `test_chat_api` (65+14) green. ruff clean.

## Follow-up (separate PR)
Persona reinforcement (`jarvis-persona`: `ocr-data-entry` wiki step + `AGENTS.md` File-Box routing + skill-writer authoring note) — reinforcement-only; the app `INBOUND_PROMPT` here is authoritative and reaches tenants on deploy without a reseed.


kavin-114:
**Review finding: approve the exact note the reviewer read (P1)**

The proposed wiki note can change while someone is reviewing it.

Example: I open a note saying “Acme's payment terms are 30 days.” File Box then refreshes that same pending proposal to say “90 days.” My screen still shows 30 days, but clicking Approve can publish 90 days.

This happens because pending proposals can be updated, while Approve sends only the proposal name, not the version I read.

Please tie approval to the version/content shown to the reviewer. If it changed, ask them to read the updated note before approving.

Code: [pending note refresh](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/api.py#L2163), [approval handler](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/chat/approvals_api.py#L876).

kavin-114:
**Review finding: adding a wiki note can remove older knowledge (P2)**

The new path is described as append-only, but once a wiki page grows beyond 20,000 characters, adding another note can silently remove the oldest text.

Even after a reviewer approves a File Box note, the page is still marked as machine-created. The existing protection for human-reviewed content therefore does not protect these pages from truncation.

Please preserve the earlier approved content. If an addition is too large, refuse it with a clear explanation or put it on another page rather than deleting old knowledge.

I reproduced this using the existing merge function with the new File Box settings. Code: [File Box write settings](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/api.py#L2067), [existing size-limit handling](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/chat/wiki.py#L1464).

kavin-114:
**Review finding: turning off the wiki should also stop pending writes (P2)**

Example: File Box proposes a wiki note while the wiki is enabled. An admin then disables the wiki. Approving or retrying that old proposal can still write to it.

The wiki-enabled check runs when the agent proposes the note, but the later approval/retry path skips that check.

Please check that the wiki is still enabled immediately before applying an approved note, including retries. Keep the proposal available so it can be retried if the wiki is enabled again.

Code: [approval/retry landing path](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/chat/approvals_api.py#L849), [write helper](https://github.com/aerele/jarvis/blob/e6b10ccb131a0d07cea5e8f79d759aea8a3e72ab/jarvis/api.py#L2025).

kavin-114:
**File Box enhancement: automatically use the right skill for bulk document uploads**

The main workflow I want File Box to support is this:

A user writes a skill for a particular kind of document. For example: “When I upload these customer sales invoices, use these mappings and create draft Purchase Invoices in our ERP.” They then upload a batch of those invoices to File Box. File Box should automatically find and follow that skill for each matching file, without requiring the user to repeat the instruction or manually select the skill each time.

The document's printed title should not override the user's intended workflow. In this example, seeing “Sales Invoice” on the source must not make File Box create Sales Invoices when the matching skill says to create Purchase Invoices. Party and field mappings should follow the skill too.

**What this PR already covers:**

- Bulk upload is supported.
- The prompt tells the agent to discover a matching skill automatically, with OCR/data-entry as the fallback.
- An optional “Also apply skill” picker is available.
- The default prompt tells every File Box run to consult the wiki for the party/topic before drafting, even when no extra skill is selected. This is the expected default when the wiki is enabled and available.

**What still needs coverage:**

- Please add a workflow test for the example above: bulk upload matching sales invoices, automatically use the saved skill, and create the intended draft Purchase Invoices with the correct mappings. The new tests check the prompt and skill selection plumbing, but do not demonstrate this complete processing flow.
- Please make the routing rules clear when the discovered/default skill and the selected extra skill disagree. A conflicting or unavailable intended skill should lead to an Approval Request rather than silently creating the wrong kind of invoice.
- Consider a **“Use in File Box” checkbox on Skills**, so users can choose which skills participate in automatic File Box processing. This is not present in the current PR. It would help keep unrelated skills out of document processing; the agent should then load the relevant matching skill from that eligible set.
- Keep relevant wiki lookup part of every file's normal processing, regardless of skill selection. If the wiki is disabled or unavailable, continue using the existing documented fallback.

Automatic discovery and wiki lookup are currently instructions to the agent, not proof that this business workflow completes correctly. Please verify the full scenario before treating it as covered.

Checked against commit `e6b10ccb`. The 8 new frontend tests passed; the database-backed Frappe suites were not run in this review.

